Issue #387, opened 2014-12-29
Auth header not being sent if tokenAttributeName is changed & Rails is expecting 'token', not 'user_token'


matthooks:
I ran into a couple of issues using ember-simple-auth and ember-simple-auth-devise (both 0.7.2) with Rails 4.2 and ember-rails.

With the default configuration, ember-simple-auth-devise expects a `user_token` and `user_email` from the server in order to set up a session, and then sends both in the `Authorization` header on subsequent requests e.g. `"Token user_token=\"p-bbux4Br1ULzvzjTGGv\", user_email=\"test@example.com\""`.

The `authenticate_with_http_token` method within Rails, however, uses [`"token="`](https://github.com/rails/rails/blob/25b14b4d3238d5474c60826ee1b359537af987ef/actionpack/lib/action_controller/metal/http_authentication.rb#L400) as a key to find the token from the `Authorization` header. This means that the token that it finds ends up parsed like: `user_token="p-bbux4Br1ULzvzjTGGv` instead of just `p-bbux4Br1ULzvzjTGGv`, causing the authentication filter to fail, at least as it is written in the ember-simple-auth-devise README. (issue 1)

To attempt to fix the problem, I set the `tokenAttributeName` setting within ember-simple-auth-devise to `token`. Since this setting also changes what ember-simple-auth expects back from the server, the [`authorize` function](https://github.com/simplabs/ember-simple-auth/blob/8cdbab604b8ea6618c00304e8750039ccdc863c1/packages/ember-simple-auth-devise/lib/simple-auth-devise/authorizers/devise.js#L69) within ember-simple-auth-devise `get`s `token`, instead of the `user_token` (which the server returns, at least as it is written in the README). Thus, the `Authorization` header on the outgoing API request is not set, causing the token authentication to fail. (issue 2)

Double whammy.

I was able to finally solve the issue by changing `user_token` to `token` in the response from the `create` action in my session controller.

The Rails function should probably be configurable to properly parse the Authorization header, but I also wonder if in the mean time we should set the default `tokenNameAttribute` to be `token` instead of `user_token` since that's what Rails is expecting, and devise is a Ruby plugin?

marcoow:
You're right, the token should always be sent as `token`. Pull request would be cool.

marcoow:
fixed in #394

Tonkonozhenko:
How can I get latest changes from master for my Rails app?
Now I attach library via Rails Assets

```ruby
# Gemfile
[...]

PR #394, opened 2015-01-08
Change default devise tokenAttributeName to token

SebastianSzturo:
Changes the default devise ```tokenAttributeName``` to ```token```.
Works out of the box with [the rails implementation of HTTP token authentication](https://github.com/rails/rails/blob/25b14b4d3238d5474c60826ee1b359537af987ef/actionpack/lib/action_controller/metal/http_authentication.rb#L400) and the  ```authenticate_with_http_token``` method from the current README.

See #387 for more details.

### DIFF
--- a/packages/ember-simple-auth-devise/README.md
+++ b/packages/ember-simple-auth-devise/README.md
@@ -64,7 +64,7 @@ class SessionsController < Devise::SessionsController
         self.resource = warden.authenticate!(auth_options)
         sign_in(resource_name, resource)
         data = {
-          user_token: self.resource.authentication_token,
+          token: self.resource.authentication_token,
           user_email: self.resource.email
         }
         render json: data, status: 201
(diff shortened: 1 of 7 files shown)


dandehavilland:
@marcoow: as minor as it is, I think this PR warranted a bump to `0.8`. Took me a few hours today to track this down as the source of failing API requests. (I was using the previous default value of `tokenAttributeName: 'user_token'`).

marcoow:
Yeah, 0.8.0 was changed to 0.7.3.

dandehavilland:
0.8 seems more appropriate. Why change back to 0.7.3?

marcoow:
Because [the bigger changes](https://github.com/simplabs/ember-simple-auth/pull/414) I hadn't planned didn't make it into the release and only smaller fixes remained.

PR #398, opened 2015-01-14
Authorize cross domain requests to all subdomain of host using wildcard

dukex:
Add a easy to developers enable request for all subdomain, like

```
ENV['simple-auth'] = {
  crossOriginWhitelist: ['http://*.other.domain']
}
````

That is important for multitenancy application, in many cases the server get the correct client data by subdomain, today using simple-auth is necessary put a [*] on crossOriginWhitelist and check url on authorizer

### DIFF
--- a/packages/ember-simple-auth/lib/simple-auth/setup.js
+++ b/packages/ember-simple-auth/lib/simple-auth/setup.js
@@ -4,7 +4,7 @@ import LocalStorage from './stores/local-storage';
 import Ephemeral from './stores/ephemeral';
 
 function extractLocationOrigin(location) {
-  if (location === '*'){
+  if (location.indexOf("*") > -1){
       return location;
   }
   if (Ember.typeOf(location) === 'string') {
@@ -24,14 +24,26 @@ function extractLocationOrigin(location) {
   return location.protocol + '//' + location.hostname + (port !== '' ? ':' + port : '');
 }
 
+function matchDomain(urlOrigin){
+  return function(domain) {
+    if (domain.indexOf('*') > -1) {
+      var domainRegex = new RegExp(domain.replace("*" , ".+"));
+      return urlOrigin.match(domainRegex);
+    }
+
+    return domain.indexOf(urlOrigin) > -1;
+  };
+}
+
 var urlOrigins     = {};
 var crossOriginWhitelist;
 function shouldAuthorizeRequest(options) {
   if (options.crossDomain === false || crossOriginWhitelist.indexOf('*') > -1) {
     return true;
   }
+
   var urlOrigin = urlOrigins[options.url] = urlOrigins[options.url] || extractLocationOrigin(options.url);
-  return crossOriginWhitelist.indexOf(urlOrigin) > -1;
+  return Ember.A(crossOriginWhitelist).any(matchDomain(urlOrigin));
 }
 
 function registerFactories(container) {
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: packages/ember-simple-auth/tests/simple-auth/setup-test.js [hunk 1]
```diff
@@ -168,6 +168,22 @@ describe('setup', function() {
...
+        expect(this.authorizer.authorize).to.have.been.calledOnce;
+
+        Ember.$.get('http://another-test.other-domain.com/path/query=string');
+
+        expect(this.authorizer.authorize).to.have.been.calledTwice;
+
+        Ember.$.get('http://test2.other-domain.com/path/query=string');
+
+        expect(this.authorizer.authorize).to.have.been.calledThrice;
+      });
```
marcoow: Can you add test cases for

- `http://test2.other-domain.com:80/path/query=string` // should be called
- `http://test2.other-domain.com:1234/path/query=string` // should not be called

plus a test for when you configure `Configuration.crossOriginWhitelist = ['http://*.sub.other-domain.com:1234'];`:

- `http://test2.sub.other-domain.com/path/query=string` // should not be called
- `http://wrong.test2.other-domain.com:1234/path/query=string` // should not be called
- `http://sub.test2.other-domain.com:1234/path/query=string` // should be called

dukex: I added this tests, and works fine on browser. but on phantomjs I have problems with `*` in url, the hostname is empty

To reproduce: 
```` bash
$ echo "var link = document.createElement('a'); link.href='http://*.other-domain.com/mytest'; console.log(link.href) ; console.log(link.hostname)" | phantomjs

phantomjs> http://*.other-domain.com/mytest // <---href
// <--- hostname
0 // <--- port
```

Do you have any ideia for it?

dukex: Tests added!

packages/ember-simple-auth/lib/simple-auth/setup.js
```diff
@@ -3,10 +3,17 @@ import Session from './session';
...
 import Ephemeral from './stores/ephemeral';
 
+var wildcardToken = "WILDCARD_TOKEN";
+
 function extractLocationOrigin(location) {
   if (location === '*'){
       return location;
   }
+
+  if (location.indexOf('*') > -1) {
```
marcoow: don't actually need this `#indexOf`

packages/ember-simple-auth/tests/simple-auth/setup-test.js
```diff
@@ -168,6 +168,42 @@ describe('setup', function() {
...
+        expect(this.authorizer.authorize).to.not.have.been.callCount(5);
+
+        Ember.$.get('http://wrong.test2.test-domain.com:1234/path/query=string');
+
+        expect(this.authorizer.authorize).to.not.have.been.callCount(5);
+
+        Ember.$.get('http://test2.sub.test-domain.com:1234/path/query=string');
+
+        expect(this.authorizer.authorize).to.have.been.callCount(5);
+      });
```
marcoow: Can you add a test for 

```
Ember.$.get('http://test2.sub.test-domain.com:1235/path/query=string'); // should not result in a call
```

I think that would fail...

dukex: Done! Don't failed

packages/ember-simple-auth/lib/simple-auth/setup.js
dukex: Removed

marcoow:
Cool, thanks!

marcoow:
Argh, merged too early - I think I'm too tired...

Can you create another PR that updates [the docs](https://github.com/simplabs/ember-simple-auth/blob/e68dddb0c732b466c2eb4726305592463044e1b7/packages/ember-simple-auth/lib/simple-auth/configuration.js#L129)?

Issue #383, opened 2014-12-23
Store data when not authenticated


tehmaestro:
Hi and thank you for this library, it's pretty great. One question, how can I persist data across page reloads, when the user is not authenticated. I want to store some information, and take that information to the server only when the user decides to create an account and thus becoming authenticated.

Thanks.

snovity:
@tehmaestro it seems that your solution with `invalidate()` call will still trigger a delete if your have two parallel anonymous sessions in two tabs. Say, if tab 1 updates anonymous session then tab 2 will delete everything and, in your case, will hit a back-end. 

I have the same problem as @tehmaestro and have an idea that maybe the best solution would be to have a configurable anonymous authenticator (not sure what would be the best name to call it). It would have only `restore` method and by default this method would do nothing. 

And it seems that this fix isn't that trivial, there are corner-cases and many parts of session are affected.

snovity:
Also, maybe it would make sense to have a subkey on session called `anonymous` or something like that wouldn't be affected by authentications/invalidations. Just thinking out loud :) Would be interesting to hear what @marcoow thinks.

marcoow:
I think the simplest fix would probably be to not clear the session when the authenticator's `restore` method returns a rejecting promise. Not sure whether this covers all edge cases though...

tehmaestro:
Hey, one problem I noticed while further researching this, is that, when not authenticated, although you can use `this.set('session.something', something)`, using `this.get('session.something;)` does not work.
  Only when the session is authenticated, setting and getting stuff from localStorage works.

  I probably don't fully understand how Ember works, but I simply cannot find the lines of code that make the localStorage data available in the session object, when authenticated. I'm trying to modify the behavior so it fits my needs, but I just can't find it.
  Could you point me to the right direction? How and when does the data became available in the session object, like `this.get('session.data')` . Thank you.

marcoow:
Yep. it uses `Ember.ObjectProxy` and the [`setUnknownProperty` method](https://github.com/simplabs/ember-simple-auth/blob/master/packages/ember-simple-auth/lib/simple-auth/session.js#L279) to keep the store up to date.

marcoow:
I thought about the topic and I think the fact that the session clears itself in certain situations war wrong from the beginning. The fact that the session's current content doesn't constitute an authenticated state doesn't require all of its content to be cleared. Yet, there's probably a simple solution for the problem:

[...]

PR #410, opened 2015-01-24
Store data in unauthenticated session

marcoow:
See #383

Right now Ember Simple Auth allows storing data also in the unauthenticated session but when the session is restored that data is completely ignored. This branch refactors the session/authenticators interplay so that no data gets lost.

### Concept

As both the authenticator's `restore` and `invalidate` method receive the session's current content as an argument it would make sense to - if that session content doesn't constitute an authenticated state which is the authenticator's task to decide anyway - have the authenticator remove the security-relevant attributes from the session content (which would in most cases be the token and if not then the authenticator would know which attributes have to be removed, e.g. expiration times or refresh tokens in the case of the OAuth 2.0 authenticator), resolve or reject with that updated session content and in the session itself simply replace its content with whatever comes back from the authenticator and never completely clear itself (although it probably makes sense to have a method in the public API for that). That solution would have several advantages:

- the public API doesn't need to change
- no new concepts like anonymous authenticators etc. have to be introduced
- the session actually behaves like a real session in the way that it keeps data regardless of whether it's authenticated or not

There's 2 drawbacks that I would consider ok for a pre 1.0 release though:

- custom authenticators would potentially have to be changed so that they always resolve or reject with the updated session content and also delete relevant attributes from it in the `restore` and `invalidate` methods
- if data is persisted in the session that's relevant to security but the authenticator is unaware of, that data would have to be deleted manually (in the `sessionInvalidationSucceeded` action if using the `ApplicationRouteMixin` or by handling the session's `sessionInvalidationSucceeded` event)

### TODO

- [ ] review docs to make sure the new behavior is correctly reflected
- [ ] add docs for the torii authenticator mentioning the need to cleanup the session manually
- [ ] add a new example that shows how to store data in the session manually
- [ ] this seems to break the existing examples somehow so those need to be fixed of course

### DIFF
--- a/examples/1-simple.html
+++ b/examples/1-simple.html
@@ -4,7 +4,7 @@
     <title>Ember Simple Auth - examples</title>
     <meta name="viewport" content="width=device-width, initial-scale=1.0">
     <link rel="stylesheet" href="http://netdna.bootstrapcdn.com/bootstrap/3.0.0/css/bootstrap.min.css">
-    <script src="http://code.jquery.com/jquery-1.9.1.js"></script>
+    <script src="http://code.jquery.com/jquery-2.1.3.js"></script>
     <script src="https://cdnjs.cloudflare.com/ajax/libs/handlebars.js/2.0.0/handlebars.js"></script>
     <script src="http://builds.emberjs.com/release/ember.js"></script>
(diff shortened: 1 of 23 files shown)


marcoow:
This is replaced by the alternative approach to fixing the problem #414 .

PR #412, opened 2015-01-25
Fixed 'approval_prompt' in ext. provider example

NathanJang:
I believe this is a typo that prevented logging in with Google from working.

### DIFF
--- a/examples/7-multiple-external-providers.html
+++ b/examples/7-multiple-external-providers.html
@@ -163,7 +163,7 @@
             gapi.auth.authorize({
               client_id:        '694766332436-1g5bakjoo5flkfpv3t2mfsch9ghg7ggd.apps.googleusercontent.com',
               scope:            ['https://www.googleapis.com/auth/plus.me'],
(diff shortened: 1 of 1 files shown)


marcoow:
According to the [API docs](https://developers.google.com/api-client-library/javascript/reference/referencedocs#gapiauthauthorize) neither an `approvalprompt` nor an `approval_prompt` argument exist for that method...

NathanJang:
However, `approval_prompt` is [a valid request parameter](https://developers.google.com/accounts/docs/OAuth2WebServer#formingtheurl) for the Google API. The javascript library just parses the json parameters and forms the URL.

marcoow:
Ah, I see. Thanks!

NathanJang:
Thanks! In fact, I probably should also have removed the extraneous quotation marks around the parameter name, but that's not important enough for another pull request :)

marcoow:
No, please go ahead and open a new one - also the little things matter.

NathanJang:
Will do then when I get off mobile.

PR #422, opened 2015-02-04
Instructions in Readme for user_token cause session invalidation

atonse:
In the SessionsController, you return the token as "token" - but when retrieving the session on page load, the authenticator looks for "user_token" so the user is logged out on page load if we follow these instructions.

If you return "user_token" with "user_email", then the user stays logged in.


marcoow:
`user_token` was changed to `token` in 0.7.3 so the documentation you see on the master branch is actually correct. Previously the master branch and the released versions had diverged.

PR #425, opened 2015-02-09
ApplicationRouteMixin: Allow to override default route names

simonihmig:
Use case: 
* you have different sections of an web app with different UI
* you have implemented login by using a modal dialog that overlays your UI
* so you need different login routes, such as /sectionA/login (for displaying login dialog over "sectionA" UI) and /sectionB/login (for displaying login dialog over "sectionB" UI)
* so you need to override the default `Configuration.authenticationRoute` and `Configuration.routeAfterAuthentication` at route level 

Solution:
* use `ApplicationRouteMixin` in different routes (/sectionA and /sectionB)
* set `authenticationRoute` and `routeAfterAuthentication` as properties on the route (properties default to those of global configuration object)

Hope this to be useful for others...
Grüße aus Hamburg! ;)

### DIFF
--- a/packages/ember-simple-auth/lib/simple-auth/mixins/application-route-mixin.js
+++ b/packages/ember-simple-auth/lib/simple-auth/mixins/application-route-mixin.js
@@ -59,6 +59,27 @@ var routeEntryComplete = false;
   @static
 */
 export default Ember.Mixin.create({
+
+  /**
+    The name of the authentication route, the authenticateSession action transitions to.
+
+    @property authenticationRoute
+    @default Configuration.authenticationRoute
+    @public
+  */
+  authenticationRoute: Configuration.authenticationRoute,
+
+
+  /**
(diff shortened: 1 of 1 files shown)


marcoow:
Hm, I'm not sure there's so much benefit in comparison to the added complexity (2 levels on which you can define properties like `authenticationRoute` etc.). Also instead of doing this you could simply use the `ApplicationRouteMixin` in the different section's root routes and override the actions that use these properties like `authenticateSession`.

Also there are not tests for the added behavior.

simonihmig:
Sure I could do that, but having to duplicate the code, especially that of `sessionAuthenticationSucceeded`, did not feel right to me.

I can add tests, if that is the only reason to not incorporate the changes...

marcoow:
I don't want to keep adding stuff to the library that could handle every possible case. As long as a clear way is provided to solve this edge case I'd say that's fine. Also the mixing are simply for convenience anyway I don't think it's necessary to add this.

PR #429, opened 2015-02-10
Add support for setting oauth2 token in an access_token parameter

kamal:
This PR implements Section 2.2 and Section 2.3 of the RFC.

We recently had to implement sending the oauth2 token via an `access_token` parameter to support IE9 users. As we host our Ember app on a subdomain, IE9 users use XDomainRequest to perform CORS. XDomainRequest does not support setting request headers so using the  `oauth2-bearer` authorizer doesn't auth our requests.

### DIFF
--- a/packages/ember-simple-auth-oauth2/lib/simple-auth-oauth2/authorizers/oauth2-access-token.js
+++ b/packages/ember-simple-auth-oauth2/lib/simple-auth-oauth2/authorizers/oauth2-access-token.js
@@ -1,0 +1,69 @@
+import Base from 'simple-auth/authorizers/base';
+
+/**
+  Authorizer that conforms to OAuth 2
+  ([RFC 6749](http://tools.ietf.org/html/rfc6749)) by sending a bearer token
+  ([RFC 6750](http://tools.ietf.org/html/rfc6750)) in the request's
+  Form-Encoded Body Parameter ([RFC 6750 Section 2.2](http://tools.ietf.org/html/rfc6750#section-2.2)
+  for POST requests and in the URI Query Parameter
+  ([RFC 6750 Section 2.3](http://tools.ietf.org/html/rfc6750#section-2.3)
+  for GET requests.
+
+  _The factory for this authorizer is registered as
+  `'simple-auth-authorizer:oauth2-access-token'` in Ember's container._
+
+  @class OAuth2
+  @namespace SimpleAuth.Authorizers
+  @module simple-auth-oauth2/authorizers/oauth2
+  @extends Base
+*/
+export default Base.extend({
+  /**
+    Authorizes an XHR request by sending the `access_token` property from the
+    session as a bearer token in the Form-Encoded Body Parameter or
+    URI Query Parameter:
+
+    ```
+    POST /resource HTTP/1.1
+    Host: server.example.com
+    Content-Type: application/x-www-form-urlencoded
+
+    access_token=mF_9.B5f-4.1JqM
+    ```
+
+    ```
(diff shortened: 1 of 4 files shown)


packages/ember-simple-auth-oauth2/lib/simple-auth-oauth2/authorizers/oauth2-access-token.js
```diff
@@ -0,0 +1,69 @@
...
+  authorize: function(jqXHR, requestOptions) {
+    var accessToken = this.get('session.access_token');
+    if (this.get('session.isAuthenticated') && !Ember.isEmpty(accessToken)) {
+      var data;
+
+      if (requestOptions.contentType && requestOptions.contentType.indexOf('application/json') > -1) {
+        data = requestOptions.data || '{}';
+        data = Ember.$.extend(JSON.parse(data), {access_token: accessToken});
+        requestOptions.data = JSON.stringify(data);
+      } else {
```
marcoow: This should probably check that the content type is actually "application/x-www-form-urlencoded". Maybe the if/else should actually be switched around so that it uses the request body when it's a "application/x-www-form-urlencoded" encoded `POST` request and the query strin otherwise.

marcoow:
Great PR! Maybe the behavior could be added to the existing OAuth 2.0 authorizer and be used as a fallback though (if it's possible to reliably detect when accessing headers isn't possible)?

kamal:
@marcoow I'm all for adding this to the existing authorizer as a fallback.

marcoow:
@kamal: :+1:

marcoow:
@kamal: Any updates? Would be great to have this in the next release!

kamal:
@marcoow no progress on my end, hopefully I can work on it this Friday

marcoow:
@kamal: great, really looking forward to merging this

bakura10:
+1, I'd love to have this too. I had to do something similar like that to avoid triggering preflight requests (which is damn expensive when you need to do tons of API requests).

alvarezloaiciga:
:+1: @kamal are you planning to work this week on it? I can help you with it if you want

kamal:
@alvarezloaiciga I'd appreciate some help. The biggest hurdle is detecting when we can't set headers on a request. One way is to tell when we are handed an `XDomainRequest` or a normal `XMLHttpRequest` (assuming the end-developer knows to use an `XDomainRequest` if doing CORS on a browser that doesn't support it)

marcoow:
@kamal, @alvarezloaiciga: this looks so promising but unfortunately I don't have time myself to help working on it. Basically the only things that's missing here is to make the behavior the fallback in the existing OAuth 2.0 authorizer if settings the `Authorization` header isn't possible for some reason.

Would love to merge this soon!

H3Chief:
@kamal I think I can help you with detecting when you can't set headers on a request.  We can determine if these headers can be set based on whether or not the client in use supports the `XMLHTTPRequest` level 2 spec, or is a XDomainRequest.

The property `withCredentials` only exists on level 2 compliant requests, so first, we can create a new request, and check the properties on it. (we never have to open the request here)

```
function getXHRMethod() {
  var xhr = new XMLHttpRequest();
  if ("withCredentials" in xhr) {
     return ''XMLHTTPRequest" ;
  } else if (typeof XDomainRequest != "undefined") {
     return "XDomainRequest";
  }
```


```

From one of my [favorite articles](http://www.html5rocks.com/en/tutorials/cors/) at [html5rocks.com](https://html5rocks.com)

PR #444, opened 2015-02-18
Update README with input type="submit"

davidpaulsson:
Changed the login form example with an Ember {{input type="submit" value="Login"}} instead of a `button` element. Otherwise the form can't be submitted using the `Enter` key.


marcoow:
Is that necessary for specific browsers only or so? It works as expected in the examples with Chrome.

marcoow:
@davidpaulsson: any updates on the question above?

davidpaulsson:
not really, no :) but there seem to be weird inconstancies using button type="submit" instead of an input type="submit" in my experience at least.

marcoow:
Ok, can you change it throughout the project then (examples, API docs etc.)?

PR #447, opened 2015-02-22
Fix potential typo in Example 4

NathanJang:
AuthenticationRoute should be named with PascalCase.


NathanJang:
Apparently not.

marcoow:
It is actually wrong but shouldn't be `App.AuthenticationRoute` but `App.LoginRoute`.

NathanJang:
Is this right?

marcoow:
yep, thanks

Issue #395, opened 2015-01-09
Default-Cookie-Experiation-Time only works 


GolexPrime:
Hi, 
I use the cookie store. I give the user the option to choose if he wants a session until the browser closes, or if he wants a 2 week session (expiration time: 1209600). This works fine if the user closes the browser just one time.

Let me explain:
After he closes the browser the first time, the cookie remains until he visits the application a second time. Because he doesnt have to log in, the expiration time wont be set to 2 weeks. If he closes the browser again the session (and the cookie) will be lost. 

Does the default-value of the expiration-time affect an existing cookie?

marcoow:
How do you implement that? I guess you're using the cookie session store for that?

GolexPrime:
We're using a custom authenticator. In the LoginController I if a certain checkbox is activated. If thats the case I execute:

this.set('session.store.cookieExpirationTime', 1209600);

Thats the only time I set the expiration-time (in the authenticate-action of the LoginController).

marcoow:
You need to set the expiration time back to its default value (or `null` if you want to use session cookies) whe the checkbox is unchecked.

GolexPrime:
Here is the authenticate-Function:

actions: {
		authenticate: function() {

			var _this 		        = this;
			var rememberMe 	= $("#remember-me");

			this._super().then(null, function(message) {
				_this.set('errorMessage', message);
			});

			if(rememberMe.prop("checked")) {
				this.set('session.store.cookieExpirationTime', 1209600);
			} else {
				this.set('session.store.cookieExpirationTime', null);
			}		
		}
	}

My Problem is that when I close the browser (after the login) and open it again the cookie still there. But when I close the Browser a second time, the cookie is gone.

marcoow:
Try setting the expiration time *before* you call `_super`.

GolexPrime:
Just tried it, still the same problem.

GolexPrime:
I think the problem is that the expiration-time of the cookie is set to default (null in my case) when the browser is opened again. But I may be wrong of course.

marcoow:
Can you post screenshots of Developer Tools -> Resources -> Cookies at the different points in time?

GolexPrime:
Ok, 

before I log in:
![bildschirmfoto 2015-01-12 um 10 03 49](https://cloud.githubusercontent.com/assets/4212661/5700740/b29637ee-9a42-11e4-8a9f-debd1ceef42f.png)

after I logged in:
[...]

PR #451, opened 2015-02-25
Fix Cookie-Expiration

msalahz:
I'm proposing the following fix for cookie expiration issue discussed in the following issue https://github.com/simplabs/ember-simple-auth/issues/395

I did the following to fix the issue : 
1- Caching cookie expiration lifetime in a new cookie with name [cookie name]:expiration-time.
2- Using cached cookie expiration for overriding session cookie in case session data changed.
3- Using cached cookie expiration for renew session cookie lifetime on session store ```read``` and ```persist```.

### DIFF
--- a/packages/ember-simple-auth-cookie-store/lib/simple-auth-cookie-store/stores/cookie.js
+++ b/packages/ember-simple-auth-cookie-store/lib/simple-auth-cookie-store/stores/cookie.js
@@ -96,6 +96,18 @@ export default Base.extend({
   _syncDataTimeout: null,
 
   /**
+    @property _renewExpirationTimeout
+    @private
+  */
+  _renewExpirationTimeout: null,
+
+  /**
+    @property pageVisibility
+    @private
+  */
+  pageVisibility: null,
+
+  /**
     @method init
     @private
   */
@@ -103,7 +115,9 @@ export default Base.extend({
     this.cookieName           = Configuration.cookieName;
     this.cookieExpirationTime = Configuration.cookieExpirationTime;
     this.cookieDomain         = Configuration.cookieDomain;
+    this.pageVisibility       = this.pageVisibilityInit();
     this.syncData();
+    this.renewExpiration();
   },
 
   /**
@@ -111,8 +125,8 @@ export default Base.extend({
     @private
   */
   readExpirationTime: function () {
-      var value = document.cookie.match(new RegExp(this.cookieName + ':expiration-time=([^;]+)')) || [];
-      return decodeURIComponent(value[1] || '');
+    var value = document.cookie.match(new RegExp(this.cookieName + ':expiration_time=([^;]+)')) || [];
+    return decodeURIComponent(value[1] || '');
   },
 
   /**
@@ -157,25 +171,11 @@ export default Base.extend({
   },
 
   /**
-    @method renew
-    @private
-  */
-  renew: function(data) {
-    data           = Ember.typeOf(data) === 'string' ? data : JSON.stringify(data || {});
-    var cachedExpirationTime = !!this.readExpirationTime() ? new Date().getTime() + this.readExpirationTime() * 1000 : null;
-    var expiration = !!this.cookieExpirationTime ? new Date().getTime() + this.cookieExpirationTime * 1000 : cachedExpirationTime;
-    this.write(data, expiration);
-  },
-
-  /**
     @method read
     @private
   */
   read: function() {
     var value = document.cookie.match(new RegExp(this.cookieName + '=([^;]+)')) || [];
-    if (!Ember.isEmpty(decodeURIComponent(value[1] || ''))) {
-        this.renew(decodeURIComponent(value[1] || ''));
-    }
     return decodeURIComponent(value[1] || '');
   },
 
@@ -189,7 +189,9 @@ export default Base.extend({
     var expires = Ember.isEmpty(expiration) ? '' : '; expires=' + new Date(expiration).toUTCString();
     var secure  = !!this._secureCookies ? ';secure' : '';
     document.cookie = this.cookieName + '=' + encodeURIComponent(value) + domain + path + expires + secure;
-    document.cookie = this.cookieName + ':expiration-time=' + encodeURIComponent(this.cookieExpirationTime || this.readExpirationTime()) + domain + path + expires + secure;
+    if(expiration !== null){ console.log('renewed to expire at : ' + new Date(expiration));
+      document.cookie = this.cookieName + ':expiration_time=' + encodeURIComponent(this.cookieExpirationTime || this.readExpirationTime()) + domain + path + expires + secure;
+    }
   },
 
   /**
@@ -206,5 +208,52 @@ export default Base.extend({
       Ember.run.cancel(this._syncDataTimeout);
       this._syncDataTimeout = Ember.run.later(this, this.syncData, 500);
     }
+  },
+
+  /**
+    @method pageVisibilityInit
+    @private
+  */
+  pageVisibilityInit: function(){
+    var stateKey, eventKey, keys = {
+      hidden: "visibilitychange",
+      webkitHidden: "webkitvisibilitychange",
+      mozHidden: "mozvisibilitychange",
+      msHidden: "msvisibilitychange"
+    };
+    for (stateKey in keys) {
+      if (stateKey in document) {
+        eventKey = keys[stateKey];
+        break;
+      }
+    }
+    return function(c) {
+      if (c) {document.addEventListener(eventKey, c);}
+      return !document[stateKey];
+    };
+  },
+
+  /**
+    @method renew
+    @private
+  */
+  renew: function(data) {
+    data           = Ember.typeOf(data) === 'string' ? data : JSON.stringify(data || {});
+    var cachedExpirationTime = !!this.readExpirationTime() ? new Date().getTime() + this.readExpirationTime() * 1000 : null;
+    var expiration = !!this.cookieExpirationTime ? new Date().getTime() + this.cookieExpirationTime * 1000 : cachedExpirationTime;
+    this.write(data, expiration);
+  },
+
+  /**
+    @method renewExpiration
+    @private
+  */
+  renewExpiration: function() {
+    var data = this.restore();
+    if (!Ember.isEmpty(data) && data !== {} && this.pageVisibility()) {
+      this.renew(data);
+    }
+    Ember.run.cancel(this._renewExpirationTimeout);
+    this._renewExpirationTimeout = Ember.run.later(this, this.renewExpiration, 60000);
   }
 });
--- a/packages/ember-simple-auth-cookie-store/lib/simple-auth-cookie-store/stores/cookie.js
+++ b/packages/ember-simple-auth-cookie-store/lib/simple-auth-cookie-store/stores/cookie.js
@@ -240,10 +240,12 @@ export default Base.extend({
     @private
   */
   renew: function(data) {
-    data           = Ember.typeOf(data) === 'string' ? data : JSON.stringify(data || {});
-    var cachedExpirationTime = !!this.readExpirationTime() ? new Date().getTime() + this.readExpirationTime() * 1000 : null;
-    var expiration = !!this.cookieExpirationTime ? new Date().getTime() + this.cookieExpirationTime * 1000 : cachedExpirationTime;
-    this.write(data, expiration);
+    if(!Ember.isEmpty(data)){
+      data           = Ember.typeOf(data) === 'string' ? data : JSON.stringify(data || {});
+      var cachedExpirationTime = !!this.readExpirationTime() ? new Date().getTime() + this.readExpirationTime() * 1000 : null;
+      var expiration = !!this.cookieExpirationTime ? new Date().getTime() + this.cookieExpirationTime * 1000 : cachedExpirationTime;
+      this.write(data, expiration);
+    }
   },
 
   /**
--- a/packages/ember-simple-auth-cookie-store/lib/simple-auth-cookie-store/stores/cookie.js
+++ b/packages/ember-simple-auth-cookie-store/lib/simple-auth-cookie-store/stores/cookie.js
@@ -215,7 +215,9 @@ export default Base.extend({
     @private
   */
   pageVisibilityInit: function(){
-    var stateKey, eventKey, keys = {
+    var stateKey;
+    var eventKey;
+    var keys = {
       hidden: "visibilitychange",
       webkitHidden: "webkitvisibilitychange",
       mozHidden: "mozvisibilitychange",
--- a/packages/ember-simple-auth-cookie-store/lib/simple-auth-cookie-store/stores/cookie.js
+++ b/packages/ember-simple-auth-cookie-store/lib/simple-auth-cookie-store/stores/cookie.js
@@ -3,196 +3,178 @@ import objectsAreEqual from 'simple-auth/utils/objects-are-equal';
 import Configuration from './../configuration';
 
 /**
-  Store that saves its data in a cookie.
-
-  __In order to keep multiple tabs/windows of an application in sync, this
-  store has to periodically (every 500ms) check the cookie__ for changes as
-  there are no events that notify of changes in cookies. The recommended
-  alternative is `Stores.LocalStorage` that also persistently stores data but
-  instead of cookies relies on the `localStorage` API and does not need to poll
-  for external changes.
-
-  By default the cookie store will use a session cookie that expires and is
-  deleted when the browser is closed. The cookie expiration period can be
-  configured via setting
-  [`Stores.Cooke#cookieExpirationTime`](#SimpleAuth-Stores-Cookie-cookieExpirationTime)
-  though. This can also be used to implement "remember me" functionality that
-  will either store the session persistently or in a session cookie depending
-  whether the user opted in or not:
-
-  ```js
-  // app/controllers/login.js
-  import LoginControllerMixin from 'simple-auth/mixins/login-controller-mixin';
-
-  export default Ember.Controller.extend(LoginControllerMixin, {
+ Store that saves its data in a cookie.
+ __In order to keep multiple tabs/windows of an application in sync, this
+ store has to periodically (every 500ms) check the cookie__ for changes as
+ there are no events that notify of changes in cookies. The recommended
+ alternative is `Stores.LocalStorage` that also persistently stores data but
+ instead of cookies relies on the `localStorage` API and does not need to poll
+ for external changes.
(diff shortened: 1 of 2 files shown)


packages/ember-simple-auth-cookie-store/lib/simple-auth-cookie-store/stores/cookie.js
```diff
@@ -156,15 +156,28 @@ export default Base.extend({
...
+  /**
+   @method write
+   @private
+   */
+  write: function (value, expiration) {
+      var path = '; path=/',
+          domain = Ember.isEmpty(this.cookieDomain) ? '' : '; domain=' + this.cookieDomain,
+          cachedExpiration = Ember.isEmpty(this.readExpiration()) ? '' : '; expires=' + new Date(this.readExpiration()).toUTCString(),
+          expires = Ember.isEmpty(expiration) ? cachedExpiration : '; expires=' + new Date(expiration).toUTCString(),
+          secure = !!this._secureCookies ? ';secure' : '';
```
marcoow: Please declare variables separately.

marcoow:
Seems like a good idea in general. The point is though that you'd also have to update the expiration when you're reading the cookie (the fact that the cookie is being read means that the session is active and thus the expiration should start running from that time on again).

Also this needs tests of course.

msalahz:
Sorry @marcoow, I didn't get your point, Can you give me an example?
and I will try to find time today to right tests for that.

marcoow:
The scenario is that you login to an application on, say, Jan 1. Let's say the cookie expires after 2 weeks so your session is valid until Jan 15th. If you open the app on Jan 3rd (which doesn't necessarily trigger a write of the cookie), the cookie should now be valid until Jan 17th.

msalahz:
Oh gotcha, But just to confirm, you want session restore to renew session lifetime with expiration time, Right?

marcoow:
Yes, I think that's the expected behavior - that all session activity renews the session lifetime.

msalahz:
I agree with you, I will proceed with changes required to that now.
Thank you you for your fast response.

msalahz:
Hi @marcoow , I have a problem I want to discuss with you regarding renewing the session lifetime.
Currently renewing session lifetime only happen if the user refresh the page or open new browser session during existing cookie lifetime but there is a case when the user successfully login and start interacting with the app and cookie expires during that which will for session expiration because there are no actions happened to renew session. Is there a way to handle this case ?

marcoow:
@msalahz: good point. That shouldn't happen of course but is also an extreme edge case. When we implement the renewing of expiration on read access, the only situation in which the described case could happen is when someone logs in and without triggering any operation on the session cookie (by reloading, opening a new page, saving data in the session etc.) stays active on the page. I would ignore that case for now as it's just so unlikely to happen - especially with regular expiration times of maybe 2 weeks or longer.

msalahz:
> When we implement the renewing of expiration on read access

What do you mean by that ? Do you mean that session renew should happen in read function ?

marcoow:
@msalahz: yes, as we discussed above the expiration time has to be updated when the cookie is read as well but the case your were describing (the cookies expires while someone is actively using the application) is just so unlikely to happen that I would ignore it for now.

msalahz:
@marcoow , If you checked my last commit you will find that I renewed session expiration in write function. I think renewing session in read function will cover the last point:

> there is a case when the user successfully login and start interacting with the app and cookie expires during that which will force session expiration because there are no actions happened to renew session

 because read function is excited each 500 ms by restore function which called in syncData functionan but this will force overriding both cookies each 500 ms and I don't like that. What you think ?

marcoow:
I think without updating the expiration in `read` as well your solution doesn't really work sufficiently. I'd add that as well, release a beta and see whether it's actually a problem in real applications. You'd just have to make sure that th `sessionDataUpdated` event isn't fired when only the cookie's expiration time changes. 

Also one of your commits seems to screw up fornatting, can you fix that?

msalahz:
Sure will fix it now, What you think  about triggering ``` sessionDataUpdated  ``` event each 1 min to renew session expatriation ? because ``` sessionDataUpdated  ``` will excite persist function.

msalahz:
@marcoow , I'm now updating the expiration in ```read``` as well but as I mentioned before updating the expiration in ```read``` will happen each 500ms because  ```read``` is excited in  ```restore``` which excited in  ```syncData```   which fires each 500ms, Do you think this will not have bad effects on application performance ?

marcoow:
I don't know whether updating the cookie's expiration time every 500ms will actually have a negative impact on real applications. That's why I said let's implement it first (as it would correctly have to be that way), release a beta or so and see whether it actually causes problems. Maybe you could also do some benchmarking to find how long it will take to update the cookie's expiration.

msalahz:
Sounds a good plan, Are there any remaining things I have to do to make my current changes ready for beta release ?

marcoow:
@msalahz: Yes, adding tests ;)

msalahz:
I don't have experience with writing tests but I will do my best and I hope you can review them ?

marcoow:
Sure, I'll do that. Checkout the existing specs for reference.

msalahz:
Does Mocha have internal support for creating promises or I have to required a third party ?

marcoow:
RSVP is used for promises - see the existing tests.

msalahz:
Please help: Why the following test always cuz the following error [Warning: PhantomJS timed out, possibly due to a missing Mocha run() call.]
```
 describe('#renew', function() {
    this.timeout(8000);

    beforeEach(function (done) {
      this.store.clear();
      this.store = Cookie.create();
      this.store.cookieName = 'test:session';
      this.store.cookieExpirationTime = 1; // 1 second
      this.store.persist({ key: 'value' });
    });

    it('cookie exists after renew', function(done) {
      this.store.renew();
      setTimeout(function() {
        expect(document.cookie).to.contain('test:session=%7B%22key%22%3A%22value%22%7D');
        done();
      },500);
    });

    it('cookie expires after cookie lifetime finish', function(done) {
      setTimeout(function() {
        expect(document.cookie).to.eq('');
        done();
      },1100);
    });
  });
```
Am I doing something terrible ?

marcoow:
You shouldn't use `setTimeout`, use `Ember.run.next` instead. Also I would run `grunt server` and then run the tests in the browser at http://localhost:8000/tests

msalahz:
http://localhost:8000/tests return Cannot GET /tests

msalahz:
I think ```Ember.run.later``` is what i need not  ```Ember.run.next```

marcoow:
`Ember.run.next` is the same as `Ember.run.later` with a delay of 1ms - you should try to use that.

The URL is [http://localhost:8000/test](http://localhost:8000/test) - sorry for the confusion.

msalahz:
I'm very frustrated right now I consumed all the day in one test suit and I only see this stupid error message : Error: timeout of 2000ms exceeded
I will push and see what will happen on Travis  build, 
Please check my tests and give me your comments ?

msalahz:
Hi @marcoow , I finally managed to right acceptable tests and it passed. Can you please review and let me know if you have any comments ?

Thanks in advance.

msalahz:
@marcoow , I got the chance today to test my changes in my project and I found the following case when the user login and left the browser tab open without any interactions which will allow cookie continues auto renew in ```read``` and the session will live forever as long as the browser session is opened and I think this doesn't make sense because user is not using the app and session should be expired after session timeout without any renew. I think we should handle this case and renew session on ```read``` only if browser tab is active and with 30 sec delay from last renew(in ```read```) because I still don't like renewing session each 500ms, What you think ?

lolmaus:
I've tried @msalahz's fix.

After figuring out how to make use of a custom `ember-simple-auth` package and removing my cookie hacks, i'm happy to say that the PR works for me fine.

It does keep me logged in  when i assign a number to `sesionStore.cookieExpirationTime` and vice versa.

I'm eager to see it merged (though i didn't review the code, i'm not familiar with `ember-simple-auth` internals). @msalahz, thank you for your effort!

PS @msalahz, why do you set an empty `ember_simple_auth:session:expiration-time` cookie when cookieExpirationTime is null? I think it would make more sense not to set the cookie in that case.

Also, i would rename the cookie to keep it consistent (dashes vs underscores).

marcoow:
@msalahz: looks good so far, thanks for your effort! You're right though regarding the edge case you describe. I'm not actually sure what to do about it though. I hope I can come up with a good idea over the weekend.

msalahz:
Good point @lolmaus , Will have a look on that. Thank you for your feedback.
Thank you @marcoow , What you think about the following proposal to handle this case :
> Renew session on read only if browser tab is active and with 30 sec delay from last renew(in ```read```)

msalahz:
@lolmaus , @GolexPrime , @snovity : Can you please share your thoughts about handling the following case ? 
> After appealing my recent changes I found the following case when the user login and left the browser tab open without any interactions which will allow cookie continues auto renew in read and the session will live forever as long as the browser session is opened and I think this doesn't make sense because user is not using the app and session should be expired after session timeout without any renew.

I'm proposing the following :
> Renew session on read only if browser tab is active and with 30 sec delay from last renew(in read)

marcoow:
@msalahz: sounds like a good plan actually. Of course it's a bit complex but it doesn't look like there's a simple solution. It would probably even be ok to increase the 30sec to a few minutes or so as it's not crucial for the cookie expiration time to be correct by a few minutes.

Detecting page/tab activity should be implemented using the [Page Visibility API](https://developer.mozilla.org/en-US/docs/Web/Guide/User_experience/Using_the_Page_Visibility_API).

msalahz:
@marcoow : Yes it's a bit complex but It worth trying. I think renew session each 1 min by default will be convenient and make it configurable will be awesome :) . What you think ? I think I can figure it out myself but I may need you help to guide me to do it configurable.

Sure I was planning to use Page [Visibility API](https://developer.mozilla.org/en-US/docs/Web/Guide/User_experience/Using_the_Page_Visibility_API) by following [this](http://stackoverflow.com/questions/19519535/detect-if-browser-tab-is-active-or-user-has-switched-away).

marcoow:
I don't think we need a configuration setting for that interval. You can just use 1min for now. I'd say the periodic updating of the expiration time should be startet in the store's `init` method and simply update the cookie's expiration if it's present.

msalahz:
@marcoow : So sorry for the delay, I had a very busy week and I didn't have any time to work on that. Will try to get it done this week.

msalahz:
@marcoow : New update are up and ready so please review and let me know what is your feedback?
@lolmaus : your comments are covered now.

lolmaus:
Thank you for your effort, @msalahz. I'm looking forward to having this merged. Unfortunately, i don't currently have a chance to test it again in a live project. Can someone please give it a shot?

packages/ember-simple-auth-cookie-store/lib/simple-auth-cookie-store/stores/cookie.js
```diff
@@ -165,6 +189,9 @@ export default Base.extend({
     var expires = Ember.isEmpty(expiration) ? '' : '; expires=' + new Date(expiration).toUTCString();
     var secure  = !!this._secureCookies ? ';secure' : '';
     document.cookie = this.cookieName + '=' + encodeURIComponent(value) + domain + path + expires + secure;
+    if(expiration !== null){ console.log('renewed to expire at : ' + new Date(expiration));
```
marcoow: leftover debug code

packages/ember-simple-auth-cookie-store/tests/simple-auth-cookie-store/stores/cookie-test.js
```diff
@@ -51,7 +51,45 @@ describe('Stores.Cookie', function() {
...
+      this.store = Cookie.create();
+      this.store.cookieName = 'test:session';
+      this.store.cookieExpirationTime = 60; // 1 min
+      this.store.expires = new Date().getTime() + this.store.cookieExpirationTime * 1000;
+      this.store.persist({key: this.store.expires});
+    });
+
+    it('session data cookie name should equal "test:session"', function() {
+      expect(this.store.cookieName).to.eq('test:session');
+    });
```
marcoow: this doesn't actually test the renew method, does it?

marcoow:
Thanks for the effort @msalahz! I'll check in the examples later.

packages/ember-simple-auth-cookie-store/lib/simple-auth-cookie-store/stores/cookie.js
msalahz: sure

packages/ember-simple-auth-cookie-store/lib/simple-auth-cookie-store/stores/cookie.js
```diff
@@ -181,5 +208,52 @@ export default Base.extend({
...
+    @method renewExpiration
+    @private
+  */
+  renewExpiration: function() {
+    var data = this.restore();
+    if (!Ember.isEmpty(data) && data !== {} && this.pageVisibility()) {
+      this.renew(data);
+    }
+    Ember.run.cancel(this._renewExpirationTimeout);
+    this._renewExpirationTimeout = Ember.run.later(this, this.renewExpiration, 60000);
```
marcoow: This might cause problems when testing - better wrap that with `if (!Ember.testing) {…`

packages/ember-simple-auth-cookie-store/tests/simple-auth-cookie-store/stores/cookie-test.js
msalahz: no it doesn't sorry. I removed "session data cookie name should equal "test:session"" and changed persist to renew.

packages/ember-simple-auth-cookie-store/lib/simple-auth-cookie-store/stores/cookie.js
msalahz: done

packages/ember-simple-auth-cookie-store/lib/simple-auth-cookie-store/stores/cookie.js
msalahz: done

packages/ember-simple-auth-cookie-store/lib/simple-auth-cookie-store/stores/cookie.js
```diff
@@ -181,5 +208,54 @@ export default Base.extend({
...
+  /**
+    @method renew
+    @private
+  */
+  renew: function(data) {
+    data           = Ember.typeOf(data) === 'string' ? data : JSON.stringify(data || {});
+    var cachedExpirationTime = !!this.readExpirationTime() ? new Date().getTime() + this.readExpirationTime() * 1000 : null;
+    var expiration = !!this.cookieExpirationTime ? new Date().getTime() + this.cookieExpirationTime * 1000 : cachedExpirationTime;
+    this.write(data, expiration);
+  },
```
marcoow: This should make sure that `data` isn't empty and don't write anything if not. Otherwise it would write an empty cookie.

packages/ember-simple-auth-cookie-store/lib/simple-auth-cookie-store/stores/cookie.js
```diff
@@ -181,5 +208,54 @@ export default Base.extend({
...
+    @method pageVisibilityInit
+    @private
+  */
+  pageVisibilityInit: function(){
+    var stateKey, eventKey, keys = {
+      hidden: "visibilitychange",
+      webkitHidden: "webkitvisibilitychange",
+      mozHidden: "mozvisibilitychange",
+      msHidden: "msvisibilitychange"
+    };
```
marcoow: Please always declare variables separately.

msalahz: done

packages/ember-simple-auth-cookie-store/lib/simple-auth-cookie-store/stores/cookie.js
```diff
@@ -96,6 +96,18 @@ export default Base.extend({
   _syncDataTimeout: null,
 
   /**
+    @property _renewExpirationTimeout
+    @private
+  */
+  _renewExpirationTimeout: null,
```
marcoow: for constistency don't prefix private members with "_".

packages/ember-simple-auth-cookie-store/lib/simple-auth-cookie-store/stores/cookie.js
```diff
@@ -181,5 +208,58 @@ export default Base.extend({
...
+      if (stateKey in document) {
+        eventKey = keys[stateKey];
+        break;
+      }
+    }
+    return function(c) {
+      if (c) {document.addEventListener(eventKey, c);}
+      return !document[stateKey];
+    };
+  },
```
marcoow: I guess I'm just too stupid but can you explain me how this works? How does it detect which one of `'visibilitychange'`, `'webkitvisibilitychange'` etc. to use? Also how does it detect when the visibility changes?

packages/ember-simple-auth-cookie-store/lib/simple-auth-cookie-store/stores/cookie.js
msalahz: I prefixed it with "_" because private _syncDataTimeout is prefixed with "_". Do you still want me to rename _renewExpirationTimeout to be renewExpirationTimeout ?

marcoow: Yeah, the library never uses "_" prefixes for private members so that should be consistent.

packages/ember-simple-auth-cookie-store/lib/simple-auth-cookie-store/stores/cookie.js
drewcovi: And is this going to break in IE9?

marcoow: If this just silently breaks (just doesn't periodically update the expiration) that's ok I'd say.

drewcovi: Ah! Didn't catch that this is just an enhancement.

msalahz: @marcoow : I'm following this [post](http://stackoverflow.com/questions/19519535/detect-if-browser-tab-is-active-or-user-has-switched-away/19519701#19519701) you can find all details you need with a demo for it.
@drewcovi : No, It will not break in IE9  and it's supporting IE10+

Issue #452, opened 2015-02-26
default oauth2-bearer authorizer doesn't work with torii


albertjan:
The torii provider returns content to the session containing a `authorizationCode` in most cases (accept for `facebook-connect` that uses `accessToken`). The oauth2-bearer authorizer looks for a property called `access_token`.

@marcoow Would you consider a PR adding a `torii-oauth2-autorizer` and maybe a `torii-fb-connect-authorizer`? Or one that catches both cases? The twitter oauth1 response token is in `oauth_token` I think. But I have no experience with that

marcoow:
In general, the authenticator and authorizer have to fit together as the authorizer uses data that the authenticator retrieves from some source. As the torii authenticator connects with various sources and doesn't actually know what the data it gets looks like, it's not actually possible to have a torii authorizer that works with all sorts of torii authenticators. That's the reason why no authorizers exist in the torii package.

Providing all sorts of authorizers for all sorts of torii providers is beyond the scope of the library. Ember Simple Auth provides a clear interface for creating these authorizers (which also requires almost no code).

albertjan:
Might be a good idea to add to the docs that you have to write your own authorizer when using torii.

marcoow:
Sure, PR welcome.

PR #453, opened 2015-02-26
add section about an authorizer to the readme

albertjan:
as talked about in #452

### DIFF
--- a/packages/ember-simple-auth-torii/README.md
+++ b/packages/ember-simple-auth-torii/README.md
@@ -34,6 +34,17 @@ the provider's name to the `authenticate` method of the session:
 this.get('session').authenticate('simple-auth-authenticator:torii', 'facebook-oauth2');
 ```
 
+## An authorizer
+
+To make sure the token you just got from torii is sent along with your ajax
+requests you need to add your own authorizer. The torii provider returns the
+token from the Provider#open function. This can be different per provider. So
+when you use several providers, you will have make sure you look at the right
+properties to find the token. Most providers also return their name you could
+use that to find out where the token is.
+
+Read this: https://github.com/simplabs/ember-simple-auth#implementing-a-custom-authorizer.
+
 ## Installation
 
 To install Ember Simple Auth Torii in an Ember.js application there are several
--- a/packages/ember-simple-auth-torii/README.md
+++ b/packages/ember-simple-auth-torii/README.md
@@ -36,14 +36,19 @@ this.get('session').authenticate('simple-auth-authenticator:torii', 'facebook-oa
 
 ## An authorizer
 
-To make sure the token you just got from torii is sent along with your ajax
-requests you need to add your own authorizer. The torii provider returns the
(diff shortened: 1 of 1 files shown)


marcoow:
I think the section you're adding is a little misleading. Authenticators and authorizers are somewhat bound together as the authorizer depends on data acquired by the authenticator. That's why e.g. the OAuth 2.0 authorizer doesn't work (while actually it might work but that would be coincidence) when using a torii authenticator as that acquires different data than the OAuth 2.0 authorizer expects.

Also, for torii that's especially difficult as even the torii authenticator doesn't know which data is acquired as that's specific for the actual torii provider it uses.

Third, the authorization mechanism your app uses isn't necessarily bound to the authentication mechanism (which could actually be many). So maybe you support Facebook, Google and Twitter authentication via torii and use your own authorization header that you inject into requests, meaning there is no one _"correct"_ torii authorizer.

So instead of adding the suggested session, I'd add sth, that simply says that there is no authorizer in the torii extension library but you need to implement your own depending on your authorization mechanism and that that authorizer must be able to handle the different types of authentication data the torii providers you're using acquire.

albertjan:
Ok I'll remove the example authorizer. And explain a bit more that there is no one implementation that will fit all use cases.

marcoow:
Hm, I'm not actually sure anymore that what you're describing (using e.g. the FB or Google token as a Bearer` token against one's API) is actually a use case of this extension library. Actually I would always extend the torii authenticator and post the authentication data to my own API to validate it and exchange it against a `Bearer` token that my own API maintains and then use that to authorize API requests.

albertjan:
@marcoow ah yes offcourse. We're using openid connect so we don't have the extra validation step. Should I close this PR?  Or should I add some more documentation about the rest of the auth flow. 

The point is I was really happy with the simple-auth torii integration but it took me a while to find out there were some parts missing, including the authorizer. So for openid connect the only thing you have to do is add a authorizer. But since openid connect hasn't really landed yet, this addition to the documentation seems a bit weird esp. from a normal oauth2 perspective. 

[This](https://github.com/Vestorly/torii/issues/148) is another piece of the puzzle that isn't implemented by default. I would probably be part of your extension of the torii provider.

marcoow:
Yes, correct `fetch` implementation is kind of lacking in torii still.

Can you maybe change the PR so that it says that the torii package is basically only a starting point for custom solutions as it cannot define an authorizer as it's unknown which exact torii providers will be used?

Sorry - I appreciate the effort, I'm just worried people might get lead in the wrong direction.

albertjan:
@marcoow Yeah sure I'll change it. And try to explain it a bit better. I don't mind, people should definitely not be led in the wrong direction :smile: no direction would be better in that case.

albertjan:
I've updated it added a hint to finishing the oauth flow and the `fetch` function. When new users get this far implementing their own authorizer will be a piece of cake :smile:

PR #460, opened 2015-03-11
Update auth token when password changes.

bravely:
This way, in the case of a user's account being compromised, a password change will render the attacker's token obsolete.


marcoow:
Good idea, but it wouldn't work in practice because Ember Simple Auth wouldn't be notified of the changed token and the next request after having changed the password would fail as the token in the `Authorization` header would still be the old one. So I think merging this might cause people seeing unexpected and hard to track down errors.

bravely:
Good thoughts. I'm gonna work on enabling this system to accept a new auth token on password changes to accompany this.

PR #461, opened 2015-03-12
Added information regarding UnauthenticatedRouteMixin

david-duncan:
It wasn't readily apparent to me how to prevent retrying transitions to /login even after Authentication. It wasn't until I dug into [`routeIfAlreadyAuthenticated`](http://ember-simple-auth.com/ember-simple-auth-api-docs.html#SimpleAuth-Configuration-routeIfAlreadyAuthenticated) that I realized I had to implement the UnauthenticatedRouteMixin. This should provide an easier quickstart and summary of the changes necessary to implement simple-auth

### DIFF
--- a/packages/ember-simple-auth/README.md
+++ b/packages/ember-simple-auth/README.md
@@ -87,6 +87,10 @@ export default Ember.Route.extend(AuthenticatedRouteMixin);
 This will make the route transition to `/login` (or a different URL if
 configured) when the session is not authenticated in the `beforeModel` method.
 
+To prevent a route from being accessed when the session is authenticated (which
+makes sense for the `login` route for example), use the
+[`UnauthenticatedRouteMixin`](http://ember-simple-auth.com/ember-simple-auth-api-docs.html#SimpleAuth-UnauthenticatedRouteMixin)).
+
 ### Authenticators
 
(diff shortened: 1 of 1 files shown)


marcoow:
Can you replace that with sth. like

```
To prevent a route from being accessed when the session is authenticated (which
makes sense for the `login` route for example), use the
[`UnauthenticatedRouteMixin`](http://ember-simple-auth.com/ember-simple-auth-api-docs.html#SimpleAuth-UnauthenticatedRouteMixin)).
```

instead? I think otherwise it gets a little too verbose for the README which is only supposed to provide a quick overview of the library.

marcoow:
Also please make sure there are no trailing spaces as the tests break otherwise.

david-duncan:
Good feedback indeed. I agree on verbosity.

marcoow:
There's still a trailing space somewhere (see failing test) - can you remove that?

david-duncan:
I believe I addressed that. It passed tests locally and on all of the build jobs except for: https://travis-ci.org/simplabs/ember-simple-auth/jobs/54155782

Looks like it stalled it out on cache. Is it possible to retry that build?

marcoow:
:+1:

PR #465, opened 2015-03-16
Add slightly clearer install instructions

mfeckie:
Make explicit the need to install the helpers when using ember-cli


marcoow:
That's already in the Addon install instructions that the README already links to.

mfeckie:
Whilst that's true, so are the config instructions below it.  Having those there also led me to think that this was all that was required.  Perhaps they could be removed?

marcoow:
Yeah, good point.

PR #467, opened 2015-03-17
Deprecations

marcoow:
This deprecates the `LoginControllerMixin`, `AuthenticationControllerMixin` and the `ApplicationRouteMixin`'s `invalidateSession` method as these don't really add any value anyways.

### TODO

- [x] remove the `ApplicationRouteMixin`'s `invalidateSession` method from the docs and examples.
- [x] deprecate the `ApplicationRouteMixin`'s `authenticateSession` method as well and add a new `sessionRequiresAuthentication` method instead.

### DIFF
--- a/docs/theme/main.hbs
+++ b/docs/theme/main.hbs
@@ -69,6 +69,11 @@
         top: -55px;
         height: 1px;
       }
+      span.label-deprecation {
+        float: left;
+        line-height: inherit;
+        margin-right: 5px;
+      }
     </style>
   </head>
   <body>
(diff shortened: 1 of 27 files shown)


lessless:
the documentation is outdated now, right? is there a built in way to invalidate session(logout)?

jpadilla:
@lessless `this.get('session').invalidate();` README seems up to date but I think http://ember-simple-auth.com is not.

marcoow:
Right, http://ember-simple-auth.com might be outdated - PR would be great. I'm currently spending all the time I have on #522.

PR #483, opened 2015-04-10
ember-simple-auth-oauth2 is not accessing the secure session introduced in #414

givanse:
Not really a PR (maybe in the future), more of a question about version 0.8.0-beta.1

When using the OAuth authenticator the `invalidateSession` action is not sending a request to the backend even though I have configured the route like this:

    ENV['simple-auth-oauth2'] = {
      serverTokenRevocationEndpoint: '/user/logout'
    };

So, I think that a fix similar to what this PR shows is needed. Is this correct?

Seems like the `invalidateSession` is not aware of the special key **secure**.

### DIFF
--- a/packages/ember-simple-auth-oauth2/lib/simple-auth-oauth2/authenticators/oauth2.js
+++ b/packages/ember-simple-auth-oauth2/lib/simple-auth-oauth2/authenticators/oauth2.js
@@ -183,20 +183,21 @@ export default Base.extend({
       resolve();
     }
     return new Ember.RSVP.Promise(function(resolve, reject) {
-      if (!Ember.isEmpty(_this.serverTokenRevocationEndpoint)) {
+      if (Ember.isEmpty(_this.serverTokenRevocationEndpoint)) {
+        success(resolve);
+      } else {
         var requests = [];
         Ember.A(['access_token', 'refresh_token']).forEach(function(tokenType) {
-          if (!Ember.isEmpty(data[tokenType])) {
+          var token = data[tokenType];
+          if (!Ember.isEmpty(token)) {
             requests.push(_this.makeRequest(_this.serverTokenRevocationEndpoint, {
-              token_type_hint: tokenType, token: data[tokenType]
+              token_type_hint: tokenType, token: token
             }));
           }
         });
         Ember.$.when.apply(Ember.$, requests).always(function(responses) {
           success(resolve);
         });
-      } else {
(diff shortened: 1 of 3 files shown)


marcoow:
The fix is correct but it only fixes the revocation request itself not the fact that the request isn't sent at all like you're saying (I have no idea why that would happen). Can you fix the specs as well?

givanse:
Ok, I'll work on this.

givanse:
I could have been clearer when I first posted this. The request is not made, because no token was found. The current change fixes it.

What I see now is that there isn't a test that guarantees that `session.invalidate` is sending the correct parameters to the actual `invalidate` implemented by the underlying authenticator.

marcoow:
@givanse: can you add a test for the case you're describing?

givanse:
Added this test `simple-auth-oauth2/session-test.js`, what do you think?

marcoow:
Hm, there is no OAuth 2.0 session actually. Session invalidation is already [tested here](https://github.com/simplabs/ember-simple-auth/blob/master/packages/ember-simple-auth/tests/simple-auth/session-test.js#L365) and the OAuth 2.0 authenticator's `invalidate` method is [tested here](https://github.com/simplabs/ember-simple-auth/blob/master/packages/ember-simple-auth-oauth2/tests/simple-auth-oauth2/authenticators/oauth2-test.js#L288).

Also I think `Session.invalidate` should not resolve with a value as it's kind of unclear what that value would be.

marcoow:
oops. didn't mean to close this…

givanse:
The point of the test that I added is to test both things working together.
Guarantee that doing a `session.invalidate()` will generate a request when
a revoke endpoint is configured.

I agree, resolving invalidate with a value is not a pretty solution. I just
made the test work with what I had.

So, what I'm suggesting is actually an integration test. How/where in the
project would you like to see this?

The initial problem I had is fixed. Maybe you'll prefer for me to undo the
second commit and the changes for an integration test can be discussed/done
in a different PR.

marcoow:
Yeah, don't think the concept of integration tests at that level make sense for the library per se as there are actually a lot of different combinations of authenticator/store/authorizer that couldn't be all tested in combination anyway. I'd say having good tests for the APIs that ensure that the components adhere to these APIs is sufficient.

givanse:
Ok, removed the test. This should be ready for merge then.

marcoow:
:+1:

PR #491, opened 2015-04-15
Added support for additional request data when authorizing with OAuth2 server

RobinvdVleuten:
Hi there,

I've added support for sending additional data to the OAuth2 server when requesting a token. This way the _ember-simple-auth_ library has support for auth SAAS'es like [Auth0](https://auth0.com/docs/protocols#9).


marcoow:
The OAuth 2.0 spec is pretty specific about what the request should look like. Please extend the OAuth 2.0 authenticator t add custom data.

PR #510, opened 2015-04-29
Add more context to snippets of code

pupeno:
As a newbie to emberjs, it would have saved me quite a bit of time if the snippets of code had more context to it. After figuring it out I added a bit, but I'm not 100% sure it's correct as I don't have auth working yet.

### DIFF
--- a/packages/ember-simple-auth-devise/README.md
+++ b/packages/ember-simple-auth-devise/README.md
@@ -137,12 +137,12 @@ In order to use the Devise authenticator (see the
 the application needs to have a login route:
 
 ```js
-App.Router.map(function() {
-  this.route('login');
+export default Router.map(function() {
+    this.route('login');
 });
 ```
 
-This route displays the login form with fields for `identification`,
+This route displays the login.hbs view with a form with fields for `identification`,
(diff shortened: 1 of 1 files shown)


packages/ember-simple-auth-devise/README.md
```diff
@@ -137,12 +137,12 @@ In order to use the Devise authenticator (see the
 the application needs to have a login route:
 
 ```js
-App.Router.map(function() {
-  this.route('login');
+export default Router.map(function() {
+    this.route('login');
```
marcoow: too many spaces

packages/ember-simple-auth-devise/README.md
```diff
@@ -137,12 +137,12 @@ In order to use the Devise authenticator (see the
...
 });
 ```
 
-This route displays the login form with fields for `identification`,
+This route displays the login.hbs view with a form with fields for `identification`,
```
marcoow: login.hbs is the template, not the view - I think the original text should be sufficient actually.

marcoow:
Don't think this is really necessary.

PR #514, opened 2015-05-01
Adding Session event "sessionRestorationSucceeded"

rbwsam:
Adding a new Session event that is triggered when authentication is restored (usually on app reload).

**Background:**
I am developing a Cordova + Ember iOS app. When the app is updated via the App Store or TestFlight the Ember app is reloaded the next time the iOS app is brought to the foreground. I need to know when the session is fully restored so that I can access the data in the store.

If there is a better way of achieving this that does not require a code change to this project I am all ears.

**Example initializer using event**
```js
export function initialize(container, application) {
  var session = container.lookup('simple-auth-session:main');
  // handle the session events
  session.on('sessionRestorationSucceeded', function() {
    console.log('RESTORED!');
  });
}

export default {
  name: 'test',
  initialize: initialize
};
```

### DIFF
--- a/packages/ember-simple-auth/lib/simple-auth/mixins/application-route-mixin.js
+++ b/packages/ember-simple-auth/lib/simple-auth/mixins/application-route-mixin.js
@@ -60,6 +60,7 @@ export default Ember.Mixin.create({
         'sessionAuthenticationSucceeded',
         'sessionAuthenticationFailed',
         'sessionInvalidationSucceeded',
+        'sessionRestorationSucceeded',
         'sessionInvalidationFailed',
         'authorizationFailed'
       ]).forEach(function(event) {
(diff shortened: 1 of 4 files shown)


jpadilla:
I can definitely see how this might actually be helpful

marcoow:
The problem is that the session is restored in 2 scenarios - once when the application starts up and also when the session content changes e.g. in another tab. The `sessionRestorationSucceeded` event would have to be triggered in both scenarios then, meaning your above code sample would probably not work as expected as `console.log('RESTORED!');` would potentially be called not only when the application just started up.

rbwsam:
@marcoow what would you suggest as a better implementation?

marcoow:
I'm not sure yet actually. It might also make sense to clearer split the 2 scenarios described above apart so that there would be less confusion.

marcoow:
With #522 session restoration will probably happen in the application route's `beforeModel` method, rendering sth. like a `sessionRestorationSucceeded` event unnecessary. Closing this as there will be no release before that anyway.

Issue #519, opened 2015-05-07
Make Ember Simple Auth an Ember CLI addon only


marcoow:
With Ember CLI being the official standard way for building Ember applications it doesn't make sense anymore to keep the globalized and AMD distributions as that adds a lot of overhead and also makes it hard to install the library in Ember Cli applications from github directly. Instead for the 1.0 release the library should use the Ember CLI addon in the main repository which would also make it much simpler to include Acceptance Tests in the main repository.

bradplank:
+1

rlivsey:
This would be great, I just noticed today that I was running an old version of simple-auth because I'd been updating the version in package.json but not in bower.json so it hadn't actually updated anything!

jpadilla:
:fireworks: that's great to hear, should simplify a few things on your side regarding maintainability

tzellman:
I'd be happy to help if needed.

piotrpalek:
@marcoow do you consider moving to the new initializers while doing this or should this be a seperate issue?
see: http://emberjs.com/deprecations/v1.x/#toc_access-to-instances-in-initializers

marcoow:
Yes, this would also switch to instance initializers.

piotrpalek:
Nice :+1:

miguelcobain:
I think that all the efforts for maintaining a globals version should go directly to this addon: https://github.com/ef4/ember-giftwrap

maysam-tayyeb:
+1

marcoow:
see #602

Issue #270, opened 2014-08-08
$.ajaxPrefilter hooking strategy too global?


jayphelps:
Since ember-simple-auth uses `$.ajaxPrefilter`, literally *any* calls that end up using `jQuery.ajax` will trigger the prefilter callback and run against any `Authorizer` defined.

This steps on the toes of other libraries and code paths that rely on jQuery's AJAX helpers, as well as makes it impossible to conditionally authorize certain models different ways (or selectively not authorize some).

Any alternatives?

ur5us:
@jayphelps with a cookie based approach your session data is send on every request as well so I tend to agree with @marcoow that after being authenticated the auth headers should be send on every request and it shouldn't matter whether it's EmberData or pure jQuery /cc @vladikoff. Moreover, if auth headers break your caching strategy, that's probably app specific, no. Just my 2 cents

lfridael:
Tying the library to JQuery feels restrictive.

What about apps that use https://github.com/visionmedia/superagent for Ajax operations? And what about future use cases where JQuery won't be available, such as during a Fastboot page render on the server?

marcoow:
Very true, especially with Ember becoming more and more independent of jQuery it would be cool to have a pluggable strategy in ESA. Since that this would most likely be a breaking change this should be done 1.0 though.

Suggestions welcome - I don't really have much time right now as I'm working on getting #522 ready.

kimroen:
There's some relevant discussion happening over in [the rfc-proposal for replacing ic-ajax with ember-fetch](https://github.com/ember-cli/rfcs/issues/19) in ember-cli.

fsmanuel:
What about adding a public function like [`shouldAuthorizeRequest`](https://github.com/simplabs/ember-simple-auth/blob/6190da91d645f01f6f8b9855720e82284cec8f8a/packages/ember-simple-auth/lib/simple-auth/setup.js#L65) that gives you the context and you can decide if authorization is required or not? The default should be `true` but you can overwrite it.

marcoow:
The problem is actually more about how request authorization is triggered at all. Currently the library uses `$.ajaxPrefilter` but when you're e.g. using the `fetch` API that simply doesn't exist so no requests would be authorized at all.

[...]

PR #522, opened 2015-05-08
Make Ember Simple Auth an Ember CLI addon only (ESA 1.0)

marcoow:
see #519 

This will be the 1.0 release of Ember Simple Auth, turning the library into a first class Ember CLI addon and removing the bower and globals distributions. The 1.0 version will of course also support Ember 2.0.

- [x] turn the session into a service (actually there's a service now that wraps the session which itself isn't directly exposed anymore)
- [x] split the initialization code into initializers and instance initializers
- [ ] make sure the documentation is correctly generated and convert all examples and references to Ember CLI/ES6 modules syntax
- [ ] move the setup of `$.ajaxPrefilter` into a dedicated initializer that can optionally be loaded (there should probably also be an initializer that e.g. sets up authorization on the Ember Data adapter instead). The current method of automatically loading it should be extracted into an addon so it's easy to upgrade projects in a non-breaking way.
- [ ] fix tests (currently fail in beta and canary because of switchfly/ember-mocha#48)

~~- [ ] rename `Authenticator.restore` to `Authenticator.validate`~~ (don't think this is actually a good idea)


fsmanuel:
@marcoow How can I help to push things forward?

marcoow:
@fsmanuel: maybe you could have a look at the documentation generation? See the `grunt docs` task on the master branch - that would basically have to be ported to Broccoli.

quaertym:
@marcoow Is this usable now?

marcoow:
@quaertym: this isn't ready yet

abradner:
I'm excited, this looks close! need any extra manpower anywhere?

marcoow:
@abradner: yeah, should hopefully be ready soon. I would really appreciate if someone would come up with a good proposal to tackle #270 as if that's ever done it should be done before the 1.0 release as it would most likely result in a breaking change…

harianus:
What is the estimated time for this? We want to integrate this soon!

marcoow:
first beta will be released next week latest

fotinakis:
Excited to see that this is in progress, just upgraded to Ember 1.13.1 and this is the last remaining piece for me to finish. Let me know if I can help beta test.

justinph:
I'm looking forward to this as well. Will be happy to help beta test when available. Thanks for the hard work on it, @marcoow.

marcoow:
see follow-up pull request #602

Issue #517, opened 2015-05-06
Deprecation warning in latest ember, since access to instances in initializers is deprecated


steffenbrem:
Hi,

In the latest ember version, you cannot access instances of a container in the intializer anymore, see: http://emberjs.com/deprecations/v1.x/#toc_deprecate-access-to-instances-in-initializers. Currently, this library triggers a lot of deprecation errors in latest ember version.

Any plans on updating simple-auth to support later ember versions?

jasich:
:+1:

marcoow:
PR would be great. Of course that would have to be backwards compatible and also work with Ember versions that don't have instance initializers.

EmergentBehavior:
:+1:

Keeo:
:+1:

mattma:
+1

rbwsam:
:+1:

jozefvaclavik:
:+1:

johnhamelink:
Yes please! :+1:

jzgit:
:+1:

vmalloc:
:+1:

hooddanielc:
:+1:

kaiwah:
:thumbsup:

GeeWizWow:
:+1:

felixkiss:
:+1:

charlatan154:
![](https://assets-cdn.github.com/images/icons/emoji/unicode/1f44d.png)

uppfinnarn:
:+1:

steffenbrem:
:+1:

quaertym:
We need more :+1:

steffenbrem:
@quaertym Yes indeed

engwan:
:+1:

akatov:
:+1:

mikejerome:
:+1:

AW-UC:
I guess the need to be backwards compatible makes this overly complicated. I just refactored my Simple-Auth-Custom-Session to use an Instance Initializer. It works nice, but a backwards compatible version would be a horrible code mess.

marcoow:
It needs to be backwards compatible. We cannot just force everyone to update to the latest ember.

chrisvdp:
But you can tell people on older Ember versions to use older simple-auth versions. [example](https://github.com/ef4/liquid-fire#ember-compatibility-table)

marcoow:
[...]

PR #532, opened 2015-05-22
Accessing container from initializer depricated, use instanceInitializer instead

Jakeii:
Since the container can no longer be accessed from an `initializer`, the initializer is pretty much useless so I replaced it with the new `instanceInitializer`. 

But the application has already been made ready by the time this is called so calling `application.deferReadiness()` in `setup` won't work anymore, so I've altered it to setup during the application route's `beforeModel` hook instead (this has the added benefit of allowing the applications loading state to show before trying to setup).

I'm not sure if it's the right approach but it works for me.

All the authenticators would need to be updated as well as the [ember-cli-simple-auth](https://github.com/simplabs/ember-cli-simple-auth), and probably all others that use initializer!

Solves #517

### DIFF
--- a/packages/ember-simple-auth/lib/simple-auth/ember.js
+++ b/packages/ember-simple-auth/lib/simple-auth/ember.js
@@ -1,5 +1,9 @@
 import initializer from './initializer';
+import instainceInitializer from './instaince-initializer';
 
 Ember.onLoad('Ember.Application', function(Application) {
   Application.initializer(initializer);
+  if(Application.instainceInitializer) {
+    Application.instainceInitializer(instainceInitializer);
+  }
 });
--- a/packages/ember-simple-auth/lib/simple-auth/ember.js
+++ b/packages/ember-simple-auth/lib/simple-auth/ember.js
@@ -1,9 +1,9 @@
 import initializer from './initializer';
-import instainceInitializer from './instaince-initializer';
+import instanceInitializer from './instance-initializer';
 
 Ember.onLoad('Ember.Application', function(Application) {
   Application.initializer(initializer);
   if(Application.instainceInitializer) {
-    Application.instainceInitializer(instainceInitializer);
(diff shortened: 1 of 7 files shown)


packages/ember-simple-auth/lib/simple-auth/ember.js
```diff
@@ -1,5 +1,9 @@
 import initializer from './initializer';
+import instanceInitializer from './instance-initializer';
 
 Ember.onLoad('Ember.Application', function(Application) {
   Application.initializer(initializer);
+  if(Application.instainceInitializer) {
```
steffenbrem: Typo

Jakeii:
`session` is now not being injected into controllers, routes etc. due to this bug https://github.com/emberjs/ember.js/issues/11172.
And perhaps even the unavailability of deferReadiness is a bug https://github.com/emberjs/ember.js/issues/11247

marcoow:
I don't think initializing the library inside of the application route's `beforeModel` method is the way to go. Also when using `reopen` what happens when the application itself implements the `beforeModel` method - does that or the `reopen`ed implementation take precedence?

OpakAlex:
user `reopen` not good for this case. It's wrong way.

marcoow:
Most of the initializers will be rewritten with #522. Also e.g. session restoration etc. will happen in the application route's `beforeModel` method. Maybe you can retarget this at the [`ember-cli-addon` branch](https://github.com/simplabs/ember-simple-auth/tree/ember-cli-addon)?

Jakeii:
Sorry I forgot about this, as it didn't work due to those bugs... But RE the beforeModel hook, as it's added in a instanceInitializer, it will take precedence over all others (unless perhaps added in another instanceInitializer..). I don't really like it either but I couldn't get it to work otherwise.

I'll take I look at the branch.

marcoow:
see #564, #522

Issue #523, opened 2015-05-12
Session data lost after refresh


bel3atar:
Hi,
I am writing some data to the session and it ends up lost after refreshing the page, or opening another tab.
To reproduce this:
```javascript
this.set('session.X', 'Y');
```
This code correctly alters the localStorage, but after a page refresh all is lost.

marcoow:
Is this when the session is authenticated? Please post the complete session content in `localStorage` here. Also which version of Ember Simple Auth are you using?

bel3atar:
I am using the latest ember-simple-auth and I'm doing authentication with torii's "facebook-connect". So basically localStorage is populated by torii.
After refreshing the page the session is still authenticated, but I lose the things I've added manually using the instruction of my previous comment.

marcoow:
What's the actual content of `localSorage` before and after the reload?

bel3atar:
before the reload, localStorage has
```json
{
   "secure":{
      "authenticator":"simple-auth-authenticator:torii",
      "userId":"...",
      "accessToken":"...",
      "provider":"facebook-connect"
   },
   "X":"Y"
}
```
after reload:
```json
{
   "secure":{
      "authenticator":"simple-auth-authenticator:torii",
      "userId":"...",
      "accessToken":"...",
      "provider":"facebook-connect"
   }
}
```

marcoow:
This should actually be working with 0.8.0-beta.1.

bel3atar:
sorry I made a mistake. accessToken and userId are wiped too,
so before refresh:
```json
{
   "secure":{
      "authenticator":"simple-auth-authenticator:torii",
      "userId":"...",
      "accessToken":"...",
      "provider":"facebook-connect"
   },
   "X":"Y"
}
```
and after refresh
```json
{
   "secure":{
      "authenticator":"simple-auth-authenticator:torii",
      "provider":"facebook-connect"
   }
}
```

marcoow:
Hm, looks like the provider's `fetch` method doesn't resolve with the correct data (see [invocation here](https://github.com/simplabs/ember-simple-auth/blob/master/packages/ember-simple-auth-torii/lib/simple-auth-torii/authenticators/torii.js#L41)). Probably the Ember Simple Auth code is wrong anyway as it shouldn't actually require the provider to resolve with the session data - will change that in the next 0.8.0 beta.

Kilowhisky:
I am getting the same issue but i am not using the torii library. 

[...]

PR #534, opened 2015-05-26
Fixed bug on main repo for issue#523: non-secure session data not bei…

Kilowhisky:
References this issue https://github.com/simplabs/ember-simple-auth/issues/523

Issue: non-secure session data was being wiped from the store and not restored when the `session.restore()` method was called. 

I altered a test to check its functionality as well. 

Only issue i could see happening here is that an event attached to session data being changed might fire twice now.

### DIFF
--- a/packages/ember-simple-auth/tests/simple-auth/session-test.js
+++ b/packages/ember-simple-auth/tests/simple-auth/session-test.js
@@ -162,12 +162,12 @@ describe('Session', function() {
 
         it('persists its content in the store', function(done) {
           var _this = this;
-
+          this.store.persist({ secure: { authenticator: 'authenticator' }, someOther: 'property' });
           this.session.restore().then(function() {
             var properties = _this.store.restore();
             delete properties.authenticator;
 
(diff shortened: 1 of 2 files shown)


packages/ember-simple-auth/tests/simple-auth/session-test.js
```diff
@@ -162,12 +162,12 @@ describe('Session', function() {
 
         it('persists its content in the store', function(done) {
           var _this = this;
-
+          this.store.persist({ secure: { authenticator: 'authenticator' }, someOther: 'property' });
```
marcoow: should be sufficient to only persist `someOther` here:

```js
this.store.persist({ someOther: 'property' });
```

Kilowhisky: That doesn't work. 

    this.session.restore().then() 

Is not called in that case

marcoow:
:+1:

kenips:
:+1: can we get a beta.3 from this? Thanks!

PR #535, opened 2015-05-26
Add instanceInitializer

quaertym:
This PR is to pave the way for fixing https://github.com/simplabs/ember-simple-auth/issues/517. I think it is logical to split the fix into multiple commits.

### DIFF
--- a/packages/ember-simple-auth/lib/simple-auth/ember.js
+++ b/packages/ember-simple-auth/lib/simple-auth/ember.js
@@ -1,5 +1,10 @@
 import initializer from './initializer';
+import instanceInitializer from './instance-initializer';
 
 Ember.onLoad('Ember.Application', function(Application) {
-  Application.initializer(initializer);
+  if (Application.instanceInitializer) {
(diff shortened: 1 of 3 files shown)


OpakAlex:
+1

quaertym:
@marcoow Any comment on this?

marcoow:
This would also need to make sure that the existing initializer doesn't use the container's `lookup` method when that's done in the instance initializer, otherwise you'd still get the deprecation notice. Also there's #532 which takes some more steps to solve the problem which I'm not sure are all actually necessary.

listepo:
+1

rafaelmezzomo:
+1

marcoow:
See https://github.com/simplabs/ember-simple-auth/pull/532#issuecomment-113643062 - would be cool to get something like this for the [`ember-cli-addon` branch](https://github.com/simplabs/ember-simple-auth/pull/532#issuecomment-113643062).

marcoow:
see #564, thanks for the effort!

Issue #543, opened 2015-06-01
routeEntryComplete not cleaned up between acceptance tests


kbullaughey:
The `routeEntryComplete` [local variable](https://github.com/simplabs/ember-simple-auth/blob/0a1120c0d324a3250eca9943ca04855855289aba/packages/ember-simple-auth/lib/simple-auth/mixins/application-route-mixin.js#L3) represents state leakage that doesn't get cleaned up between acceptance tests. On the very first acceptance test this flag gets set true, but then on subsequent acceptance tests it's still true, which means the target used for sending actions is the route rather than the transition and I get the error: 

> Error: Can't trigger action 'sessionInvalidationSucceeded' because your app hasn't finished transitioning into its first route. To trigger an action on destination routes during a transition, you can call `.send()` on the `Transition` object passed to the `model/beforeModel/afterModel` hooks.

Maybe this needs to be stored in something accessible to an initializer and set to false each time the application is created?

marcoow:
Right - PR would be great!

kbullaughey:
Submitted as #544.

I wasn't sure how to actually test this by writing a spec, so I just tested that the function can be called and then call it in the `setup()` which is called by the initializer. Would one need a dummy application an acceptance tests to test this? I didn't see any in the repo... I can confirm, however, in my own application that it solves my original problem.

marcoow:
This will be fixed by #522 when all this will be handled in the application route's `beforeModel` method.

PR #544, opened 2015-06-01
fix routeEntryComplete state leakage (#543).

kbullaughey:


marcoow:
I think you can remove the test as it doesn't really test anything anyways.

kbullaughey:
Removed the test.

packages/ember-simple-auth/lib/simple-auth/mixins/application-route-mixin.js
```diff
@@ -2,6 +2,11 @@ import Configuration from './../configuration';
 
 var routeEntryComplete = false;
 
+var resetApplicationRouteMixinInternalState = function() {
+  routeEntryComplete = false;
+  return 'reset';
```
marcoow: this can be removed

marcoow:
I think it's better to store this in a property of the application route that in the local variable anyway. That way we can also have a simple method in `setup.js` that unsets it and don't need the extra export in `application-route-mixin.js`. Sorry for the change in mind.

kbullaughey:
I think as a property on the application route there's no need to reset it during setup because the route object should be created fresh for each acceptance and unit test, no? 

I've pushed a new version to my branch.

packages/ember-simple-auth/lib/simple-auth/mixins/application-route-mixin.js
```diff
@@ -37,13 +35,18 @@ var routeEntryComplete = false;
   @extends Ember.Mixin
   @static
 */
-export default Ember.Mixin.create({
+var ApplicationRouteMixin = Ember.Mixin.create({
```
marcoow: Can you change this back to `export default` so it's consistent with the other files again as the extra var isn't needed anymore?

marcoow:
You're right of course and that makes it even simpler. Can you add a test?

kbullaughey:
Okay. Updated the tests and fixed the export statement.

packages/ember-simple-auth/tests/simple-auth/mixins/application-route-mixin-test.js
```diff
@@ -44,6 +44,7 @@ describe('ApplicationRouteMixin', function() {
       });
 
       it("translates the session's 'sessionAuthenticationSucceeded' event into an action invocation", function(done) {
+        expect(this.route.get('_authRouteEntryComplete')).to.be.true;
```
marcoow: Sorry for being so finicky, I actually really appreciate the PR, but this should have it's own test…

packages/ember-simple-auth/tests/simple-auth/mixins/application-route-mixin-test.js
```diff
@@ -52,6 +53,23 @@ describe('ApplicationRouteMixin', function() {
...
+        route2 = Ember.Route.extend(ApplicationRouteMixin, {
+          transitionTo: function() {}
+        }).create({ session: this.session });
+        var transition = { send: function() {} };
+        sinon.spy(transition, 'send');
+        sinon.spy(route2, 'send');
+        route2.beforeModel(transition);
+        this.session.trigger('sessionAuthenticationSucceeded');
+        Ember.run.next(this, function() {
+          expect(route2.get('_authRouteEntryComplete')).to.not.be.true;
```
marcoow: no need to assert on this here.

kbullaughey: Fixed.

packages/ember-simple-auth/tests/simple-auth/mixins/application-route-mixin-test.js
kbullaughey: No worries. Fixed.

marcoow:
:+1:

PR #560, opened 2015-06-19
Split the initialization into 2 different steps

juggy:
- initialization: registration/injection
- setup: session and store configuration and initialization

Setup is now done within the beforeModel of the application route as
suggested by https://github.com/emberjs/ember.js/issues/11247#issuecomment-113287227

Setup is executed one time only.

The registration and injection occurs correctly at standard init time
so it affects all controllers/routes/components

All tests pass.

Fixes https://github.com/simplabs/ember-simple-auth/issues/517

### DIFF
--- a/packages/ember-simple-auth-cookie-store/tests/simple-auth-cookie-store/configuration-test.js
+++ b/packages/ember-simple-auth-cookie-store/tests/simple-auth-cookie-store/configuration-test.js
@@ -25,25 +25,25 @@ describe('Configuration', function() {
 
   describe('.load', function() {
     it('sets cookieDomain correctly', function() {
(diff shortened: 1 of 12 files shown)


marcoow:
Thanks! All the development is happening in the [`ember-cli-addon` branch](https://github.com/simplabs/ember-simple-auth/tree/ember-cli-addon) now tough. I'd actually like to get rid of most of the initializer stuff really…

juggy:
Might be difficult to remove the initializers... with this code, you do not have to check both initializer and instanceInitializer so this is a plus.

Feel free to refactor this to match you ember-cli-addon branch.

marcoow:
see #564, thanks for the effort!

PR #564, opened 2015-06-22
ember-cli-addon - split initializer

Jakeii:
Hopefully this should be the last of the initializer pulls! Targets the ember-cli-addon branch.

As per https://github.com/emberjs/ember.js/pull/10256#issuecomment-113846021, it no longer relies on `deferRediness`, and incorporates most of the setup into the `ApplicationRouteMixin.beforeModel` method.

The `initializer` now only registers the stores/session, and `instanceInitializer` now only injects `session` into components, controllers, routes.

### DIFF
--- a/addon/mixins/application-route-mixin.js
+++ b/addon/mixins/application-route-mixin.js
@@ -57,22 +57,19 @@ export default Ember.Mixin.create({
   beforeModel(transition) {
     setup(this.container).finally(() => {
       this._super(transition);
-      if (!this.get('_authEventListenersAssigned')) {
-        this.set('_authEventListenersAssigned', true);
-        Ember.A([
-          'sessionAuthenticationSucceeded',
-          'sessionAuthenticationFailed',
-          'sessionInvalidationSucceeded',
-          'sessionInvalidationFailed',
-          'authorizationFailed'
-        ]).forEach((event) => {
-          this.get(Configuration.base.sessionPropertyName).on(event, Ember.run.bind(this, function() {
-            Array.prototype.unshift.call(arguments, event);
-            let target = routeEntryComplete ? this : transition;
-            target.send.apply(target, arguments);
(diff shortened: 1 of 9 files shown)


addon/mixins/application-route-mixin.js
```diff
@@ -54,23 +55,25 @@ export default Ember.Mixin.create({
...
-        this.get(Configuration.base.sessionPropertyName).on(event, Ember.run.bind(this, function() {
-          Array.prototype.unshift.call(arguments, event);
-          let target = routeEntryComplete ? this : transition;
-          target.send.apply(target, arguments);
-        }));
-      });
-    }
+    setup(this.container).finally(() => {
+      this._super(transition);
+      if (!this.get('_authEventListenersAssigned')) {
```
marcoow: Is this still necessary then?

Jakeii: Good point, no need to check if listeners are assigned as `ApplicationRoute.beforeModel` is only called once per page load.

Jakeii:
The `ApplicationRouteMixin` could be deprecated in favour of using `reopen` in the instance initializer to start setup and add the actions.

marcoow:
It would definitely be cool to deprecate the `ApplicationRouteMixin` but I don't think making `reopen` the _"official"_ way to get along without it is great - there should be a cleaner solution.

Jakeii:
Whats wrong with reopen? it's not a hack, it's used internally within Ember and referenced in the official guide and API docs.

marcoow:
It's basically the same as monkey patching in Ruby - it's very powerful and makes a lot of things very simple in the first place. At the same time though if used heavily it complicates stuff as you might be looking at some method that might be overridden somewhere else and then you wonder why it doesn't behave the way it should by looking at the original source code.

Issue #553, opened 2015-06-11
Session `secure` hash is cleared when testing


engwan:
Here's what I have in my test:

    authenticateSession();
    currentSession().set('something', 'Test');
    currentSession().set('secure.user_id', 1);

    visit('/');

And in my route's model hook:

    console.log(this.get('session.something')); // Correctly logs 'Test'
    console.log(this.get('session.secure.user_id')); // Undefined

Is there something I'm missing here?

marcoow:
You cannot currently set the contents of the secure part of the session manually in the test env. I guess that should be fixed by changing the test authenticator so that it [resolves with `data` from the `restore` method](https://github.com/simplabs/ember-simple-auth/blob/master/packages/ember-simple-auth-testing/lib/simple-auth-testing/authenticators/test.js#L5) - PR would be appreciated!

engwan:
I'm trying to trace the code and here's what I found out:

- `updateStore()` isn't being called when setting a "deep" property like `secure.user_id`
- I could call `updateStore()` manually after setting `secure.user_id` but somehow during the `visit('/')` call, it's getting overwritten because of a call to `session.authenticate` (Not sure what calls this) which sets the `secure` hash to the one resolved in `authenticate()` which is `{}` when using `authenticateSession()`.

Should we just allow something like:

    authenticateSession({user_id: 1});

Instead of:

    authenticateSession();
    currentSession().set('secure.user_id', 1);
    currentSession().updateStore();

This would entail passing the params of `authenticateSession()` to the authenticator as "credentials", then the test authenticator just resolves this data and this gets set into the secure hash.

What do you think?

marcoow:
`authenticateSession({user_id: 1});` sound like a good idea. Please make sure you base a PR on the [`ember-cli-addon` branch](https://github.com/simplabs/ember-simple-auth/pull/532#issuecomment-113643062).

tschoartschi:
I also need this functionality. Is it already available?

marcoow:
this will be releases with 1.0 - see #602 for instructions on how to use it already

tschoartschi:
@marcoow thanks for the information. Right now I don't have the possibility to change the whole auth-library of our app. I'll wait until ember-simple-auth is stable 1.0. For now it's ok to nest two andThen.

marcoow:
should be fixed in 1.0

PR #568, opened 2015-06-23
Authenticate tests with data

engwan:
Added option in test helpers to authenticate with session data as discussed in: https://github.com/simplabs/ember-simple-auth/issues/553


engwan:
Sorry wrong branch

PR #569, opened 2015-06-23
Authenticate tests with data

engwan:
Added option in test helpers to authenticate with session data.

Closes https://github.com/simplabs/ember-simple-auth/issues/553

### DIFF
--- a/tests/unit/authenticators/test-test.js
+++ b/tests/unit/authenticators/test-test.js
@@ -23,6 +23,12 @@ describe('Authenticators.Test', function() {
         done();
       });
     });
+
+    it('resolves with session data', function() {
+      return this.authenticator.authenticate({ userId: 1, otherData: 'some-data' }).then(function(data) {
+        expect(data).to.eql({ userId: 1, otherData: 'some-data' });
+      });
+    });
   });
 
   describe('#invalidate', function() {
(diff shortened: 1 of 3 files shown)


engwan:
Do I also have to add `data` to the restore function of the test authenticator?

I'm not sure when restore is being used / called and if it is used in tests.

tests/unit/authenticators/test-test.js
```diff
@@ -23,6 +23,14 @@ describe('Authenticators.Test', function() {
...
     });
+
+    it('resolves with session data', function(done) {
+      this.authenticator.authenticate({ userId: 1, otherData: 'some-data' }).then(function(data) {
+        expect(true).to.be.true;
+        expect(data).to.eql({ userId: 1, otherData: 'some-data' });
+        done();
+      });
+    });
   });
```
marcoow: you can change that to 

```js
it('resolves with session data', function() {
  this.authenticator.authenticate({ userId: 1, otherData: 'some-data' }).then(function(data) {
    expect(data).to.eql({ userId: 1, otherData: 'some-data' });
  });
});
```

By returning the promise you get rid of the `done` callback. Also the first expect isn't actually needed.

mphasize:
This looks exactly like what I could use right now... :+1:

PR #579, opened 2015-07-07
Add yuidoc generation to brocfile

fsmanuel:
Don't merge. We need to wait for ember-cli-yuidoc to make the options available (https://github.com/cibernox/ember-cli-yuidoc/issues/21).

Generate the docs: `npm run docs`


marcoow:
Thanks for the effort - I can tell these doc tasks are usually a pain ;) We'd also need to figure out how we actually want to document classes etc. as the class names that are used now (`SimpleAuth.Session` etc.) won't exist anymore in the future…

fsmanuel:
Oh yeah it was a pain.
We have to watch out for the new filename as well it's now `dist/docs/index.html` and i wonder how you treat the other packages (ember-cli-simple-auth-torii, etc.). They are all bundled in this addon, right?

marcoow:
Yes, there will also be one single bundle and only one API docs HTML.

fsmanuel:
What do you mean by "how we actually want to document classes"? As i see it, we should remove the `@namespace` and the box with "Standard AMD".

marcoow:
Right, that's what I meant. Just removing the `@namespace`s makes sense of course - guess I was just thinking too complicated…

fsmanuel:
There is an inconsistency with the imports in the examples and the ES6 import statements:
Examples: `ember-simple-auth/`
ES6: `simple-auth/`

marcoow:
`ember-simple-auth` will be the namespace for all the modules.

fsmanuel:
I'll open a new PR that uses `ember-simple-auth` and removes the `@namespace`

marcoow:
Should be fine to do everything in this I'd say.

fsmanuel:
What about the namespace of `Authenticators`, `Authorizers`, etc. Should they still exist?

marcoow:
The authenticators, authorizers and stores will be merged directly into the host application (see e.g. https://github.com/simplabs/ember-simple-auth/tree/ember-cli-addon/app/authenticators) and so shouldn't have a namespace as they'll actually be available under the app's own namespace.

fsmanuel:
got it!

fsmanuel:
Travis fails because of a timeout...

marcoow:
restarted the build manually…

fsmanuel:
@marcoow ready to merge.

marcoow:
@fsmanuel: how do I actually build the docs (I though `ember docs`)?

marcoow:
@fsmanuel: what's the command to run the docs generation?

fsmanuel:
@marcoow `npm run docs`

marcoow:
Ah, I see - thanks. Wouldn't it be nice to have a custom task like `ember docs`? Not sure whether that would be feasible though…

fsmanuel:
I will have a look at the ember commands and how `ember-cli-yuidoc` implements the `ember g ember-cli-yuidoc` command. I guess it should be easy to add `ember docs` that is an alias for `ember build --docs -e docs`. What I'm not sure about is if we expose that command to the host app...

marcoow:
It doesn't really seem clean to me to have a dedicated environment for docs. Is that actually necessary when there's a custom command instead?

fsmanuel:
I don't like the idea of a custom command because it would be exposed in every app that uses ember-simple auth and would probably fail in all that apps if invoked.
I used the environment to not generate docs on every build (e.g. Travis tests). I switched that to the way ember-cli-yuidocs does it [Code](https://github.com/cibernox/ember-cli-yuidoc/blob/master/index.js#L19).

marcoow:
OK, I see. The above was really more an unfiltered stream of thoughts I had about the topic - don't actually know much about custom tasks etc. in ember cli.

PR #588, opened 2015-07-22
Pass `content` all the way through Promise chain

kad3nce:
…making auth responses accessible to consumer code. Very helpful for e.g.
ember-cli-simple-auth-firebase, where the response from Firebase
contains a wealth of useful data.


marcoow:
thanks for the effort but this was already rejected in #560

PR #601, opened 2015-08-05
Switch AJAX request content type to JSON 

jbescoyez:
Switch AJAX request content type to JSON instead of 'x-www-form-urlencoded'.

In order to process data in Ember CLI Mirage as described in http://www.ember-cli-mirage.com/docs/v0.1.x/defining-routes/#dynamic-paths-and-query-params
For consistency purpose: all Ember-data requests are encoded in JSON
Code from: http://stackoverflow.com/questions/12693947/jquery-ajax-how-to-send-json-instead-of-querystring

### DIFF
--- a/addon/authenticators/devise.js
+++ b/addon/authenticators/devise.js
@@ -159,8 +159,9 @@ export default Base.extend({
     return Ember.$.ajax({
       url:        this.serverTokenEndpoint,
       type:       'POST',
+      data: JSON.stringify(data),
+      contentType: 'application/json; charset=UTF-8',
       dataType:   'json',
-      data,
       beforeSend(xhr, settings) {
         xhr.setRequestHeader('Accept', settings.accepts.json);
       }
--- a/tests/unit/authenticators/devise-test.js
+++ b/tests/unit/authenticators/devise-test.js
@@ -109,7 +109,8 @@ describe('Devise', () => {
         expect(args).to.eql({
           url:      '/users/sign_in',
           type:     'POST',
-          data:     { user: { email: 'identification', password: 'password' } },
+          data:     '{"user":{"password":"password","email":"identification"}}',
+          contentType: 'application/json; charset=UTF-8',
           dataType: 'json'
         });
         done();


jbescoyez:
FYI: The tests were failing before this PR.

marcoow:
I'm actually not sure this is so great at all - sorry for changing my mind so late. One problem is that it's a breaking change, the other is that the reason to do it (streamline it with the JSON API/Ember Data requests) isn't valid I think as authentication is really separate from loading resources etc. with Ember Data. Also I don't think Mirage is a great tool to use for mocking the authentication request. It's really built for maintaining a consistent set of fixture data in the mock server and what you need for mocking the authentication request is really just sth. that validates the credentials and responds with a token.

I'd rather not merge this, especially as it would probably break a lot of apps out there.

allthesignals:
I'm struggling with this as well, @jbescoyez. We are using jsonapi-resources gem for our backend to format everything according to the JSONAPI 1.0 spec. Is it not possible to configure the content-type header differently?

jbescoyez:
@allthesignals: I updated the plugin to make it optional. Maybe @marcoow will be ok to include it master? Otherwise, you can fork the repo and cherry-pick my commit.

jbescoyez:
Looks like the PR has not been updated. You can see it here: https://github.com/jbescoyez/ember-simple-auth/tree/patch-1 - I'm waiting for the feedback of @marcoow before sending a new PR.

marcoow:
As I say above this would probably break a lot of apps using devise on the server side so I'd rather not merge it. You can easily implement you own authenticator though of course.

jbescoyez:
@marcoow I think you are right. That's why I made it an opt-in option. Here is the commit: https://github.com/jbescoyez/ember-simple-auth/commit/5a3141e82080caa01f8d87df5f08b69081abbaf0

allthesignals:
@jbescoyez thank you - I was in the process of making it an option as well but ran out of time. @marcoow I agree, and I went ahead and wrote a custom authenticator. 

thank you for your work on this.

HeroicEric:
For anybody else that runs into this and isn't ready to customize their adapter, here's a function for converting the query string to a POJO:

```javascript
function queryStringToObject(query)  {
  let pairs = query.split('&');

  return pairs.reduce((obj, pair) => {
    let [key, value] = pair.split('=');
    obj[key] = decodeURIComponent(value);
    return obj;
  }, {});
};
```

You can use it like:

```javascript
this.post('/api/v1/token', (_schema, request) => {
  let { password } = queryStringToObject(request.requestBody);
});
```

Probably doesn't cover a lot of cases but works with a simple query string.

PR #605, opened 2015-08-07
Restructure test helpers

lan0:
This should update the test helpers to work with the new addon style. Everything seems to get loaded correctly.

I also added an acceptance test to test the behavior, but it currently fails with `Cannot call get with 'currentPath' on an undefined object.` - no idea why.

### DIFF
--- a/addon/initializers/setup-session.js
+++ b/addon/initializers/setup-session.js
@@ -1,7 +1,9 @@
 import Session from '../session';
 import Configuration from '../configuration';
+import TestAuthenticator from '../authenticators/test';
 
 export default function setupSession(registry) {
   registry.register('simple-auth-session:main', Session);
+  registry.register('simple-auth-authenticator:test', TestAuthenticator);
   registry.injection('simple-auth-session:main', 'store', Configuration.base.store);
 }
--- a/test-support/helpers/simple-auth.js
+++ b/test-support/helpers/simple-auth.js
@@ -1,22 +1,18 @@
-import Configuration from 'configuration';
+Ember.Test.registerAsyncHelper('authenticateSession', function(app, sessionData) {
+  var session = app.__container__.lookup('simple-auth-session:main');
+  session.authenticate('simple-auth-authenticator:test', sessionData);
+  return wait();
+});
 
-export default function() {
-  Ember.Test.registerAsyncHelper('authenticateSession', function(app, sessionData) {
-    var session = app.__container__.lookup(Configuration.session);
-    session.authenticate('simple-auth-authenticator:test', sessionData);
-    return wait();
-  });
+Ember.Test.registerHelper('currentSession', function(app) {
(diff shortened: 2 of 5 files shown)


test-support/helpers/simple-auth.js
```diff
@@ -1,22 +1,18 @@
-import Configuration from 'configuration';
+Ember.Test.registerAsyncHelper('authenticateSession', function(app, sessionData) {
+  var session = app.__container__.lookup('simple-auth-session:main');
```
marcoow: this should actually be

```js
const session = app.__container__.lookup('service:session');
```

tests/helpers/start-app.js
```diff
@@ -1,6 +1,7 @@
 import Ember from 'ember';
 import Application from '../../app';
 import config from '../../config/environment';
+import './simple-auth';
```
marcoow: This shouldn't be necessary as everything under test-support should be merged into the app tree anyway.

addon/initializers/setup-session.js
```diff
@@ -1,7 +1,9 @@
 import Session from '../session';
 import Configuration from '../configuration';
+import TestAuthenticator from '../authenticators/test';
 
 export default function setupSession(registry) {
   registry.register('simple-auth-session:main', Session);
+  registry.register('simple-auth-authenticator:test', TestAuthenticator);
```
marcoow: The Test authenticator should only be registered in test env. Is it actually possible to have the authenticator in `test-support` that would then only be loaded in test mode, e.g. `test-support/authenticators/test.js`?

tests/helpers/start-app.js
lan0: Without it, the tests fail with `ReferenceError: authenticateSession is not defined`

marcoow: Ah, I see - the helpers are only merged into the app tree but not auto-loaded.

lan0:
I'm not really sure if the test authenticator is the right way to do this, since the mock-API gets called and responds with `401` in the acceptance tests because there is no `Bearer` token.

If I authenticate the session in the helper with `authenticator:application`, the tests work - although `sessionData` for OAuth2 needs to be supplied (e.g. `authenticateSession({ identification: 'letme', password: 'in' });`.

Also, to access the mock API in the tests I needed to proxy `/api` and `/token` in `testem.json` and `ember serve` needs to be running in the background.

marcoow:
@lan0: when you want to make sure requests are authenticated as well you need to set up an authorizer for the tests as well - maybe just a test authorizer. I would recommend using Pretender though and never actually hit the backend from tests.

lan0:
Acceptance tests now work for us like done in https://github.com/simplabs/ember-simple-auth/commit/c5cb76d81b51d2187dfdfbbc5cc9237cb94b7775.

Thank you! :+1:

PR #609, opened 2015-08-11
Fixed asynchronous call of this._super (fixes #608)

### LINKED ISSUE #608: [jj-abrams] application route's beforeModel() not called
The `beforeModel(transition)` function on the application route is not being called, due to the `setup-session-restoration.js` instance-initializer reopening the route class and overriding beforeModel() without calling `this._super` correctly.

`this._super` is called asynchronously (see https://github.com/simplabs/ember-simple-auth/blob/jj-abrams/addon/instance-initializers/setup-session-restoration.js#L9), which is apparently not supported:

http://emberjs.com/blog/2014/03/30/ember-1-5-0-and-ember-1-6-beta-released.html#toc_ever-present-_super-breaking-bugfix

Also see this answer: http://stackoverflow.com/a/25902755

mdehoog:
See issue #608.

### DIFF
--- a/addon/instance-initializers/setup-session-restoration.js
+++ b/addon/instance-initializers/setup-session-restoration.js
@@ -1,13 +1,14 @@
-import Ember from 'ember';
-
 export default function setupSessionRestoration(instance) {
   const { container } = instance;
   const applicationRoute = container.lookup('route:application');
   const session = container.lookup('simple-auth-session:main');
   applicationRoute.reopen({
-    beforeModel(transition) {
-      const superCall = Ember.run.bind(this, this._super, transition);
-      return session.restore().then(superCall, superCall);
+    beforeModel() {
+      const superResult = this._super(...arguments);
+      const superReturn = function() {
+        return superResult;
+      };
+      return session.restore().then(superReturn, superReturn);
     }
   });
 }
--- a/addon/instance-initializers/setup-session-restoration.js
+++ b/addon/instance-initializers/setup-session-restoration.js
@@ -5,10 +5,7 @@ export default function setupSessionRestoration(instance) {
   applicationRoute.reopen({
     beforeModel() {
       const superResult = this._super(...arguments);
-      const superReturn = function() {
-        return superResult;
-      };
-      return session.restore().then(superReturn, superReturn);
+      return session.restore().then(() => superResult, () => superResult);
     }
   });
 }


marcoow:
Not sure about the semantics of this. The original idea was to have it so that it would only call super until the session was either restored successfully or not but of course you're right and that's not actually possible. Calling super while the session is still being restored though isn't so great really as the application route's `beforeModel` method might depend on session state…

mdehoog:
If `beforeModel` must depend on session state, the only way I can think of is if the session is restored elsewhere (maybe in an initializer).

Alternatively `beforeModel` could be documented as being called before restoring session state, and that people cannot depend on session state at that stage. Perhaps another function (eg `beforeModelAfterSessionRestore`) could be called after session restoration whose returned promise (if returned) is chained with the restore session promise so that it completes before the transition resumes.

marcoow:
Yeah, there would probably just have to be a session event for when restoration completes.

addon/instance-initializers/setup-session-restoration.js
```diff
@@ -1,13 +1,14 @@
...
   applicationRoute.reopen({
-    beforeModel(transition) {
-      const superCall = Ember.run.bind(this, this._super, transition);
-      return session.restore().then(superCall, superCall);
+    beforeModel() {
+      const superResult = this._super(...arguments);
+      const superReturn = function() {
+        return superResult;
+      };
+      return session.restore().then(superReturn, superReturn);
```
marcoow: I think this could just be written as `return session.restore().finally(() => superResult);`

mdehoog: If the `session.restore()` promise is returned from beforeModel(), and the session restore promise is rejected, the transition errors and does not complete. This means finally cannot be used.

`return session.restore().then(() => superResult, () => superResult);` works.

addon/instance-initializers/setup-session-restoration.js
```diff
@@ -1,13 +1,11 @@
...
   const { container } = instance;
   const applicationRoute = container.lookup('route:application');
   const session = container.lookup('simple-auth-session:main');
   applicationRoute.reopen({
-    beforeModel(transition) {
-      const superCall = Ember.run.bind(this, this._super, transition);
-      return session.restore().then(superCall, superCall);
+    beforeModel() {
+      const superResult = this._super(...arguments);
+      return session.restore().then(() => superResult, () => superResult);
```
marcoow: Won't `finally` work instead?


```js
return session.restore().finally(() => superResult);
```

mdehoog: If `finally` is used, and the session restore promise rejects, the transition is canceled because the rejected promise is returned by `beforeModel`. Handling the error state with `then` works.

marcoow: ah sure, you're right 👍

marcoow:
There's a simple workaround for the problem of async `_super` as pointed out by @stefanpenner here: https://github.com/emberjs/ember.js/issues/4632#issuecomment-120581786.

Can you submit another PR changing the code so that it uses that? It just makes the whole API a lot nicer when you can be sure that the session is fully restored when the application route's `beforeModel` hook ran.

mdehoog:
Done, see pull request #625. Nice find.

PR #611, opened 2015-08-12
Test failures when mixing Acceptance and Unit tests: Check for presence of secure data before accessing it

thriqon:
When unit tests and acceptance tests are "mixed", i.e. both are executed in the same page, a call to `Ember.$.ajax` fails if an acceptance tests was running before. This is a regression in 0.8.0.
The authorizer in use is Devise.

http://emberjs.jsbin.com/qinuka/edit?js,output

or as screenshot:

image

The code tries to call `get` on `undefined`, which is fetched from the session object. However, this session object has already been destroyed when the container from the last application instance was destroyed. Due to a glitch in Ember (I'm still investigating) instances of ObjectProxy lose their functionality when destroyed, and thus giving `session.get('secure')` an `undefined`, while `session.get('content.secure')` still works.

(introduced in https://github.com/simplabs/ember-simple-auth/commit/db4d635cfe40d2b2920f33939613c96778097aeb#diff-f4489d454b1136f52fda48e97daec5ddR70 in combination with https://github.com/danwang/ember-simple-auth/commit/668d5aaaf1ec070dba3240aa5190c572d33544ea)

Even though the bug above is easily solvable, one problem remains: Authorizers remain attached to `Ember.$.ajax` long after the application that created them has been destroyed. I'm not sure what kind of functionalities Ember provides to hook into the lifecycle though.

There are two obvious fixes to this bug (getting `content.secure` or asking for presence with an `if`), this PR just contains the one looking less ugly to me.

### DIFF
--- a/packages/ember-simple-auth-devise/lib/simple-auth-devise/authorizers/devise.js
+++ b/packages/ember-simple-auth-devise/lib/simple-auth-devise/authorizers/devise.js
@@ -66,12 +66,14 @@ export default Base.extend({
   },
 
   authorize: function(jqXHR, requestOptions) {
-    var secureData         = this.get('session.secure');
-    var userToken          = secureData[this.tokenAttributeName];
-    var userIdentification = secureData[this.identificationAttributeName];
-    if (this.get('session.isAuthenticated') && !Ember.isEmpty(userToken) && !Ember.isEmpty(userIdentification)) {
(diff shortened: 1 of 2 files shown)


marcoow:
Will release this as 0.8.1.

Be aware thought that authorizers (or at least the automatic registration of the `$.ajaxPrefilter`) will probably go away in 1.0.

thriqon:
Thanks for merging :)

pogopaule:
@marcoow any plans on a release date for 0.8.1?

PR #618, opened 2015-08-18
Cleanup Initializer

seanpdoyle:
When calling `simple-auth/configuration`'s `#load` function, the
`config` parameter is ignored.

Once it is removed from the parameters list, the call site (within the
initializer) no longer requires the `'./utils/get-global-config'` module
to be imported.


marcoow:
`config` argument is used in https://github.com/simplabs/ember-simple-auth/blob/master/packages/ember-simple-auth/lib/simple-auth/utils/load-config.js

PR #623, opened 2015-08-18
Move `app/service/session` to `addon`

seanpdoyle:
According to [Dockyard's Blog Post][dockyard] (section #2):

> A consumer should have the ability to easily extend your addon
> to do whatever they want. This means organizing your code a
> certain way.
> To provide this you should put all of your business logic into
> `addon/` and then include wrapper classes in `app/` that just `import`
> then `export` the extended class.

[dockyard]: https://dockyard.com/blog/2015/03/22/tips-for-writing-ember-addons

### DIFF
--- a/app/services/session.js
+++ b/app/services/session.js
@@ -1,35 +1,3 @@
-import Ember from 'ember';
+import SessionService from 'ember-simple-auth/services/session';
 
-const { computed, on }  = Ember;
-const { alias, oneWay } = computed;
-
-export default Ember.Service.extend(Ember.Evented, {
-  isAuthenticated: oneWay('session.isAuthenticated'),
-  data:            alias('session.content'),
-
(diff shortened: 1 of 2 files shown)


app/services/session.js
```diff
@@ -1,35 +1,3 @@
...
-    const session = this.get('session');
-    return session.invalidate.apply(session, arguments);
-  },
-
-  authorize(authorizerFactory, block) {
-    const authorizer = this.container.lookup(authorizerFactory);
-    authorizer.authorize(block);
-  }
-});
+export default SessionService.extend();
```
marcoow: this shouldn't `extend` but just return the class

seanpdoyle:
@marcoow :ok_hand:  ready for re-review

marcoow:
👍

PR #625, opened 2015-08-19
Restore session before running application route's beforeModel hook

mdehoog:
See issue #608 and pull request #609.

Inspired by https://github.com/emberjs/ember.js/issues/4632#issuecomment-120581786, thanks @marcoow for the suggestion.

### DIFF
--- a/addon/instance-initializers/setup-session-restoration.js
+++ b/addon/instance-initializers/setup-session-restoration.js
@@ -4,8 +4,8 @@ export default function setupSessionRestoration(instance) {
   const session = container.lookup('session:main');
   applicationRoute.reopen({
     beforeModel() {
-      const superResult = this._super(...arguments);
-      return session.restore().then(() => superResult, () => superResult);
+      const boundSuper = this._super.bind(this);
+      return session.restore().then(() => boundSuper(...arguments), () => boundSuper(...arguments));
     }
   });
 }


marcoow:
yay!

rlivsey:
This is causing my application route's beforeModel not to be called at all.

It does its thing, then comes to call super:

screenshot 2015-08-19 13 40 40

But then at this point `__nextSuper` is undefined for some reason so it continues without calling my original method:

screenshot 2015-08-19 13 40 58

Going back to the previous commit (3915a2b1a0d75a22e5d493e2b468f4537ad535eb) fixes the app.

swastik:
Can confirm this. It seems like the _super() call has to be invoked from the scope of the object.

At the moment, I've changed it to send two events (`sessionRestored` and `sessionRestorationFailed`) on the transition. Something like this:

screen shot 2015-08-19 at 7 11 50 pm

That works well for now. The beforeModel is executed, the session is restored, and it sends the right action before moving on to the other stuff. I can handle exceptional cases (like not being able to retrieve the currentUser) in the actions themselves.

rlivsey:
Would simply calling `apply` in the context of the route work vs `undefined` as it is now?

```javascript
boundSuper.apply(_this, _arguments)
```

swastik:
I'd tried that; it results in transpiled code something like this:

screen shot 2015-08-19 at 7 23 54 pm

... and the result is the same. The _super() isn't called. I found [this](http://stackoverflow.com/questions/24704256/issue-with-nextsuper) question. It seems that you have to invoke the _super() from the scope of the object.

marcoow:
Shouldn't `boundSuper(...arguments)` result in `boundSuper.apply(this, arguments)`?

swastik:
Ideally, yeah, but it applies `undefined` instead of `this` (and that's probably the reason it never fires the applicationRoute's beforeModel).

marcoow:
@rlivsey, @swastik: can you check whether a6de2fa14a79d819435b0616c27b7988b0a904d8 fixes the problem?

marcoow:
@rlivsey, @swastik: ab712a2a4637d3fc60de52d890c78ba5f7075b3d should fix it - at least the tests run.

swastik:
Thank you! :tada:

mdehoog:
Strange, my `beforeModel` is still being called with the previous code with `boundSuper`:

screen shot 2015-08-20 at 8 04 34 am
screen shot 2015-08-20 at 8 04 55 am

Thanks @marcoow for the fix. New code works for me as well. The transpiled code still results in an `undefined` `this` passed to `apply` but if it works for everybody then great!

screen shot 2015-08-20 at 8 08 12 am

rlivsey:
@mdehoog what version of Ember? My app's on 2.0.0

mdehoog:
I am using canary. Just tested with 2.0.0, and can confirm `beforeModel` is not called with the `boundSuper` method. Ember team must've fixed something.

rlivsey:
@mdehoog cool that explains things, there appear to have been a few changes to `this._super` since 2.0.0

marcoow:
I think this is actually more about how `reopen` works. While `_super` will point to the original method when `reopen`ing a class, when `reopen`ing an instance that just doesn't seem to work.

PR #633, opened 2015-08-22
allow the devise authenticator json to contain a root key

ssendev:
the devise authenticator has problems with nested attributes. i fixed it like this but when adding tests i noticed it should already be possible by specifying `Configuration.devise.identificationAttributeName = 'employee.email';` but the problem is that this causes the login request to send `user[employee.email]` and not `employee[email]` or `user[employee][email]`. solutions would be to do it like in this pull request and use `resourceName` during verification or to add separate 'identificationAttributeName' attributes for send and receive.

to make it clear what i am trying to do: the server expects to receive `user[email]=foo` and responds with `{user: {email: 'foo'}}`

i would tend to add separate attributes since that allows more flexibility, but you decide. maybe you also have a better idea.


marcoow:
The best solution for this is probably to simply extend a custom devise authenticator from the default one and override the respective method there.

ssendev:
this is what i am currently using and is very generic. if you have inerest i could es6ify it and update the pull request otherwise feel free to close it

```coffee
`import deviseAuthenticator from 'ember-simple-auth/authenticators/devise'`
`import config from '../config/environment'`

{ get, run, RSVP, isEmpty } = Ember

userAuthenticator = deviseAuthenticator.extend

  serverTokenEndpoint: config.apiHost + '/login'
  credentialNames: 'user.password': 'password', 'user.email': 'identification'
  requiredAttributes: ['data.attributes.api-key']

  validate: (data) ->
    for attribute in @get 'requiredAttributes'
      return false if isEmpty get data, attribute
    true

  restore: (properties) ->
    new RSVP.Promise (resolve, reject) =>
      if @validate properties
        resolve properties
      else
        reject properties

  authenticate: (credentials) ->
    new RSVP.Promise (resolve, reject) =>
      if typeof credentials == 'object'
        data = {}
        credentialNames = @get 'credentialNames'
        for own k,v of credentialNames
          currentData = data
          keys = k.split '.'
          lastKey = keys.pop()
          for key in keys
            currentData = currentData[key] ||= {}
          currentData[lastKey] = get credentials, v

        @makeRequest(data).then (response) =>
          if @validate response
            run null, resolve, response
          else
            run null, reject, response
        , (xhr) ->
          run null, reject, xhr.responseJSON || xhr.responseText
      else
        reject()


`export default userAuthenticator`
```

```coffee
`import BaseAuthorizer from 'ember-simple-auth/authorizers/base'`
`import config from '../config/environment'`
`import Ember from 'ember'`

{ isEmpty } = Ember
{ service } = Ember.inject
{ alias } = Ember.computed

UserApiTokenAuthorizer = BaseAuthorizer.extend

  session: service()

  tokenAttributeNames: api_key: 'data.attributes.api-key'

  authorize: (block) ->
    tokenAttributeNames = @get 'tokenAttributeNames'
    token = 'Token'
    attributeMissing = false
    for own attributeName, attributeKey of tokenAttributeNames
      attributeValue = @get "session.data.authenticated.#{attributeKey}"
      if isEmpty attributeValue
        attributeMissing = true
      else
        token += " #{attributeName}=\"#{attributeValue}\""
    unless attributeMissing
      block 'Authorization', token


`export default UserApiTokenAuthorizer`
```

PR #636, opened 2015-08-24
Fix ember 2.x compatibility

### LINKED ISSUE #616: ember-simple-auth dependency version of torii have deprecated code that won't work in ember 2.0.0
The current torii version in the dependency is old and contains code that was deprecated in the last versions of ember (1.13.X) and removed in ember 2.0.0. This causes deprecation warnings in 1.13 but crashes when using ember > 2.0.0. Is it possible to update the dependency version?

ismay:
Closes #616

### DIFF
--- a/package.json
+++ b/package.json
@@ -59,7 +59,8 @@
     "authorization"
   ],
   "dependencies": {
-    "ember-cli-babel": "^5.0.0"
+    "ember-cli-babel": "^5.0.0",
+    "torii": "^0.5.1"
   },
   "ember-addon": {
     "configPath": "tests/dummy/config"


ismay:
@marcoow So this should be sufficient?

marcoow:
I think it needs to be under `"dependencies"`, not `"devDependencies"`.

ismay:
Right, such a simple fix and I still messed it up :). Done.

marcoow:
👍

rlivsey:
Does this mean that every app which uses simple-auth also ends up bundling all the Torii JS (~25kB minified) even if we don't use it?

quaertym:
@rlivsey I have the same concern. Should this be another addon that consumes `ember-simple-auth`? It can be named as `ember-simple-auth-torii`.

rlivsey:
@quaertym other alternatives could be to document that if you want to use the Torii adapter then you need to add `torii` to your package.json manually, or for `ember install` to add it by default to the project's package.json so it can be removed if you don't need it.

quaertym:
@rlivsey Agree. @marcoow What do you think?

marcoow:
Hm, valid points. Would Ember CLI actually include code that's never used (e.g. modules that are never imported) in a production build though?

rlivsey:
@marcoow currently it will yes, in future there's talk on stripping out unused code (tree shaking etc…) but as it stands that's a way off AFAIK.

I've now also got deprecation warnings in my app from Torii even though I'm not using it.

marcoow:
@rlivsey: I think the best solution is then to simply document the need to add torii to the project manually. ESA could probably also log a warning if it recognizes the torii authenticator to be used without torii being present.

PR #641, opened 2015-08-25
Revert "Fix ember 2.x compatibility"

quaertym:
Reverts simplabs/ember-simple-auth#636

### DIFF
--- a/package.json
+++ b/package.json
@@ -59,8 +59,7 @@
     "authorization"
   ],
   "dependencies": {
-    "ember-cli-babel": "^5.0.0",
-    "torii": "^0.5.1"
+    "ember-cli-babel": "^5.0.0"
   },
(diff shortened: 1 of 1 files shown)


marcoow:
Keeping that open until docs have been added as well - no need to add that here though as the complete docs have to be rewritten anyway. I'd say there should be a warning though when the torii authenticator is used but torii is not installed.

quaertym:
I agree. So merge this?

marcoow:
Can you add an `Ember.Logger.warn` whenever the torii authenticator is used but torii isn't installed? That should probably go into the torii authenticator's `restore` and `authenticate` methods.

quaertym:
Closed in favor of  https://github.com/simplabs/ember-simple-auth/pull/674

PR #647, opened 2015-09-01
Prevent iOS crash in private browsing mode

knownasilya:
Prevent certain iOS Safari versions from crashing
in private browsing mode http://stackoverflow.com/a/14555361/483616

### DIFF
--- a/packages/ember-simple-auth/lib/simple-auth/stores/local-storage.js
+++ b/packages/ember-simple-auth/lib/simple-auth/stores/local-storage.js
@@ -43,7 +43,13 @@ export default Base.extend({
   */
   persist: function(data) {
     data = JSON.stringify(data || {});
-    localStorage.setItem(this.key, data);
+
+    // Prevent certain iOS Safari versions from crashing
+    // in private browsing mode http://stackoverflow.com/a/14555361/483616
+    try {
+      localStorage.setItem(this.key, data);
+    } catch(e) {}
+
     this._lastData = this.restore();
   },
 
--- a/packages/ember-simple-auth/lib/simple-auth/stores/local-storage.js
+++ b/packages/ember-simple-auth/lib/simple-auth/stores/local-storage.js
@@ -48,7 +48,9 @@ export default Base.extend({
     // in private browsing mode http://stackoverflow.com/a/14555361/483616
     try {
       localStorage.setItem(this.key, data);
-    } catch(e) {}
+    } catch(e) {
+      console.warn('Data not saved to localStorage due to limitations of this browser, see http://stackoverflow.com/a/14555361/483616');
+    }
 
     this._lastData = this.restore();
   },


marcoow:
I don't think this is a great solution as it simply swallows all errors and makes it really hard to debut why there's no session persistence when using local storage in iOS Safari in private browsing mode.

A better solution would probably to add sth. to the documentation that shows how to configure the store dynamically depending on what the browser offers or that shows how to extend the store to customize it and e.g. add error handling.

knownasilya:
> A better solution would probably **to add sth. to** the documentation

What are you trying to say here?

I could add a warning, if it was caught, that would be a better solution, since it wouldn't require the user to have to overwrite, or even waste hours trying to debug the issue (how long it took us).

marcoow:
I'm saying suppressing the error isn't a real solution to the problem as it just silently disables session persistence without anybody having a clue why it's not working - I guess that would be as hard to debug as the original problem in Safari you're solving with that.

knownasilya:
The issue is that the session wouldn't persist anyway, since the error happens when browsing in "private mode" (incognito), so the persistence is disabled.

This prevents your whole app from breaking on those devices. I can add a console log, so that it would be obvious what happened, and why. Would you merge that?

marcoow:
I think the best solution is to specifically handle quota exceeded errors which has to be done anyway. Not sure how these should be handled though and whether it's even possible to reliably identify them cross browser.

Thanks for the effort but I really think this needs more work and thought.

knownasilya:
I agree, but why not have an interim solution, is a broken app really something we don't want to prevent for users?

Docs would be good as well.

marcoow:
I think the point is just that you cannot always use the local storage adapter but maybe have to fall back to the cookie adapter. I think the first thing to do would be to make that clear in the docs so people are aware of the limitations.

It might also be an option to use sth. like http://www.jstorage.info (while I was against that initially it might actually make sense).

marcoow:
https://github.com/Wisembly/basil.js might be an alternative as well.

ivanvotti:
We recently had exactly the same problem with iOS 8 safari in private mode. It was pretty hard to find that the problem was in the local storage. Our app in the private mode looked as usual but its components didn't  react on gestures. Remote debugging didn't show any errors. Accidentally we found that if we change ESA session-store to ephemeral everything works even in private mode. I think @knownasilya looks better than nothing. It could save us a lot of time.

knownasilya:
@ivanvotti for my project I created a local store, and just copied everything over with my change.

marcoow:
I think the best way to fix this is to add a new store that uses sth. like https://github.com/Wisembly/basil.js which would wrap the functionality of deciding which store mechanism to use. This would be post 1.0 though as I don't want to postpone the 1.0 any longer if not necessary.

david-duncan:
@marcoow Looking into how to dynamically over to usings cookies based on this try catch block. Any insight as to how to this would be done?

marcoow:
@david-duncan: you register an initializer that runs before the `'simple-auth'` initializer and change the config if necessary.

david-duncan:
@marcoow that worked swimmingly. Do you want me to contribute any documentation for this?

marcoow:
I'll add docs for that for the 1.0 release. Adding a store that uses cookies as a fallback when `localStorage` isn't available is sth. for 1.1. then.

marcoow:
ESA will have an "adaptive" store that uses the `localStorage` store internally when `localStorage` is available and the cookie store if it is not: 443027bc0e553c720eb70dae3af7dd1aa73e8ad4

knownasilya:
Good to hear.

ivanvotti:
Awesome. Thank you.

PR #651, opened 2015-09-04
Assert that authorizer is set when using DataAdapterMixin

ivanvanderbyl:
This PR adds a simple assertion to alert the end developer that they need to add an `authorizer` in their adapter.

### DIFF
--- a/addon/mixins/data-adapter-mixin.js
+++ b/addon/mixins/data-adapter-mixin.js
@@ -30,10 +30,14 @@ export default Ember.Mixin.create({
   session: service('session'),
 
   ajaxOptions() {
+    const authorizer = this.get('authorizer');
     let hash = this._super(...arguments);
     let { beforeSend } = hash;
+
+    Ember.assert("You're using the DataAdapterMixin without specifying an authorizer. Please add `authorizer: 'authorizer:application'` to your adapter.", Ember.isPresent(authorizer));
+
     hash.beforeSend = (xhr) => {
-      this.get('session').authorize(this.get('authorizer'), (headerName, headerValue) => {
+      this.get('session').authorize(authorizer, (headerName, headerValue) => {
         xhr.setRequestHeader(headerName, headerValue);
       });
       if (beforeSend) {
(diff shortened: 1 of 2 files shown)


addon/mixins/data-adapter-mixin.js
```diff
@@ -30,10 +30,14 @@ export default Ember.Mixin.create({
   session: service('session'),
 
   ajaxOptions() {
+    const authorizer = this.get('authorizer');
     let hash = this._super(...arguments);
     let { beforeSend } = hash;
+
+    Ember.assert("You're using the DataAdapterMixin without specifying an authorizer. Please add `authorizer: 'authorizer:application'` to your adapter.", Ember.isPresent(authorizer));
+
```
marcoow: I think this should be done before anything else in this method, e.g. right after `const authorizer = this.get('authorizer');`.

tests/unit/mixins/data-adapter-mixin-test.js
```diff
@@ -39,6 +39,13 @@ describe('DataAdapterMixin', () => {
       expect(hash).to.have.ownProperty('beforeSend');
     });
 
+    it('asserts the presence of authorizer', () => {
+      adapter.set('authorizer', null);
+      expect(function() {
+        adapter.ajaxOptions();
+      }).to.throw(/You\'re using the DataAdapterMixin without specifying an authorizer. Please add `authorizer: \'authorizer:application\'` to your adapter\./);
```
marcoow: Just asserting that it throws is fine, otherwise the test has to be updated whenever the wording of the assertion message changes.

addon/mixins/data-adapter-mixin.js
ivanvanderbyl: What are your thoughts on defaulting `authorizer` to `authorizer:application`,  then checking with the container for whether it exists? If it doesn't, then assert that this needs customising?

marcoow:
can you rebase `jj-abrams`?

ivanvanderbyl:
@marcoow :shipit:

marcoow:
👍

PR #666, opened 2015-09-15
API documentation update & remove Configuration.base.session

trev:
Start injecting a bit of documentation love. I also removed Configuration.base.session as it seems it's no longer used. Please confirm before merging.


trev:
Hmm, I just noticed that you don't actually use the Yuidocs template but consume the information in a handlebars template. That means the nested properties & crossLink syntax of Yuidocs won't work. Why go through all the trouble to consume the documentation information via Handlebars when Yuidocs works just fine?

addon/mixins/application-route-mixin.js
```diff
@@ -60,11 +60,11 @@ export default Ember.Mixin.create({
   /**
     This action is triggered whenever the session is successfully
     authenticated. If there is a transition that was previously intercepted
-    by
-    [`AuthenticatedRouteMixin#beforeModel`](#SimpleAuth-AuthenticatedRouteMixin-beforeModel)
+    by 
+    {{#crossLink "Mixins.AuthenticatedRouteMixin/beforeModel:method"}}AuthenticatedRouteMixin#beforeModel{{/crossLink}}
```
marcoow: `Mixins` doesn't exist

marcoow:
Plain yuidoc didn't work with earlier versions of ESA. Also yuidoc's default template is horrible.

The custom handlebars task wouldn't  really work anymore neither though I think so the perfect solution would be to use regular yuidoc with a custom theme that looks more or less like the API docs generated for the current version.

trev:
Ok, well for now are you happy to just have the doc/comments updated as I go?

addon/mixins/application-route-mixin.js
trev: True but it was required because you were using `@namespace Mixins` in your yuidoc markup for your mixin classes. We can remove the `@namespace Mixins`.

marcoow: I'd say the `@namespace`s should be removed - this would of course require e.g. the `Base` authenticator to be renamed to `BaseAuthenticator` so it doesn't conflict with e.g. the `Base` authorizer (which would have to be renamed to `BaseAuthorizer`) etc.

Another solution would be to keep the namespaces but then there should be matching modules that export a namespace with all the correct members. Otherwise I think that will be confusing.

marcoow:
Yeah, let's just update the docs and then use a custom yuidoc theme to generate the HTML.

PR #669, opened 2015-09-15
Fix custom authenticator example in README

kevinansfield:
Imported variable had a different name than what was being extended in the custom authenticator example.


marcoow:
👍

PR #674, opened 2015-09-17
Warn when torii-authenticator is used but torii is not installed

quaertym:


addon/authenticators/torii.js
```diff
@@ -32,11 +32,16 @@ export default Base.extend({
     @public
   */
   restore(data) {
+    let torii = this.get('torii');
```
marcoow: should be `const`

addon/authenticators/torii.js
```diff
@@ -32,11 +32,16 @@ export default Base.extend({
...
+    if (Ember.isEmpty(torii)) {
+      Ember.Logger.warn('You are trying to use `torii-authenticator` but torii is not installed. Install torii using `ember install torii`.');
+    }
```
marcoow: Can you extract this into its own method (maybe `_assertToriiIsPresent`) as it's exactly the same as below?

Also this should use `Ember.assert` instead of checking and then logging.

marcoow:
great!

quaertym:
Good to merge.

marcoow:
this should really be extracted into its own method:

```js
assertToriiIsPresent() {
  const torii = this.get('torii');
  Ember.assert('You are trying to use `torii-authenticator` but torii is not installed. Install torii using `ember install torii`.', torii);
}
```

quaertym:
Done.

marcoow:
please use `Ember.assert`

quaertym:
@marcoow Done and also added tests.

marcoow:
👍

Issue #670, opened 2015-09-15
 Cannot read property 'authorize' of undefined TypeError: Cannot read property 'authorize' of undefined


gordonbisnor:
I'm getting an error that I have not been able to figure out:

```Cannot read property 'authorize' of undefined TypeError: Cannot read property 'authorize' of undefined```

Which is coming from ember-simple-auth/services/session.js:
```authorize: function authorize(authorizerFactory, block) {
      var authorizer = this.container.lookup(authorizerFactory);
      authorizer.authorize(block);
    }
```

This is my adapter:
```import ENV from 'hustle/config/environment';
import DS from 'ember-data';
import ActiveModelAdapter from 'active-model-adapter';
import DataAdapterMixin from 'ember-simple-auth/mixins/data-adapter-mixin';

export default ActiveModelAdapter.extend(DataAdapterMixin, {
	authorizer: 'authorizer:application',
	host: ENV.apiHost,
	namespace: "api/v1",
  buildURL: function(record, suffix) {
    return this._super(record, suffix) + ".json";
  }
});
```
For some reason ```this.container.lookup(authorizerFactory);``` is coming up undefined, and then the next line is calling ```authorize()``` on undefined.

*I'm using:*
- Ember Data 2.0.0
- Ember 2.0.0
- "ember-simple-auth": "simplabs/ember-simple-auth#jj-abrams"

Any ideas?

marcoow:
Did you define the application authorizer in `app/authorizers/application.js`?

gordonbisnor:
I hadn't. And thank for the incredibly fast response!

corrspt:
@marcoow I also had this problem but I was able to figure it out by comparing the dummy app that you usually link to in the issues.

Seems as I wasn't the only that tripped on the subject but considering that this addon is in a "transition" state (with the whole jj-abrams branch), what would be the best way to document this situation right now? I'm asking this in terms of how I could help out (with a documentation PR, or something else).

Best regards and thanks for the addon, I'm currently using it to create a new Ember App!

marcoow:
@corrspt: this will eventually be covered in the README of course. If you want to go ahead and submit a PR that updates the authorization section accordingly that would be great of course!

corrspt:
Ok, I'll try to help! I assume this would a be PR on the jj-abrams branch?

marcoow:
exactly

PR #675, opened 2015-09-17
Update README with documentation for ESA 1.0

corrspt:
Trying to help with the documentation. 

I've seen @marcoow Ember Munich meetup video saying that the 1.0 release is basically pending due to the documentation (and since I struggled these last few days with some issues, I thought I might help).

Any input is appreciated. Hope it helps


README.md
```diff
@@ -285,6 +288,26 @@ export default BaseAuthorizer.extend({
...
+
+export default Base.extend({
+  authorize(block) {
+    const accessToken = this.get('session.authenticated.token');
+    if (this.get('session.isAuthenticated') && !isEmpty(accessToken)) {
+      block('Authorization', `Bearer ${accessToken}`);
+    }
+  }
+
+});
```
marcoow: This authorizer is already part of Ember Simple Auth.

README.md
```diff
@@ -267,6 +267,9 @@ export default DS.JSONAPIAdapter.extend(DataAdapterMixin, {
 });
 ```
 
+**If you set the authorizer to authorizer:application** you need to create an `authorizers/application.js` inside your application (app/authorizers/applications.js). See the following section on implementing a custom authorizer.
+
```
marcoow: The README should focus on how things work in general rather than a specific case. Custom authorizers can be defined in `app/authorizers` while the specific name of these authorizers are up to the developer.

README.md
corrspt: Hum, interesting. You're right, I didn't find it while implementing my own solution, so I just rolled my own. Although there's something curious (I will have to test it to make sure it works) because:

I had to use this
```js
const accessToken = this.get('session.authenticated.token');
```
instead of the default on "oauth2-bearer.js
```js
const accessToken = this.get('session.authenticated.access_token');
```
Otherwise the token would never get sent, because `access_token` would always be undefined

README.md
corrspt: I see what you're saying and I agree, but the issue I had (and It could be simply because I don't have that much experience with Ember in general) was that I read here in one of the issues (or possibly seeing the dummy app) that you needed to set `authorizer: 'authorizer:application'`. 

But then I went and did that and of course then I got an error of undefined (issue #670). So I thought that for people like me this reference would be a good example to not trip on this. Because effectively when you set the authorizer to something it expects to find a file somewhere. That was the reasoning for adding that line.

corrspt: Any comments on my replies @marcoow ? I want to help, although I maybe failing at that :)

marcoow: I think for now the most important thing is to get the [How does it work?](https://github.com/simplabs/ember-simple-auth/blob/jj-abrams/README.md#how-does-it-work) and [How do I use it](https://github.com/simplabs/ember-simple-auth/blob/jj-abrams/README.md#how-do-i-use-it) sections right. Would be great if you could review them.

Especially the [How do I use it](https://github.com/simplabs/ember-simple-auth/blob/jj-abrams/README.md#how-do-i-use-it) section should really cover the steps 80% of the users would have to take to set up a fairly standard integration which it currently doesn't really.

corrspt:
Sorry for not having the time to contribute more with this PR. Great work with the project, hope to be able to help sometime in the future!

marcoow:
@corrspt: would be great - there's still lots of work left

PR #676, opened 2015-09-18
Update cookie.js

bttf:
Update injection name to match what is used in `setup-session-stores` initializer


marcoow:
This will completely be rewritten for 1.0.0.

marcoow:
sorry, actually still makes sense to merge this of course

Issue #678, opened 2015-09-20
The mechanism used to register ESA library version doesn't work with a parent addon


bquenin:
Hi,

When I develop my own ember-cli addon and depend on ESA, it failed to build with an unresolvable resource coming form the index.js.
After investigating, it seems that the "vendor" folder is not available to the parent addon and thus, ESA library version registration fails.

marcoow:
What's the specific error you're seeing?

bquenin:
Hi, this is the error I'm getting:
```
version: 1.13.8
Building...
Cannot read property 'import' of undefined
TypeError: Cannot read property 'import' of undefined
    at Class.module.exports.included (/Users/bq/workspace/pro-frontend/node_modules/ember-cli-common-frontend/node_modules/ember-simple-auth/index.js:9:13)
    at /Users/bq/workspace/pro-frontend/node_modules/ember-cli/lib/models/addon.js:253:32
    at Array.map (native)
    at Class.eachAddonInvoke (/Users/bq/workspace/pro-frontend/node_modules/ember-cli/lib/models/addon.js:251:22)
    at Class.Addon.included (/Users/bq/workspace/pro-frontend/node_modules/ember-cli/lib/models/addon.js:357:8)
    at EmberApp.<anonymous> (/Users/bq/workspace/pro-frontend/node_modules/ember-cli/lib/broccoli/ember-app.js:328:15)
    at Array.filter (native)
    at EmberApp._notifyAddonIncluded (/Users/bq/workspace/pro-frontend/node_modules/ember-cli/lib/broccoli/ember-app.js:323:45)
    at new EmberApp (/Users/bq/workspace/pro-frontend/node_modules/ember-cli/lib/broccoli/ember-app.js:118:8)
    at module.exports (/Users/bq/workspace/pro-frontend/ember-cli-build.js:5:13)
```

marcoow:
Please check whether the problem persists with the latest revision of the `jj-abrams` branch.

bquenin:
Now I get this:
```
version: 1.13.8
Building...
undefined is not a function
TypeError: undefined is not a function
    at Class.module.exports.included (/Users/bq/workspace/pro-frontend/node_modules/ember-cli-common-frontend/node_modules/ember-simple-auth/index.js:10:15)

```

marcoow:
This looks more like a bug in Ember CLI actually…

bquenin:
you're right, it seems that "included" is called twice for some reason.

[...]

PR #679, opened 2015-09-20
Jj abrams

bquenin:
This is a fix for #678. It basically register the library version on the app initializer instead of relying on an external file.


marcoow:
This doesn't work as it till register the library after the installed libraries and versions have been printed to the log so that ESA will never appear in that list

PR #687, opened 2015-09-28
Fixed issue where the lookup for the torii library was failing

ghurlman:
With Torii 0.6.0, the lookup for the torii module in the initializer was failing, returning null. This simple PR fixes the lookup.


marcoow:
There will be no more release based on the current master branch. It will be replaced by the `jj-abrams` branch which is to be released soon.

PR #689, opened 2015-09-28
Update test authenticator to resolve with data in restore method

andrei1089:
While writing some acceptance tests I've noticed that calling `authenticateSession` before visiting any route resulted in losing the `data` passed to `authenticateSession`.

This happened because the `restore` method of the test authenticator was not returning any data and `session.restore()` was overwriting session data previously set using `authenticateSession` with the value returned from the authenticator (`undefined` in this case).


marcoow:
👍

PR #705, opened 2015-10-07
Fix oauth2 authenticate data object

dustinfarris:
I think this was typoed in https://github.com/simplabs/ember-simple-auth/commit/300edb530bdc3dee836c13e186c962b018a03681#diff-931dc51edd167948e773858f72a47233L137

### DIFF
--- a/addon/authenticators/oauth2-password-grant.js
+++ b/addon/authenticators/oauth2-password-grant.js
@@ -143,7 +143,7 @@ export default BaseAuthenticator.extend({
   */
   authenticate(identification, password, scope = []) {
     return new RSVP.Promise((resolve, reject) => {
-      const data                = { 'grant_type': 'password', username: identification, password };
+      const data                = { 'grant_type': 'password', username: identification, password: password };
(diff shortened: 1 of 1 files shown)


dustinfarris:
Actually this isn't quite right yet.. one more commit coming.

dustinfarris:
Actually skip this.  Need to know whether authenticate should accept and object as in the examples, or 2 args as in the implementation.  Either way it needs to be fixed.

pichfl:
@dustinfarris You might want to read this, though: https://babeljs.io/docs/learn-es2015/#enhanced-object-literals

dustinfarris:
@pichfl this was causing problems with authentication where the ajax call was POSTING

```
username[identification]: "myusername",
username[password]: undefined
```

maybe my fix isn't right, but this is definitely broken.

dustinfarris:
@pichfl ok i see what you're saying.  so this wasn't a typo after all.

i think the conflict of passing an object vs 2 args is still a problem though.

PR #602, opened 2015-08-05
ESA 1.0

marcoow:
[Build Status](https://travis-ci.org/simplabs/ember-simple-auth)

_(this was previously #522 and the `ember-cli-addon` branch but as 1.0 will actually include more changes than just making ESA an Ember CLI Addon-only project the branch was renamed to `jj-abrams` following a nice [ember tradition](https://github.com/emberjs/data/tree/jj-abrams))_

This will be the 1.0 release of Ember Simple Auth, turning the library into a first class Ember CLI addon and removing the bower and globals distributions. The 1.0 version will of course also support Ember 2.0.

__For more info on Ember Simple Auth 1.0 see the [talk @marcoow gave at the September Ember.js Munich Meetup](https://www.youtube.com/watch?v=aeH_9YGgt9E).__
## Installation Instructions

- remove the `ember-simple-auth` package from `bower.json`
- remove and `ember-cli-simple-auth*` packages from `package.json`
- add the new Ember CLI Addon to `package.json`:  `"ember-simple-auth": "simplabs/ember-simple-auth#jj-abrams"`
- run `bower install && npm install`

See the [dummy app](https://github.com/simplabs/ember-simple-auth/tree/jj-abrams/tests/dummy) for example usage of 1.0 ESA which contains quite a few breaking changes.

I gave a talk introducing Ember Simple Auth 1.0 and the changes that come with it at [Ember.js Munich](http://www.meetup.com/Ember-js-Munich/): https://www.youtube.com/watch?v=aeH_9YGgt9E

## TODO

- [x] set up the configured session store as `session-stores:application` which can be overridden in `app/session-stores/application.js`
- [x] fix README
- [x] go to the app's root url instead of simply reloading in `ApplicationRouteMixin.sessionInvalidated`
- [x] Facebook auth restoration doesn't seem to work in dummy app
- [x] make sure ESA registers as a library and thus shows up in the initial startup log
- [x] make sure acceptance tests work nice
- [x] turn the session into a service (actually there's a service now that wraps the session which itself isn't directly exposed anymore)
- [x] split the initialization code into initializers and instance initializers
~~- [ ] clean up code and tests (e.g. session code is pretty messy, some tests use session instead of mocked session service etc.)~~ (this can actually be moved to after the 1.0 release as it won't affect the public API)
- [x] make sure the documentation is correctly generated and convert all examples and references to Ember CLI/ES6 modules syntax
[...]

PR #714, opened 2015-10-12
Making stores async

rmachielse:
Hi @marcoow,

Here is the pull request on #602 we discussed some time ago, to make stores async by default.
This will allow people to build stores that persist on asynchronous storage (like chrome app storage).

You mentioned that you would like to postpone this to release 1.1, but this change might be major, since everyone with a custom stores needs to adapt to the asynchronous workflow.


marcoow:
This is great, thanks!

I still think this can be merged without requiring a new major rebase though. Basically all that's needed is a method in the internal session that wraps all store access into a Promise if the store doesn't return a promise anyway. 1.0 has been delayed so much already it really needs to get released now. Of course making this PR work with non-async stores will require some work but I think it's worth it to not delay 1.0 any longer.

Definitely going to merge this post 1.0!

rmachielse:
Alright, I'll add backward compatibility soon. Thanks for the help!

marcoow:
thanks for the effort!

addon/session-stores/cookie.js
```diff
@@ -159,33 +172,43 @@ export default BaseStore.extend({
...
+          this._lastData = data;
+          this.trigger('sessionDataUpdated', data);
+        }
+        if (!Ember.testing) {
+          Ember.run.cancel(this._syncDataTimeout);
+          this._syncDataTimeout = Ember.run.later(this, this._syncData, 500);
+        }
+        resolve();
+      });
+    });
```
marcoow: Does this have to return a promise as well? It's actually synchronous and doesn't belong to the public API (which is only `persist`, `restore`, `clear`).

marcoow: forget that - I see how it all makes sense now ;)

rmachielse:
Should I open a new pr on master?

marcoow:
yes please - thanks!

PR #715, opened 2015-10-13
Fix dummy app authorizers filename

anilmaurya:
Authorizer filename should be oauth2 instead application


marcoow:
`app/authorizers/application.js` is fine

anilmaurya:
@marcoow It was not working for me, changing to oauth2.js fixed it. 

In Readme also, its mentioned to add app/authorizers/oauth2.js .

marcoow:
You just have to use the name name you use when calling the session service's `authorize` method or you set the `DataAdapterMixin`'s `authorizer` property to.

Also this change breaks the build.

anilmaurya:
Ok , I didn't knew that. Thanks for the info.

PR #722, opened 2015-10-16
Only invalidate session if authenticated

### LINKED ISSUE #721: DataAdapterMixin error invalidating session if not yet authenticated
This error is raised from the DataAdapterMixin if the server returns a 401 request *and* the session has not been authenticated yet.
> Uncaught Error: Assertion Failed: Session#invalidate requires the session to be authenticated!

This is a fatal error that prevents Ember from transitioning to the error route.

wireframe:
Fix #721

### DIFF
--- a/addon/mixins/data-adapter-mixin.js
+++ b/addon/mixins/data-adapter-mixin.js
@@ -94,7 +94,9 @@ export default Ember.Mixin.create({
   */
   handleResponse(status) {
     if (status === 401) {
-      this.get('session').invalidate();
+      if (this.get('session.isAuthenticated')) {
+        this.get('session').invalidate();
+      }
       return true;
     } else {
       return this._super(...arguments);
--- a/tests/unit/mixins/data-adapter-mixin-test.js
+++ b/tests/unit/mixins/data-adapter-mixin-test.js
@@ -13,10 +13,10 @@ describe('DataAdapterMixin', () => {
 
   beforeEach(() => {
     hash = {};
-    sessionService = {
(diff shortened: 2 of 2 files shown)


addon/mixins/data-adapter-mixin.js
```diff
@@ -94,8 +95,10 @@ export default Ember.Mixin.create({
   */
   handleResponse(status) {
     if (status === 401) {
-      this.get('session').invalidate();
-      return true;
+      if (this.get('session.isAuthenticated')) {
+        this.get('session').invalidate();
+      }
+      return new DS.InvalidError(['Session is not authenticated']);
```
marcoow: I don't think this should return an `InvalidError` as a 401 response doesn't indicate that the record is invalid but that the request wasn't allowed due to missing authorization or insufficient rights. When returning an `InvalidError` Ember Data would actually try to get errors for the record from the error which isn't really applicable in this case as the error doesn't apply to the record but to the request as such.

wireframe: ah, ok.  I can revert that then, and can be addressed as a separate branch if needed.

The error in my case (#721) is related to session invalidation anyways.

marcoow:
Would be good to have a test for this. Also this breaks existing tests.

wireframe:
thanks for the feedback @marcoow.

I've added unit tests and rebased against latest master to cleanup the history.  let me know if there's anything else you'd like to see done.

marcoow:
thanks!

PR #724, opened 2015-10-17
Add blueprints for authorizer and authenticator

quaertym:
I think `ember-simple-auth` should provide blueprints for authorizer and authenticator. Consuming apps are expected to define their own authenticators and authorizers so these blueprints will be handy.

### DIFF
--- a/blueprints/authenticator/files/__root__/authenticators/__name__.js
+++ b/blueprints/authenticator/files/__root__/authenticators/__name__.js
@@ -1,0 +1,8 @@
+<%= imports %>
+
+export default <%= baseClass %>.extend({
+  <%= properties %>
+  authenticate: function(data) {
+    return Ember.RSVP.resolve(data);
+  }
+});
--- a/blueprints/.jshintrc
+++ b/blueprints/.jshintrc
@@ -1,0 +1,7 @@
+{
+  "predef": [
+    "console"
+  ],
+  "strict": false,
+  "node": true
+}
--- a/blueprints/authenticator/index.js
+++ b/blueprints/authenticator/index.js
@@ -1,0 +1,40 @@
+var EOL = require('os').EOL;
+
+module.exports = {
+  description: 'Generates an ember-simple-auth authenticator',
+
+  availableOptions: [
+    { name: 'torii', type: Boolean },
+    { name: 'devise', type: Boolean },
+    { name: 'oauth2', type: Boolean },
+    { name: 'base', type: Boolean }
+  ],
+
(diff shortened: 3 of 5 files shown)


marcoow:
In the majority of cases one wouldn't extend the base authenticator or authorizer but one of the concrete ones. Also in lost if these cases one wouldn't override any methods - I think the blueprints should respect that.

quaertym:
How should we proceed then?

marcoow:
Don't have a good idea but the authenticator/authorizer files this currently generates would be changed completely almost all of the times.

quaertym:
How about this?
```
ember g authenticator <name> --torii
```
This can be used to extend correct concrete type.

marcoow:
Sounds good

marcoow:
@quaertym: would be great to get that into the next release!

quaertym:
@marcoow I have implemented this as mentioned above. Alternative syntax could be:
```
ember g authenticator <name> --type=torii
```
Which one do you prefer?

marcoow:
Looks good. Is the argument optional? It would be good if `ember g authenticator <name>` generated an authenticator that inherits from the base authenticator. Also it would be nice to have tests for this - not sure how/whether that works at all though.

quaertym:
Yep, argument is optional as you said. I am not sure either how to test these.

blueprints/authenticator/files/__root__/authenticators/__name__.js
```diff
@@ -0,0 +1,8 @@
+<%= imports %>
+
+export default <%= baseClass %>.extend({
+  <%= properties %>
+  authenticate: function(data) {
+    return Ember.RSVP.resolve(data);
+  }
+});
```
marcoow: This should have the `restore` and `invalidate` methods as well since when inheriting the base authenticator it's mandatory to implement them. Also the default implementation of `authenticate` should just be empty I'd say.

marcoow:
It would be great to include a generator for a session store as well. This would not take a name as the generated store would always be `app/session-stores/application.js` but it would take a `--<type>` as well that could either be `cookie`, `local-storage`, `ephemeral` or `adaptive`.

marcoow:
In order to test blueprints, there is the [ember-cli-blueprint-test-helpers addon](https://github.com/ember-cli/ember-cli-blueprint-test-helpers) - see example usage here: https://github.com/emberjs/data/commit/7f03ec70ff41739af51ffc8e2e14a96f8f3e0878

marcoow:
replaced by #879

PR #726, opened 2015-10-17
make clientSecret work

ServiceTo:
I wasn't able to update my simple-auth implementation to ember-simple-auth without this little modification.

My `app/authenticators/oauth2.js` changed to:
```
import ENV from 'appname/config/environment';
 
       serverTokenEndpoint: ENV.oauth2.serverTokenEndpoint,
       serverTokenRevocationEndpoint: ENV.oauth2.serverTokenRevocationEndpoint,
       clientId: ENV.APP.client_id,
       clientSecret: ENV.APP.client_secret
});
```


marcoow:
You should not use `clienr_secret` in an Ember app as i iss not a secret anymore when it is included in your app's source.

PR #729, opened 2015-10-19
Update README.md to fix login controller error

albertosouza:
from wrong Ember.Component class in example

If set controller with Ember.conponent will return : `Error while processing route: login controller._updateCacheParams is not a function TypeError: controller._updateCacheParams is not a function`


marcoow:
👍

PR #733, opened 2015-10-20
Use shorthand method syntax consistently

knownasilya:
Some of the examples used the new syntax, while others didn't. I converted all of the examples to
use the new syntax.


marcoow:
👍

Issue #782, opened 2015-11-12
Silently failing, rejecting promise...


devinrhode2:
I first asked this question on stack overflow here: http://stackoverflow.com/questions/33665152/ember-simple-auth-session-not-getting-authenticated-after-calling-resolve-in-cus but I have now figured out that there is a failure happening deep in ember-simple-auth and the error is not appearing anywhere... 

I'm going to try and fill in more details.. cc @marcoow 

The scenario where a developer, such as myself, get's lost is, when writing a custom authenticator, if they have a syntax error in their authenticate promise:
```js
import Ember from 'ember';
import Base from 'simple-auth/authenticators/base';

export default Base.extend({
  authenticate: function(email, password, authenticateCallback) {
    //returning new Promise or new Ember.RSVP.Promise resulted in simple-auth's then callback not getting called.
    //returning a $.ajax, we can be sure the .then callbacks get called; and this works, whereas new Promise doesn't, for some reason..
    console.log('custom authenticator authenticate..')
    return new Ember.RSVP.Promise(function(resolve, reject) {
      throw new SyntaxError('foo') // <-- Developers need to find out about any syntax error causing an auto-rejection..
```
They will probably not find out about it (unless they are listening to the authentication failed event..

Relevent code in ember-simple-auth: https://github.com/simplabs/ember-simple-auth/blob/master/addon/internal-session.js#L24

devinrhode2:
Handling in other threads.

PR #783, opened 2015-11-12
log exceptions in custom authenticate methods

devinrhode2:
If a developer is just going right in and writing a custom authenticator, they need to see errors, and not have them set as an errorMessage property. If we want to use this errorMessage property, which I think is foolish, developers should fully understand and be informed about how to make sure they are seeing this error message...

Resolves https://github.com/simplabs/ember-simple-auth/issues/782

### DIFF
--- a/README.md
+++ b/README.md
@@ -172,7 +172,7 @@ export default Ember.Controller.extend({
     authenticate() {
       let { identification, password } = this.getProperties('identification', 'password');
       this.get('session').authenticate('authenticator:oauth2', identification, password).catch((reason) => {
-        this.set('errorMessage', reason.error);
+        console.error('exception in your authenticators authenticate method', reason)
       });
     }
   }
--- a/README.md
+++ b/README.md
@@ -172,7 +172,7 @@ export default Ember.Controller.extend({
     authenticate() {
       let { identification, password } = this.getProperties('identification', 'password');
       this.get('session').authenticate('authenticator:oauth2', identification, password).catch((reason) => {
-        console.error('exception in your authenticators authenticate method', reason)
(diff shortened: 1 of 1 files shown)


marcoow:
This doesn't work as you'd log an exception now ales if the authenticator just rejected authentication because e.g. credentials were invalid or so. Might make more sense to log the fact that authentication was rejected (which doesn't necessarily indicate an exception) with `Ember.Logger.debug`.

devinrhode2:
I encountered Ember.Logger and see that this would indeed be a better approach than using console.error.
```js
  actions: {
    authenticate() {
      let { identification, password } = this.getProperties('identification', 'password');
      this.get('session').authenticate('authenticator:oauth2', identification, password).catch((reason) => {
        Ember.logger.debug('exception in your authenticators authenticate method', reason)
      });
    }
  }
```

marcoow:
Yes, the point is though that you would log that message now also every time the user enters wrong credentials which doesn't make much sense.

devinrhode2:
So perhaps the best thing would be to tell users to include a bit of template that displays the error message. ? `{{authenticationError}}`.
Sorry for just drunkenly throwing darts at the wall.

marcoow:
Maybe it would help adding sth. to the docs that says that the promise returned from `this.get('session').authenticate` might reject and that case should be handled?

devinrhode2:
What do you mean "sth"?

You are actually right with the controller code. The real problem is that the suggested login template code does not display the `errorMessage` which the controller is setting. I've suggested this change here: https://github.com/simplabs/ember-simple-auth/pull/788

PR #786, opened 2015-11-14
Correct registered version

quaertym:


marcoow:
1.0.1 is already released anyway. This will be fixed for 1.0.2 or 1.1.0 which the current master will eventually be released as.

PR #790, opened 2015-11-18
Better instance initialiser example for ApplicationRouteMixin API docs

constantm:
The current instance initialiser example doesn’t work. I’ve replaced it with an example that works correctly.

### DIFF
--- a/addon/mixins/application-route-mixin.js
+++ b/addon/mixins/application-route-mixin.js
@@ -3,6 +3,8 @@ import Configuration from './../configuration';
 
 const { inject, on } = Ember;
 
+
+
 /**
   The mixin for the application route; __defines methods that are called when
   the session was successfully authenticated (see
@@ -15,20 +17,22 @@ const { inject, on } = Ember;
 
   ```js
   // app/instance-initializers/session-events.js
-  Ember.Application.initializer({
-    name:       'session-events',
-    after:      'ember-simple-auth',
-    initialize: function(container, application) {
-      var applicationRoute = container.lookup('route:application');
-      var session          = container.lookup('service:session');
+  export function initialize(instance) {
+      var applicationRoute = instance.container.lookup('route:application');
+      var session          = instance.container.lookup('service:session');
       session.on('authenticationSucceeded', function() {
         applicationRoute.transitionTo('index');
       });
       session.on('invalidationSucceeded', function() {
         window.location.reload();
       });
-    }
   });
+
+  export default {
+    name:       'session-events',
+    after:      'ember-simple-auth',
+    initialize: initialize
+  };
   ```
 
   __When using the `ApplicationRouteMixin` you need to specify
--- a/addon/mixins/application-route-mixin.js
+++ b/addon/mixins/application-route-mixin.js
@@ -3,8 +3,6 @@ import Configuration from './../configuration';
 
 const { inject, on } = Ember;
 
-
-
 /**
   The mixin for the application route; __defines methods that are called when
   the session was successfully authenticated (see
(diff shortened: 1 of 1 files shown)


addon/mixins/application-route-mixin.js
```diff
@@ -15,20 +17,22 @@ const { inject, on } = Ember;
...
         window.location.reload();
       });
-    }
   });
+
+  export default {
+    name:       'session-events',
+    after:      'ember-simple-auth',
+    initialize: initialize
+  };
```
marcoow: I'd say changing `container.lookup` to `instance.container.lookup` should be sufficient. I don't think exporting the initialize function doesn't benefit the purpose of the example.

addon/mixins/application-route-mixin.js
```diff
@@ -3,6 +3,8 @@ import Configuration from './../configuration';
 
 const { inject, on } = Ember;
 
+
+
```
marcoow: unnecessary newline

addon/mixins/application-route-mixin.js
constantm: True. However, I think a more "complete" example would go far in helping people new to Ember/ES6 figure out how things work in relation to the rest of the ecosystem (I'm new, and this would have helped me understand it better). This way is the example of how initialisers are used I see most often, so I think it would help to keep things consistent.

marcoow: Sure, this isn't a complex example though ;) It only illustrates that session events can be handled in an instance initializer.

constantm: Can we make it a complex example though? I really think it would help other new Ember users to understand the context better. Pleeeeease. :)

constantm: Seriously though. I think consistency is very important, and this is the Ember CLI way as far as I understand. Why would we not include a full example if it'll actually help people?

marcoow: Yeah, guess you're right. In that case it should be

```js
export default {
  initialize,
  name:  'session-events',
  after: 'ember-simple-auth',
};
```

though. Also please fix the whitespaces above.

constantm: Done!

addon/mixins/application-route-mixin.js
```diff
@@ -15,20 +15,22 @@ const { inject, on } = Ember;
...
       session.on('invalidationSucceeded', function() {
         window.location.reload();
       });
-    }
   });
+
+  export default {
+    initialize,
+    name:    'session-events',
+    after:   'ember-simple-auth',
```
marcoow: 1 small thing - unnecessary space and comma here

```js
export default {
  initialize,
  name:  'session-events',
  after: 'ember-simple-auth'
};
```

constantm: Your own example you just gave included that comma, so I thought it was from some style guide or something. I'll remove the spaces for you. -_-

addon/mixins/application-route-mixin.js
```diff
@@ -15,20 +15,22 @@ const { inject, on } = Ember;
...
-      var session          = container.lookup('service:session');
+  export function initialize(instance) {
+      var applicationRoute = instance.container.lookup('route:application');
+      var session          = instance.container.lookup('service:session');
       session.on('authenticationSucceeded', function() {
         applicationRoute.transitionTo('index');
       });
       session.on('invalidationSucceeded', function() {
         window.location.reload();
       });
```
marcoow: Can you fix the spacing here? Should be intended by 2 spaces but is 4.

constantm: Done.

marcoow:
👍

PR #808, opened 2015-12-01
Adding ability to pass authenticator name and session data

kevintechie:
This change enables custom Torii authentication providers to use session data when the close() method is called.

### DIFF
--- a/addon/authenticators/torii.js
+++ b/addon/authenticators/torii.js
@@ -104,23 +104,11 @@ export default BaseAuthenticator.extend({
     @public
   */
   invalidate(data) {
-    this._assertToriiIsPresent();
-
-    data = data || {};
     return new RSVP.Promise((resolve, reject) => {
-      if (!isEmpty(data.provider)) {
-        let { provider } = data;
-        this.get('torii').close(data.provider, data).then((data) => {
-          data = data || {};
-          this._resolveWith(provider, data, resolve);
-        }, () => {
-          delete this._provider;
-          reject();
-        });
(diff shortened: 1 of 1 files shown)


addon/authenticators/torii.js
```diff
@@ -103,12 +103,24 @@ export default BaseAuthenticator.extend({
...
+        }, () => {
+          delete this._provider;
+          reject();
+        });
+      } else {
         delete this._provider;
-        resolve();
-      }, reject);
+        reject();
+      }
```
marcoow: Why this change?

addon/authenticators/torii.js
```diff
@@ -103,12 +103,24 @@ export default BaseAuthenticator.extend({
     @return {Ember.RSVP.Promise} A promise that when it resolves results in the session being invalidated
     @public
   */
-  invalidate() {
+  invalidate(data) {
+    this._assertToriiIsPresent();
+
+    data = data || {};
     return new RSVP.Promise((resolve, reject) => {
-      this.get('torii').close(this._provider).then(() => {
```
marcoow: I guess just adding `data` as a 2nd argument here would be sufficient?

kevintechie: For my API, yes, it is sufficient to just add `data` as a parameter to the method and then pass it as a second argument in the `close()` method. However, it's possible that some APIs may return data in the response that may be needed by the custom Torii provider. I don't know if that's a realistic expectation.

PR #817, opened 2015-12-10
Simplify handling of current user

marcoow:
This changes the sessionAccount service so that it explicitly loads the current user which has the advantage that it doesn't have to use a promise object anymore but when the current user is present it will be an actual record.

### DIFF
--- a/tests/dummy/app/routes/application.js
+++ b/tests/dummy/app/routes/application.js
@@ -1,4 +1,21 @@
 import Ember from 'ember';
 import ApplicationRouteMixin from 'ember-simple-auth/mixins/application-route-mixin';
 
-export default Ember.Route.extend(ApplicationRouteMixin);
+const { service } = Ember.inject;
+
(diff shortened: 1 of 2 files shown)


marcoow:
If this pattern proves to work well it might be good to have it officially supported by adding a setupOnAuthentication method or so to the `ApplicationRouteMixin` that provides a defined extension point for adding things like this.

Kilowhisky:
So i tried this out and had to structure my `sessionAuthenticated()` differently because of the race condition between the super call and the loadCurrentUser().  This way the user is for sure downloaded before they are redirected to an authenticated route. 

```
    sessionAuthenticated() {
        let _base = this._super.bind(this, ...arguments);
        this._loadCurrentUser().then(() => {
            _base();
        }).catch(() => this.get('session').invalidate());
    },
```

ballPointPenguin:
This is really helpful. Thank you :hamster:

PR #822, opened 2015-12-15
Remove container from internal session to fix deprecation message

jcope2013:
more changes related to Ember 2.3 deprecating access to the container property on framework generated objects, feedback wanted

### DIFF
--- a/addon/internal-session.js
+++ b/addon/internal-session.js
@@ -6,7 +6,6 @@ const { RSVP, on } = Ember;
 export default Ember.ObjectProxy.extend(Ember.Evented, {
   authenticator:       null,
   store:               null,
-  container:           null,
   isAuthenticated:     false,
   attemptedTransition: null,
   content:             { authenticated: {} },
--- a/tests/unit/internal-session-test.js
+++ b/tests/unit/internal-session-test.js
@@ -17,7 +17,7 @@ describe('InternalSession', () => {
     let container = { lookup() {} };
     store         = EphemeralStore.create();
     authenticator = Authenticator.create();
-    session       = InternalSession.create({ store, container });
+    session       = InternalSession.create({ store });
     sinon.stub(container, 'lookup').withArgs('authenticator').returns(authenticator);
   });
 
--- a/tests/unit/internal-session-test.js
+++ b/tests/unit/internal-session-test.js
@@ -17,7 +17,7 @@ describe('InternalSession', () => {
     let container = { lookup() {} };
     store         = EphemeralStore.create();
     authenticator = Authenticator.create();
-    session       = InternalSession.create({ store });
+    session       = InternalSession.create({ store, container });
     sinon.stub(container, 'lookup').withArgs('authenticator').returns(authenticator);
   });
 
--- a/tests/unit/services/session-test.js
+++ b/tests/unit/services/session-test.js
@@ -20,7 +20,7 @@ describe('SessionService', () => {
     };
     let container = { lookup() {} };
     sinon.stub(container, 'lookup').withArgs('authorizer').returns(authorizer);
-    sessionService = Session.create({ session });
+    sessionService = Session.create({ container, session });
   });
 
   it('forwards the "authenticationSucceeded" event from the session', (done) => {
--- a/tests/unit/services/session-test.js
+++ b/tests/unit/services/session-test.js
@@ -20,7 +20,7 @@ describe('SessionService', () => {
     };
     let container = { lookup() {} };
     sinon.stub(container, 'lookup').withArgs('authorizer').returns(authorizer);
-    sessionService = Session.create({ container, session });
+    sessionService = Session.create({ session });
   });
 
   it('forwards the "authenticationSucceeded" event from the session', (done) => {


marcoow:
This breaks a lot of functionality. Can you elaborate on why it is needed?

jcope2013:
yeah you get this deprecation message in your console without this change which stack trace doesn't really point to anything helpful, when applying this PR the deprecation message goes away, I am not sure it's the best approach though
``` 
DEPRECATION: Using the injected `container` is deprecated. Please use the `getOwner` helper instead to access the owner of this object. [deprecation id: ember-application.injected-container] See http://emberjs.com/deprecations/v2.x#toc_injected-container-access for more details.
        at HANDLERS.(anonymous function) (http://localhost:4200/provider/assets/vendor.js:16968:7)
        at raiseOnDeprecation (http://localhost:4200/provider/assets/vendor.js:16878:12)
        at HANDLERS.(anonymous function) (http://localhost:4200/provider/assets/vendor.js:16968:7)
        at invoke (http://localhost:4200/provider/assets/vendor.js:16984:7)
        at deprecate (http://localhost:4200/provider/assets/vendor.js:16937:32)
        at Object.deprecate (http://localhost:4200/provider/assets/vendor.js:27060:37)
        at Object.defineProperty.get (http://localhost:4200/provider/assets/vendor.js:12170:26)
        at Function.ClassMixinProps._lazyInjections (http://localhost:4200/provider/assets/vendor.js:47661:19)
        at instantiate (http://localhost:4200/provider/assets/vendor.js:12125:34) ```

**Comment by marcoow** [comment]
I don't think it's actually possible to remove the container as ember-getowner-polyfill simply delegates to that in any Ember version < 2.3 I think.

**Comment by mdehoog** [comment]
Removing the container fixes the deprecation in my environment (Ember 2.3), and I cannot see any side-effects so far. Is it possible to somehow inject the container only on < 2.3?

File: tests/unit/internal-session-test.js [hunk 1]
```diff
@@ -17,7 +17,7 @@ describe('InternalSession', () => {
     let container = { lookup() {} };
     store         = EphemeralStore.create();
     authenticator = Authenticator.create();
-    session       = InternalSession.create({ store, container });
```
marcoow: This is still needed - shouldn't trigger deprecations in your code though.

tests/unit/services/session-test.js
```diff
@@ -20,7 +20,7 @@ describe('SessionService', () => {
     };
     let container = { lookup() {} };
     sinon.stub(container, 'lookup').withArgs('authorizer').returns(authorizer);
-    sessionService = Session.create({ container, session });
```
marcoow: This is still needed - shouldn't trigger deprecations in your code though.

addon/internal-session.js
```diff
@@ -6,7 +6,6 @@ const { on } = Ember;
 export default Ember.ObjectProxy.extend(Ember.Evented, {
   authenticator:       null,
   store:               null,
-  container:           null,
```
marcoow: I think this is actually the only line that should be removed.

marcoow:
@jcope2013, @mdehoog: can someone have a look why those 2 tests fail?

jcope2013:
@marcoow I will look this weekend, some type of stubbing issue

adding a beforeEach into the describe fixes these 2, but causes a different one to fail

```
  beforeEach(() => {
      sinon.stub(store, 'restore').returns({});
  });
```

https://github.com/simplabs/ember-simple-auth/blob/master/tests/unit/internal-session-test.js#L196

mdehoog:
@jcope2013 I created a [PR](https://github.com/jcope2013/ember-simple-auth/pull/1) on your repo's  `ember-2.3-container-deprecation` branch that fixes the unit tests.

You were right, it was indeed a stubbing issue. `container.lookup` is called with an argument of `'authenticator:test'` for the failing tests. The container stub needed to be changed from:
```javascript
    sinon.stub(container, 'lookup').withArgs('authenticator').returns(authenticator);
```
to:
```javascript
    sinon.stub(container, 'lookup').withArgs('authenticator').returns(authenticator)
      .withArgs('authenticator:test').returns(authenticator);
```

jcope2013:
good to go, thanks @mdehoog

tests/unit/internal-session-test.js
```diff
@@ -18,7 +18,8 @@ describe('InternalSession', () => {
     store         = EphemeralStore.create();
     authenticator = Authenticator.create();
     session       = InternalSession.create({ store, container });
-    sinon.stub(container, 'lookup').withArgs('authenticator').returns(authenticator);
+    sinon.stub(container, 'lookup').withArgs('authenticator').returns(authenticator)
+      .withArgs('authenticator:test').returns(authenticator);
```
marcoow: This should actually never be called with `'authenticator:test'` in this test. I think the actual problem isn't that this stub was missing but that some lookup is made that isn't supposed to be made. Some previous test might not be cleaning up the session store correctly. I assume that the test passes in isolation.

dcalhoun: From my diving into this, either the specs within `tests/acceptance/authentication-test.js` or the methods within `test-support/helpers/ember-simple-auth.js` seem to be the culprit for not properly cleaning the session store. I found this because commenting out the acceptance tests caused the two failing specs to pass.

I [opened a PR](https://github.com/jcope2013/ember-simple-auth/pull/2) to resolve it. I'm not sure if my commit is the best approach to resolve it, but it does seem to fix the specs without improperly stubbing for `'authenticator:test'`.

addon/internal-session.js
```diff
@@ -6,7 +6,6 @@ const { RSVP, on } = Ember;
...
   isAuthenticated:     false,
   attemptedTransition: null,
   content:             { authenticated: {} },
```
marcoow: The problem is actually this. This `content` object will be shared across all instances of the `InternalSession` class as it is defined on the prototype…

This would have to be replaced with 

```js
_init() {
    this._super(...arguments);
    this.set('content', { authenticated: {} });
  },
```

which might cause other problems with other tests but this is actually a sever bug - not sure what I was thinking when adding this…

kbullaughey:
I started looking into a fix to the container deprecation warnings and kept coming back to the problem that the testing of `internal-session` does all sorts of hairy faking of the container. Would it make sense to wire up `internal-session` using the standard testing helpers that handle creating isolated containers for unit tests? i.e., `describeModule`. I think this would require having a resolvable name for it or customizing the resolver to handle the fact that `internal-session` is not in a subdirectory.

marcoow:
fixed via #894

PR #824, opened 2015-12-15
add clientSecret option to oauth2-password-grant with related test

jamemackson:
This adds the ability to pass a clientSecret per https://tools.ietf.org/html/rfc6749#appendix-A.2

This allows implementing an oauth2 authenticator that uses the existing clientId and the clientSecret property that has been added in this PR.  This should not cause any backwards compatibility issues, as it should still return the same value as it did previously `clientId:` for the Authorization header if no clientSecret is provided.  Please let me know if anyone see's any issues with this approach and I'm happy to update based on suggestions here.  

```
// example /app/authenticators/oauth2.js

import config from '../config/environment';

  clientId: `${config.apiClientId}`,
  clientSecret: `${config.apiClientSecret}`
});
```

### DIFF
--- a/addon/authenticators/oauth2-password-grant.js
+++ b/addon/authenticators/oauth2-password-grant.js
@@ -42,6 +42,19 @@ export default BaseAuthenticator.extend({
   clientId: null,
 
   /**
+   The client_secret to be sent to the authentication server (see
+   https://tools.ietf.org/html/rfc6749#appendix-A.1). __This should only be
+   used for statistics or logging etc. as it cannot actually be trusted since
+   it could have been manipulated on the client!__
+
+   @property clientSecret
+   @type String
+   @default null
+   @public
+   */
+  clientSecret: null,
+
+  /**
     The endpoint on the server that authentication and token refresh requests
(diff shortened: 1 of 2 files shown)


jamemackson:
Weird this fails with just ember-beta, and because the `bower.json` file isn't there... I verified that this file does indeed [exist in the source](https://github.com/emberjs/ember.js/blob/v2.3.0-beta.2/bower.json) of the latest beta release... any suggestions here?  

```
ember-data#4fe0482be3 bower_components/ember-data
For scenario ember-beta, using:
Versions do not match: Expected: components/ember#beta but saw 2.3.0-beta.2+41030996 This might be ok, depending on the scenario
  ember 2.3.0-beta.2+41030996
File /home/travis/build/simplabs/ember-simple-auth/bower_components/ember-data/bower.json does not exist
undefined
```

marcoow:
You cannot use a client secret in a web app as including the secret in the web app's source renders it not a secret anymore.

Thanks for the effort but this will be rejected like all the other PRs suggesting to add client secret support.

PR #829, opened 2015-12-24
Exposes the invalidation error

xcambar:
Makes invalidations more explicit


marcoow:
This breaks the build. Also it would be better to only log the error if it is actually present.

xcambar:
Fixed. Hope you like it better so.

marcoow:
👍

PR #833, opened 2015-12-29
Set browserify.tests only for addon tests

### LINKED ISSUE #832: Ember-CLI test runner fails with simple-auth and browserify
If I create a new app with Ember CLI 1.13.13 and add ember-browserify + ember-simple-auth, the test runner fails to load.

I think this is because you have browserify.tests = true in your config/environment.js, but that shouldn't be the case for applications consuming browserify. Either adding browserify.tests = false to my application's config/environment.js or moving yours to tests/dummy/config/enviornment.js makes everything happy again.

From my experimentation, this has to do with where browserify_stubs.js gets inserted into test-support.js when browserify.tests = true.

Also interesting to note if I downgrade ember-cli-qunit from ^1.0.4 (which currently evaluates to 1.1.0) to explicitly 1.0.4 the problem also goes away.

Here's a simple Ember-CLI project demonstrating the issue: https://github.com/mmelvin0/simple-auth-browserify-test

Even though I don't think it's the whole story, I don't think ember-simple-auth should be setting browserify.tests = true for every application that includes it, only for its own tests. I'll submit a pull request that changes that.

mmelvin0:
See also https://github.com/mmelvin0/simple-auth-browserify-test and ember-cli/ember-cli#5231 for more info.

Closes #832

### DIFF
--- a/config/environment.js
+++ b/config/environment.js
@@ -1,9 +1,5 @@
 'use strict';
 
 module.exports = function(/* environment, appConfig */) {
-  return {
-    browserify: {
-      tests: true
-    }
-  };
+  return { };
 };
--- a/tests/dummy/config/environment.js
+++ b/tests/dummy/config/environment.js
@@ -23,6 +23,10 @@ module.exports = function(environment) {
     }
   };
 
+  ENV.browserify = {
+    tests: true
+  };
+
   ENV.torii = {
     providers: {
       'facebook-oauth2': {


kmiyashiro:
This is actually a fix for ember-cli 2.2. If you use the 2.2 and ember-simple-auth, the test runner will fail without this fix. Would you consider releasing this as a patch version? I have to pin simple-auth to this sha for my tests to work.

marcoow:
1.1 will be released soon - for now please use this commit

robclancy:
"soon"

marcoow:
@robclancy: you're welcome to help and make "soon" even sooner…

robclancy:
What by making a PR for this change back into 1.0.2? This is a VERY basic change and you just ignore the issue for your current stable release because 1.1 is coming "soon".

marcoow:
Either by that or by helping us fix the issues that prevent us from releasing 1.1.0.

Remember this is Open Source - nobody is paying anyone for it but everybody is welcome to contribute.

robclancy:
So you need to get paid to merge this into stable and make a new stable release like literally every other open source project manages to do?

mmelvin0:
@robclancy: Can you use the workaround on simplabs/ember-simple-auth#832 in the meantime?

robclancy:
I had to upgrade to a beta release for this. Then other things broke and then this issue was just ignored instead. I have in the past had to set npm to use a specific commit from this package due to stable being broken or having to use master but not wanting new breaking changes that will come up. There is now a pending task for us to remove this dependency since it is unreliable.

PR #835, opened 2015-12-30
add method adjustAjaxConfig to allow customization of ajax request

kbullaughey:
Customization the ajax request is necessary if one wants to use `contentType: 'application/json'` for example.

### DIFF
--- a/addon/authenticators/devise.js
+++ b/addon/authenticators/devise.js
@@ -140,7 +140,7 @@ export default BaseAuthenticator.extend({
   */
   makeRequest(data) {
     const serverTokenEndpoint = this.get('serverTokenEndpoint');
-    return Ember.$.ajax({
+    return Ember.$.ajax(this.adjustAjaxConfig({
       url:      serverTokenEndpoint,
       type:     'POST',
       dataType: 'json',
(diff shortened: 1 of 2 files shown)


marcoow:
I think it would be simpler to introduce a second argument `options` to the `makeRequest` method that would get merged with the default options. That way you can simply override `makeRequest` in a custom authenticator:

```js
makeRequest(data) {
  return this._super(data, {
    contentType: 'application/json'
  });
}
```

marcoow:
@kbullaughey: any updates on this (see my above comment)?

kbullaughey:
That occurred to me, but `makeRequest` is called automatically by `authenticate`, and it would be strange to pass an extra parameter there, which would in turn get passed to `makeRequest`.  I think it depends whether we see this as a configuration option or a behavior. If you'd like it to be a configuration option, where would be the best place to put it? Should it be devise-specific or available to the other authenticators? If it's conceptually more like a behavior, I think achieving it by overriding a method is a natural way to accomplish it, and this seems to be practiced quite broadly in ember (e.g., model hooks, lifecycle hooks, event handlers and such). I'm happy to implement whatever solution you think would fit well.

marcoow:
Yes, the solution I point out above is based on your specific devise authenticator overriding the default one. In your custom authenticator you'd override the `makeRequest` method, passing on any XHR options you want to set:

```js
makeRequest(data) {
  return this._super(data, {
    contentType: 'application/json'
  });
}
```

That way the content type would be set for all requests that the devise authenticator makes (if the `makeRequest` method accepts a second argument with request options as I said above).

kbullaughey:
Ah. That works. I guess we can count on anyone with this particular problem doing the same, and don't need to introduce any additional public API for it.

kbullaughey:
I'll make a separate PR

kbullaughey:
PR is issue #886.

PR #840, opened 2016-01-03
Add tokenRefreshOffset property to OAuth2PasswordGrantAuthenticator

### LINKED ISSUE #823: Make offset for token refresh a config property
In _scheduleAccessTokenRefresh for the OAuth 2 Password Grant Authenticator we schedule based on 
`const offset = (Math.floor(Math.random() * 5) + 5) * 1000; `
(along with expiresAt,  and now). Id like to be able to overwrite the offset with a config property. 

If you have a slow connection and you make a call to refresh the token but the server doesn't process it in time (or receive it in time), the server will check and the token that you sent up is now expired so the refresh will fail.

Charizard:
Fixes #823

### DIFF
--- a/addon/authenticators/oauth2-password-grant.js
+++ b/addon/authenticators/oauth2-password-grant.js
@@ -245,6 +245,23 @@ export default BaseAuthenticator.extend({
     return Ember.$.ajax(options);
   },
 
+  /**
+    Specify the offset time in milliseconds to refresh the token.
+
+    By default, this returns a random number between 5 and 10. This randomization
+    is needed because in case of multiple tabs, we need to prevent the tabs from
+    sending refresh token request at the same exact moment.
+
+    @method tokenRefreshOffset
+    @return {Integer} The time in milliseonds to refresh the token.
+  */
+  tokenRefreshOffset: computed(function() {
+    const min = 5;
+    const max = 10;
+
+    return (Math.floor(Math.random() * min) + (max - min)) * 1000;
+  }).volatile(),
+
   _scheduleAccessTokenRefresh(expiresIn, expiresAt, refreshToken) {
     const refreshAccessTokens = this.get('refreshAccessTokens');
     if (refreshAccessTokens) {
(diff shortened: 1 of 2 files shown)


addon/authenticators/oauth2-password-grant.js
```diff
@@ -276,6 +298,13 @@ export default BaseAuthenticator.extend({
     });
   },
 
+  _tokenRefreshOffset() {
+    const min = this.get('minRefreshOffset');
+    const max = this.get('maxRefreshOffset');
+    Ember.assert('minRefreshOffset cannot be set greater than maxRefreshOffset', min < max);
+    return (Math.floor(Math.random() * min) + (max - min)) * 1000;
+  },
```
marcoow: Can you turn this into a computed property?

Charizard: Sorry for the delay. I have updated the PR with changing it into a computed property.

marcoow:
I think instead of adding `minRefreshOffset` and `maxRefreshOffset` which are then used in the new private CP `_tokenRefreshOffset` it would actually be simpler to add `tokenRefreshOffset` as a protected (and thus overridable) CP. That would keep the API smaller while allowing for even more flexibility. Also in the few cases where people need customized refresh offsets it should be ok having to override the full CP instead of just setting the min and max properties.

Charizard:
@marcoow That's a valid point. But, when we allow for overriding the `tokenRefreshOffset` CP itself, people generally tend to return a constant number without randomizing making it easy for race conditions in token refresh to happen when there is two tabs open at the same time.

marcoow:
I think that could be handled with documentation. Also the property needs to be marked `volatile` (which is missing in your implementation actually) so it will be recalculated every time.

Charizard:
@ryanlabouve @marcoow I have made the appropriate changes. Can you check them out?

ryanlabouve:
Would you mind using the destructured version of `Ember.computed`, since we have that available (i.e. `computed(...` instead of `Ember.computed(...`

Also, with the use of `volatile` you don't need the dependent key for `computed`. You may check out the docs on there here: [emberjs.com/...method_volatile](http://emberjs.com/api/classes/Ember.ComputedProperty.html#method_volatile) or check out another example of where it's current used in ESA here: https://github.com/simplabs/ember-simple-auth/blob/9d5a37b1ece0b82ee504e6cd6f367458b1d3120d/addon/session-stores/cookie.js#L85

Thanks for your work on this!

Charizard:
@ryanlabouve Good catch there. I have updated the PR, please check it out.

marcoow:
This seems good to merge. I'll add it to the 1.2 milestone though as I don't want to add more changes (and potential bugs) to 1.1 which is already significantly delayed.

Issue #841, opened 2016-01-03
Still 1.0.0 version but i'm on 1.0.1.


ginolon:
Why still 1.0.0 version also after I run 

```
ember install ember-simple-auth
```

![esauth](https://cloud.githubusercontent.com/assets/7945858/12081062/e39a97f0-b26e-11e5-808a-24383e3013f1.png)

marcoow:
I forgot to update the version in https://github.com/simplabs/ember-simple-auth/blob/master/vendor/ember-simple-auth/register-version.js when I released 1.0.1 - will be fixed with 1.1.

PR #842, opened 2016-01-04
Bump version number to 1.0.1

h0lyalg0rithm:
Fixes the wrong version issue in debug console. #841


marcoow:
This won't be released though as 1.0.1 is already out. 1.1.0 will show the correct version number again.

PR #845, opened 2016-01-10
Fixes #844

### LINKED ISSUE #844: Authorizer does not go to proper depth when authenticator is specified.
I was having trouble with the DataAdapterMixin not sending data in the headers.  When I dug down a bit, I discovered that in the authorizer's (I discovered it in the DeviseAuthorizer) authorize() function ended up with data that looked like this:

data = {authentication: "authenticator:devise", user: Object}

When it tried to then find the user information, it didn't go deep enough to properly populate the data.

I am unsure whether my input data is incorrect, but it seems like the function should look for my user information.

kylemellander:
In DeviseAuthorizer, data value searches to see if user data is embedded.

#844

### DIFF
--- a/addon/authenticators/devise.js
+++ b/addon/authenticators/devise.js
@@ -111,7 +111,7 @@ export default BaseAuthenticator.extend({
       data[resourceName][identificationAttributeName] = identification;
 
       return this.makeRequest(data).then(
-        (response) => run(null, resolve, response),
+        (response) => run(null, resolve, response[resourceName] || response),
         (xhr) => run(null, reject, xhr.responseJSON || xhr.responseText)
       );
     });
--- a/tests/unit/authenticators/devise-test.js
+++ b/tests/unit/authenticators/devise-test.js
@@ -127,6 +127,21 @@ describe('DeviseAuthenticator', () => {
         done();
       });
     });
+
+    it('can handle a resp with the namespace of the resource name', (done) => {
+      server.respondWith('POST', '/users/sign_in', [
+        201,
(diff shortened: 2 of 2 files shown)


marcoow:
This hardcodes the resource name while it's actually configurable on the authenticator's side. I think what actually should be done is to respect the configured `resourceName` correctly in the authenticator [here](https://github.com/simplabs/ember-simple-auth/blob/master/addon/authenticators/devise.js#L115) and only resolve with that part of the response, e.g.

```js
run(null, resolve, response[resourceName]);
```

While the fact that the `resourceName` is not used for the response is actually a bug IMHO, changing the code as suggested would be a breaking change (which in this case would probably break many apps) and thus couldn't be released before 2.0. Instead what would be possible would be to check whether the response has a property `resourceName` and if so resolve with that. Otherwise the authenticator should apply the current behavior and create a deprecation saying that the server's response is actually invalid and won't work anymore with ESA 2.0.

kylemellander:
Thanks for the direction on this.  I will look into making the fix that
direction.

marcoow:
Any updates on this?

kylemellander:
Sorry I got a new job and it distracted me on this.  I will try to spend
some time with it this weekend.

marcoow:
👍🏻

kylemellander:
@marcoow I think this should be ready to be pulled in and shouldn't break current implementations.

kylemellander:
This is now tested.

marcoow:
merged: 2759f6d72b6d5a326e4079b7c882bebbc609c56d

PR #850, opened 2016-01-14
Check that triggered has changed

kat3kasper:


marcoow:
good catch ;) I think the breaking build isn't related to your change - I'll figure it out!

marcoow:
Can you rebase on the latest master? That should fix the build.

kat3kasper:
Sorry I am not great at git. Will try to fix all the commits

PR #852, opened 2016-01-15
Check that triggered has changed

kat3kasper:


kat3kasper:
found more error replaced with https://github.com/simplabs/ember-simple-auth/pull/853

PR #853, opened 2016-01-15
Update tests to be run correctly

kat3kasper:
In the 1st test, wasn't testing against triggered.
In the 2nd and 3rd `sessionDataUpdated` was not actually being triggered so triggered would always be false


marcoow:
great, thanks!

Issue #862, opened 2016-01-21
Support accessing request options from custom authorizer


eduardoweiland:
When implementing a custom authorizer, sometimes it may be necessary to have access to the request data. An example is when accessing a [HMAC-secured API](http://www.9bitstudios.com/2013/07/hmac-rest-api-security/), in which case the request data must be hashed and the resulting hash is used to authorize the request.

Currently, the authorizer is called only with the session data and a callback function that adds the header to the request in the [session service](https://github.com/simplabs/ember-simple-auth/blob/49edaa18e4860acf0d355badcae6d797bd18dc2b/addon/services/session.js#L224). The [callback function](https://github.com/simplabs/ember-simple-auth/blob/49edaa18e4860acf0d355badcae6d797bd18dc2b/addon/mixins/data-adapter-mixin.js#L78) is passed from the DataAdapterMixin to the authorizer.

A possible workaround is to override ajaxOptions() and, instead of passing a callback function, pass an object with the request data and the callback function. But it will be nice to have a standard way to access the request data from the authorizer.

marcoow:
see #870

marcoow:
closing this as we won't need it anymore once we have #979

PR #864, opened 2016-01-24
Pass xhr settings to authorizer.

patrickfournier:
In response to issue #862 and for my own similar needs, I added a parameter to SessionService.authorize() to permit access to the XHR settings.


marcoow:
The authorizer can authorize any code block really and has no knowledge of XHR. If you're not using the `DataAdapterMixin` you'd be using it through the session sth. like this:

```js
this.get('session').authorize('authorizer:some', (headerName, headerValue) => {
  xhr.setRequestHeader(headerName, headerValue);
});
```

so no need to pass XHR options.

patrickfournier:
From what you say, I understand that DataAdapterMixin is tied in a certain way to the authorizer. So if my authorizer needs to access the xhr settings, I would need to write a new data adapter mixin with a different `beforeSend`, something like this:

  ajaxOptions() {
    const authorizer = this.get('authorizer');
    Ember.assert("You're using the DataAdapterMixin without specifying an authorizer. Please add `authorizer: 'authorizer:application'` to your adapter.", Ember.isPresent(authorizer));

    let { beforeSend } = hash;

    hash.beforeSend = (xhr, settings) => {
      this.get('session').authorize(authorizer, (publicKey, secretKey) => {
        var stringToSign = extractDataFromXHRSettings(settings);
        var signature = HMAC_SHA256(stringToSign, secretKey);
        xhr.setRequestHeader('Authorization', Credential=' + publicKey + ',Signature=' + signature);
      }, settings);
      if (beforeSend) {
        beforeSend(xhr, settings);
      }
    };
    return hash;
  },

marcoow:
Authorizers just authorize a block (which doesn't necessarily issue an XHR request) of code by converting the session data that was acquired by the authenticator when it authenticated the session to some other data, e.g. a header name and value. If you need to convert that data further you'd have to do so in the block that the session's `authorize` method calls with the authorization data - so yes, you'd have to define your own DataAdapterMixin like above.

patrickfournier:
Thanks for the confirmation!

marcoow:
This will actually be added via #870

marcoow:
oops

PR #870, opened 2016-01-29
Auth options

### LINKED ISSUE #862: Support accessing request options from custom authorizer

wuher:
This pull-request gives more flexibility for custom authorization implementations by passing the XHR query options to the authorization function. This is required for example when implementing [Hawk](https://github.com/hueniverse/hawk) authentication scheme which requires the requested URL and HTTP method as part of signing the Authorization header.

This closes #862, #903


wuher:
I also have working `authorizer` and `authenticator` implementation for the Hawk protocol in case you are interested in including those as part of `ember-simple-auth`

wuher:
Just noticed this issue: #864  So, I'm guessing the answer to my pull-request is the same?

marcoow:
Authorizers are not tied to AJAX or any other thing that the authorization information the generate is eventually injected to.

I understand the use case though - depending on the thing that the auth info will be injected into in the authorizer itself. We'd need a generic solution though as this PR would couple authorizing to jQuery AJAX.

marcoow:
Having a Hawk authenticator/authorizer in the library would definitely be nice!

wuher:
So, if we wanted to have working Hawk support in ember-simple-auth we'd need to ship three files:

* `authenticators/hawk.js`
* `authorizers/hawk.js`
* `mixins/hawk-data-adapter-mixin.js`

Or do you see another approach?

marcoow:
It might make sense to add a 2nd argument to the session's `authorize` method that can be used to pass additional options to the authorizer similar to the `authenticate` method that takes different arguments depending on the type of authenticator used to authenticate the session.

This approach would be similar to what you already have of course but more generic so that `authorize` could take all kinds and an arbitrary number of arguments depending on the specific authorizer in user. The `DataAdapterMixin` could then just always pass the XHR object or the request options or so.

wuher:
Thanks for the quick input. The function signature in `base` is currently `authorize(data, block)`. It would look nicer if the `options` parameter would be the second parameter but that would be a breaking change. Should we go with `authorize(data, block, options)` ?

One thing that makes `authorize()` a bit different from `authenticate()` (as you compared them above) is that `authenticate()` is called directly from the application code while `authorize()` is called by the session. Therefore it's a bit tricker for the application to configure what parameters should be provided.

marcoow:
Right, actually the new syntax should be `authorize(data, any, args, required, by, the, authorizer, block)` I'd say so that authorizer could accept an arbitrary number of additional arguments depending on the authorizer in use. Of course that would be a breaking change though so it would wither need to be behind a feature flag or the session service would have to check for the arity of the authorizer's `authorize` method any only pass the additional arguments on when the arity matches the number of passed arguments.

wuher:
I updated the pull-request as per your comments. The way it is now done, this shouldn't be a breaking change. 

If an authorizer needs information about the HTTP request, they can simply override `getAuthorizerParams()` in the data adapter. For example:

```
export default DS.JSONAPIAdapter.extend(DataAdapterMixin, {
    getAuthorizerParams(xhr, ajaxSetup) {
        return [ajaxSetup.url, ajaxSetup.type];
    }
});
```

The above data adapter would call the authorizer with following parameters:

```
    authorize(data, url, method, block) {
```

Not related to this pull-request but one thing that concerns me a little bit is that all of this relies on a private undocumented method `ajaxOptions`. The whole authorization stops working if ember data introduces new data adapters or modifies the existing ones so that this method goes away.

addon/services/session.js
```diff
@@ -217,11 +216,17 @@ export default Ember.Service.extend(Ember.Evented, {
...
   */
-  authorize(authorizerFactory, block) {
+  authorize(...args) {
+    Ember.assert('session.authorize() must be called with at least 2 parameters', args.length >= 2);
+
+    const authorizerFactory = args.shift();
+
     if (this.get('isAuthenticated')) {
-      const authorizer = getOwner(this).lookup(authorizerFactory);
+      const authorizer = this.container.lookup(authorizerFactory);
```
marcoow: Using `this.container` is deprecated - this should actually be `getOwner`.

marcoow:
I guess the change to the session's `authorize` method should work and also not be a breaking change. Of course that needs tests, updated docs etc.

Also the point about it all being based on Ember Data private API is valid of course. I guess it would probably be ok to add some sort of public API to Ember Data that allows for adding request options as the `DataAdapterMixin` does it? /cc @igorT

addon/services/session.js
```diff
@@ -212,16 +212,31 @@ export default Ember.Service.extend(Ember.Evented, {
...
     @method authorize
     @param {String} authorizer The authorizer to authorize the block with
     @param {Function} block The block of code to call with the authorization data generated by the authorizer
     @public
   */
-  authorize(authorizerFactory, block) {
+  authorize(...args) {
+    Ember.assert('session.authorize() must be called with at least two parameters', args.length >= 2);
+    Ember.assert('First parameter to session.authorize() must be string with the name of the used authorizer', typeof args[0] === 'string');
+    Ember.assert('Second parameter to session.authorize() must be function', typeof args[args.length - 1] === 'function');
```
wuher: I added two assertions here. What do you think?

I figured that since we now support any number of arguments but still make assumptions about the first argument and the last argument, it would probably make it easier for developers to fail early with decent error message.

wuher:
I added tests for the pull request. I also tried to enhance my comments in the code. I hope I covered all. 

Let me know if you want me to add documentation anywhere else (for example README).

addon/services/session.js
marcoow: Yeah, that probably helps. The text of the last assertion isn't correct though.

addon/mixins/data-adapter-mixin.js
```diff
@@ -56,6 +56,23 @@ export default Ember.Mixin.create({
...
+    authorizer needs more information about the issued HTTP request, you should
+    override this method in your authorizer implementation and return any
+    data from the `xhr` and `ajaxSetup` objects.
+
+    @param {Object} xhr the jQuery xhr object.
+    @param {Object} ajaxSetup the jQuery ajaxSetup object.
+   */
+  getAuthorizerParams(/*xhr, ajaxSetup*/) {
+    return [];
+  },
```
marcoow: Not so happy with adding an extra method for this actually. An alternative would be to always pass e.g. the XHR but that would mean that the `DataAdapterMixin` would only be compatible with authorizers that take the XHR as an argument.

Actually the goal should be that the `DataAdapterMixin` works with authorizers that take additional arguments out of the box.

addon/services/session.js
```diff
@@ -212,16 +212,31 @@ export default Ember.Service.extend(Ember.Evented, {
...
     @method authorize
     @param {String} authorizer The authorizer to authorize the block with
     @param {Function} block The block of code to call with the authorization data generated by the authorizer
     @public
   */
-  authorize(authorizerFactory, block) {
+  authorize(...args) {
+    Ember.assert('session.authorize() must be called with at least two parameters', args.length >= 2);
+    Ember.assert('The first parameter to session.authorize() must be string with the name of the used authorizer', typeof args[0] === 'string');
+    Ember.assert('The last parameter to session.authorize() must be function', typeof args[args.length - 1] === 'function');
```
marcoow: After thinking about this longer I don't actually think this is the way to go. Passing additional arguments as individual arguments before the callback means that the calling site has to know the signature of the authorizer in use which is not actually the case for e.g. the `DataAdapterMixin` where the authorizer is configurable and it should just work with all authorizers out of the box. Instead what would be easier would be to add an `options` argument that combines all the additional arguments for the authorizer and that would be passed for all authorizers so that the calling site doesn't have to know the signature.

As the `options` argument would have to default to `{}` though it cannot be the second argument but would have to be the last one, making it a breaking change. A solution might be to also make the new API the opt in for promise based authorization (which would actually make sense anyway) so that when the second argument is a function it would still use the old API and if the second argument is an object that would be the opt-in for promise based authorization, e.g.: 

```js
this.get('session').authorize('authorizer:custom').then(…)
```

The problem is that jQuery's `beforeSend` (as used in the `DataAdapterMixin`) doesn't support promises so we'd need to find a solution for that then.

marcoow:
I think the outcome of this and #903 is that authorization should be promise based, resulting in an API like this:

```js
this.get('session').authorize('authorizer:some', additional arguments).then((...) => {
  ...
}).catch(() => {
  ...
});
```

The internals of the `authorize` method would then look somewhat like this:

```js
authorize(authorizerFactory) {
  return new Promise((resolve, reject) => {
    if (this.get('isAuthenticated')) {
      const authorizer = getOwner(this).lookup(authorizerFactory);
      const sessionData = this.get('data.authenticated');
      authorizer.authorize(sessionData, ...arguments).then(resolve, reject);
    } else {
      reject();
    }
  });
}
```

It should actually be possible to implement this in a backwards compatible way like this:

* when the 2nd argument to the service's `authorize` method is a function use the old implementation trigger deprecations
* when the 2nd argument is not a function use the new implementation (returning a promise), passing all arguments to the authorizer
* the authorizers should use the same mechanism - when the 2nd argument is a function they'd use the old implementation, otherwise they'd return a promise
* the session needs a method `callAuthorizerAsync` that detects whether the authorizer returns a promise and if not wraps its result in one

/cc @RSSchermer

RSSchermer:
@marcoow Looks good to me :+1:

I would perhaps suggest using es6 rest arguments for the additional arguments for clarity?

```js
authorize(authorizerFactory, ...args) {
  return new Promise((resolve, reject) => {
    if (this.get('isAuthenticated')) {
      const authorizer = getOwner(this).lookup(authorizerFactory);
      const sessionData = this.get('data.authenticated');
      authorizer.authorize(sessionData, ...args).then(resolve, reject);
    } else {
      reject();
    }
  });
}
```

marcoow:
👍🏻

wuher:
Sounds good to me. I'm happy to do this if we are settled on the API.

marcoow:
Yeah, I think it should look like in @RSSchermer's latest comment. This won't go out before 1.2.0 though as it's another big change and I don't think it should be included in the same release that already has async session stores (which is 1.1).

wuher:
Fair enough. I would still like to get your input on providing the extra parameters from the adapter.

If we just update the session, it would mean that if an implementation wants to provide extra parameters to the authorizer, they would have to fully override (private and undocumented) `ajaxOptions()` function in the data adapter mixin. I was hoping we could come up with an easier API for the data adapter implementations and keep them from copy-pasting the code from the data adapter mixin. I did that with `getAuthorizerParams()` but you expressed concerns in doing that. Can you think of an alternative approach or do you see this as unnecessary?

Anyway, I'll make the changes to session regardless.

marcoow:
Sure, my idea would be that the `DataAdapterMixin` just always passes the XHR as an extra argument to the authorizer, e.g.

```js
this.get('session').authorize(authorizer, xhr).then(…)
```

That way all authorizers supporting the `DataAdapterMixin` would have to support the same API and the XHR would allow access to everything that might be important to the authorizer.

wuher:
Cool, however it's actually the second parameter to `beforeSend(xhr, ajaxSetup)` that is probably more interesting to the implementations (of course who knows...) since it contains the requested URL, path, headers etc. Amazingly, there doesn't seem to be a way to access that info from the `xhr` object (I'm looking at it in the debugger right now). XHR object seems to be more about the progress and results of the query.

Should we pass the second argument instead?

marcoow:
Yeah, would probably have to include the request options then.

wuher:
Did we ever come up with a solution to the fundamental problem that `ajaxOptions()` and jQuery's `beforeSend()` are synchronous?

Not sure how we would implement asynchronous authorization with this limitation. Unless we would override the whole `ajax()` function in ember data's REST adapter.

marcoow:
Argh, you're right. Overriding the `ajax` method should be work though as it has an `options` argument that is passed to `ajaxOptions` already.

wuher:
It would work but not only would we have to copy/paste `ajax()` code from the rest adapter but we would also be introducing networking in the `data-adapter-mixin` which is supposed to do authorization only. We would take over responsibilities of Ember Data's adapter inside the autorization mixin because I don't think we could even call `this._super()` in our `ajax()` function, we'd have to copy the whole thing. Are we sure we want to do this?

It would be nice if Ember Data supported async `ajaxOptions()` or better yet, supported authorization by introducing `authorize()` extension point (by default it wouldn't do anything). I'm not very familiar with that community so I don't know how these kinds of suggestions would be received.

marcoow:
I think sth. (roughly) like this should actually work:

```js
ajax(url, type, options) {
  const authorizer = this.get('authorizer');
  return this.get('session').authorize(authorizer, (headerName, headerValue) => {
    options.beforeSend = (xhr) => {
      xhr.setRequestHeader(headerName, headerValue);
    });
    this._super(url, type, options);
  };
}
```

wuher:
REST adapter overwrites `beforeSend()` but you're right, something like this should work. I'll work out the details and see what it looks like. Thanks.

marcoow:
👍🏻

wuher:
One test is failing in an older version of ember (i missed the fact that we are testing with other versions so it got by me). I'll look into it. In the meantime, besides the one failing test the pull-request is complete and I welcome any feedback. I added couple of comments to PR for clarity.

addon/authorizers/devise.js
```diff
@@ -52,16 +52,28 @@ export default BaseAuthorizer.extend({
...
+
+      if (typeof args[0] === 'function') {
+        Ember.deprecate(`Ember Simple Auth: Synchronous authorizers have been deprecated. You shouldn't pass callback function to authorizer.authorize anymore, it returns promise.`, false, {
+          id: 'ember-simple-auth.DeviseAuthorizer.authorize',
+          until: '2.0.0'
+        });
+
+        args[0](headerName, headerValue);
+      } else {
+        return Ember.RSVP.resolve({ headerName, headerValue });
```
wuher: Since promises can ever only resolve with a single value, I had to combine `headerName` and `headerValue` into a single object. In this new async API, all authorizers are expected to return a promise that resolves with this kind of object.

addon/mixins/data-adapter-mixin.js
```diff
@@ -56,6 +57,39 @@ export default Ember.Mixin.create({
...
+    @param {Object} options Additional XHR options.
+  */
+  ajax(url, type, options) {
+    // tests benefit from doing it like this (instead of just `options = options || {}`).
+    if (!options) {
+      options = {};
+    }
+    const authorizer = this.get('authorizer');
+    Ember.assert("You're using the DataAdapterMixin without specifying an authorizer. Please add `authorizer: 'authorizer:application'` to your adapter.", Ember.isPresent(authorizer));
+
```
wuher: The below is a workaround to be able to call `this._super()` in the promise handler. I spent quite a bit of time on this. Googling tells me that others are having issues with mixins, promises and calling `_super()` too. I tried various tricks (like passing `this` explicitly`) but this was the simplest workaround I could come up with.

If anybody know a better way to overcome this, I'd be glad to hear it.

addon/services/session.js
```diff
@@ -207,21 +207,83 @@ export default Ember.Service.extend(Ember.Evented, {
...
+          const sessionData = this.get('data.authenticated');
+          return this._callAuthorizerAsync(authorizer, sessionData, ...args).then(resolve, reject);
+        } else {
+          reject();
+        }
+      });
+    };
+
+    // if the caller passes callback function, use the old implementation
+    if (typeof args[0] === 'function') {
```
marcoow: Wouldn't this be `args[1]`?

marcoow: ah, nevermind…

addon/mixins/data-adapter-mixin.js
```diff
@@ -67,21 +101,24 @@ export default Ember.Mixin.create({
...
+      if (mixinBeforeSend) {
+        mixinBeforeSend(xhr);
+      }
       if (beforeSend) {
         beforeSend(xhr);
       }
     };
+
     return hash;
   },
```
marcoow: Actually it should be possible to delete `ajaxOptions` completely as everything is done in `ajax` now.

tests/acceptance/authentication-test.js
```diff
@@ -30,20 +30,18 @@ describe('Acceptance: Authentication', function() {
...
-      visit('/protected');
-
-      return andThen(() => {
+      return visit('/protected').then(() => {
         expect(currentPath()).to.eq('protected');
         let session = currentSession(application);
         expect(session.get('data.authenticated.userId')).to.eql(1);
         expect(session.get('data.authenticated.otherData')).to.eql('some-data');
-      });
+      }).then(done, done);
```
marcoow: Returning the `andThen` should be fine.

tests/acceptance/authentication-test.js
```diff
@@ -30,20 +30,18 @@ describe('Acceptance: Authentication', function() {
       });
     });
 
-    it('can be visited when the session is authenticated', () => {
+    it('can be visited when the session is authenticated', (done) => {
```
marcoow: don't need to use `done` when returning a promise from the test

tests/unit/authorizers/devise-test.js
```diff
@@ -32,11 +32,21 @@ describe('DeviseAuthorizer', () => {
...
-      it('calls the block with a header containing "token" and "email"', () => {
+      it('synchronously calls the block with a header containing "token" and "email"', () => {
         authorizer.authorize(data, block);
 
         expect(block).to.have.been.calledWith('Authorization', 'Token token="secret token!", email="user@email.com"');
       });
+
+      it('asynchronously calls the block with a header containing "token" and "email"', (done) => {
+        const result = authorizer.authorize(data);
+        expect(result.then).to.be.a('function');
```
marcoow: This is covered by the next line already actually.

tests/unit/authorizers/devise-test.js
```diff
@@ -52,11 +62,21 @@ describe('DeviseAuthorizer', () => {
...
-        it('calls the block with a header containing "employee_token" and "employee_email"', () => {
+        it('synchronously calls the block with a header containing "employee_token" and "employee_email"', () => {
           authorizer.authorize(data, block);
 
           expect(block).to.have.been.calledWith('Authorization', 'Token employee_token="secret token!", employee_email="user@email.com"');
         });
+
+        it('asynchronously calls the block with a header containing "employee_token" and "employee_email"', (done) => {
+          const result = authorizer.authorize(data);
+          expect(result.then).to.be.a('function');
```
marcoow: see above

marcoow:
This looks great - thanks for the hard work! I left a few comments but I'm optimistic we'll have this in 1.2!

addon/mixins/data-adapter-mixin.js
wuher: We need it because REST adapter overwrites `beforeSend` in its own `ajaxOptions`, it doesn't check if `beforeSend` has already been assigned. We re-assign it here.

tests/unit/authorizers/devise-test.js
wuher: yeah, i sometimes like to do this for readability. i'll remove it.

addon/mixins/data-adapter-mixin.js
```diff
@@ -56,6 +57,44 @@ export default Ember.Mixin.create({
...
+      options = {};
+    }
+    const authorizer = this.get('authorizer');
+    Ember.assert("You're using the DataAdapterMixin without specifying an authorizer. Please add `authorizer: 'authorizer:application'` to your adapter.", Ember.isPresent(authorizer));
+
+    // Because of the way _super() works, we need to store references to base
+    // class implementation here in order to be able to use it in the callback below.
+    const me = this;
+    const supr = this._super;  // for ember 2.x
+    const nextsuper = this.__nextSuper;  // for ember 1.x
```
marcoow: `this._super` should work in 1.x as well actually.

wuher: I agree but after debugging this for couple of hours (both, in command line and in the browser) I found no other way. In Ember 1.x, `me` and `supr` references change somehow when the promise callback is executed. If you step in the debugger into the super function implementation it will never find the base class implementation (REST Adapters `ajax()` function) and will always just return `undefined`. I tried all kinds of tricks (using `function.bind(this)` for example).

Why this happens is beyond me. Maybe Babel can affect this too? Because those references shouldn't just change. It all works if I try calling `supr()` before the async part. 

I don't like this either but so far I've found no other way unfortunately. Using `__nextSuper` is clearly private and undocumented property to outside access but now we only use it in Ember 1.x so maybe we can get away with it?

Again, if anybody knows a better way of doing this, I'm more than happy to change this.

marcoow: I'll investigate next week. Should be possible to get this working with sth. like `const super = this._super.bind(this);`

wuher: Thanks. Hope you find a solution. 

I tried the above too yesterday, but no luck. That's what got me thinking if Babel has something to do with this too.

addon/authorizers/devise.js
jesjos: I have a use case where I need to set multiple headers in an authorizer. Maybe we could let the authorizer resolve wit an array like:

```
[ { headerName: 'x-foo', headerValue: 'bar'}, {headerName: 'x-baz', headerValue: 'qux' } ]
```

alternatively:

```
[ {'x-foo': 'bar', 'x-baz': 'qux' } ]
```

Perhaps it could be useful to take the opportunity to perform both API changes at once.

marcoow: You'll have a custom authorizer that resolves with multiple headers. The devise one only resolves with one. Both is fine - the point is that the value the authorizer resolves with is the authorizer's API that can perfectly be different for different authroizers.

jesjos: It seems to me the main consumer of the authorizer is the data-adapter-mixin, and the mixin currently only supports setting one header. I could of course create my own mixin, but maybe I'm misunderstanding something?

marcoow: That's right, the `DataAdapterMixin` will consume the headers returned by the authorizers in most apps probably. However, if we choose to support arrays of headers there that should be implemented there and not force all other authorizers to return/resolve with arrays as well.

addon/mixins/data-adapter-mixin.js
wuher: Did you ever have time to investigate this? Maybe I could have another look next week if I find the time.

marcoow: No, didn't have the time yet. There still seem to be some problems with 1.1 which need to be fixed before I can take on 1.2 things.

addon/services/session.js
```diff
@@ -207,21 +207,83 @@ export default Ember.Service.extend(Ember.Evented, {
     this method does nothing.
 
     ```js
-    this.get('session').authorize('authorizer:oauth2-bearer', (headerName, headerValue) => {
-      xhr.setRequestHeader(headerName, headerValue);
+    this.get('session').authroize('authorizer:oauth2-bearer').then((auth) => {
```
Dhaulagiri: /s/authroize/authorize

marcoow:
closing due to inactivity - thanks for the effort though @wuher!

Issue #868, opened 2016-01-28
Logout redirects to /undefined when baseURL is not set in environment.js


sandreas:
If baseURL is not set in config/environment.js (i do not use baseURL-Setting, because i would like to deploy the ember app to any desired server subdirectory), logout redirects to /undefined. 

Seems that baseURL is a "must have" for ember-simple-auth and a check for undefined is missing. I would prefer to use an empty string value if undefined.

config/environment.js snippet:

```javascript
module.exports = function (environment) {
    var ENV = {
        modulePrefix: 'ember-client',
        environment: environment,
        // baseURL: '',  => commented out leads to /undefined on logout
        locationType: 'hash',
        // ...
   };
  // ...
};
```

marcoow:
Why does `baseURL: ''` not work for you?

sandreas:
Thank you for the quick reply.

It does indeed work. But i spent hours to find out what the problem is, because there was no hint, no log, no debug info or anything else. 

Ember.js itself did not fail on any point, the only problem was the simple-auth-logout. I admit that it is not a common bug, but if you run into it, it is nearly impossible to find.

I would suggest one of the following solutions:
- Assert, that baseURL is set
- Set the baseURL to '' (empty string) if it is unset (i think emberjs core acts like this )

marcoow:
> Set the baseURL to '' (empty string) if it is unset (i think emberjs core acts like this )

Would be great if you could setup a PR with this

ehntoo:
Setting `baseURL` to an empty string in the configuration causes ember-cli to emit a tag of `<base href="/">`, which causes havoc with certain SVG/CSS features. This is not really a usable solution. :(

https://github.com/ember-cli/ember-cli/issues/4905
https://github.com/ember-cli/ember-cli/issues/2633
https://github.com/emberjs/guides/issues/612

marcoow:
So this should be fine then as it only defaults ESA's `baseURL` setting to `''` even when the setting in `config/environment.js` is `null` or `undefined`.

ehntoo:
Ah, sorry, I didn't quite grok the configuration management here. This should be totally fine. :)

PR #873, opened 2016-01-29
#868 - Fixed missing default value for baseURL

sandreas:
Just changed in the configuration.js, i think the value for baseURL was missing completely.


marcoow:
Can you update the [tests](https://github.com/simplabs/ember-simple-auth/blob/master/tests/unit/configuration-test.js) as well?

marcoow:
replaced by #881

PR #877, opened 2016-02-01
throw Error on AuthenticatedRouteMixin

romulomachado:
A couple of days ago I ran into an issue and I managed to fix it.

When an unauthorized user tries to access a route that extends `AuthenticatedRouteMixin`, (s)he would be redirected to the `authenticationRoute`. If the `authenticationRoute` is wrong for whatever reason (the developer forgot to update `config/environment.js` when refactoring things (guilty of that :grimacing:)), the user would not be redirected, the template would not be loaded and there would be no error on the console either.

silent-error

This PR adds a try/catch before redirecting to `Configuration.authenticationRoute` just to make sure the `authenticationRoute` is there.

explicit-error

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,6 +1,6 @@
 The MIT License (MIT)
 
-Copyright (c) 2013-2015 simplabs GmbH/Marco Otte-Witte and contributors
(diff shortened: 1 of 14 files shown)


romulomachado:
Messed up some git config here, sending another one.

PR #878, opened 2016-02-01
Thow error on AuthenticatedRouteMixin

romulomachado:
A couple of days ago I ran into an issue and I managed to fix it.

When an unauthorized user tries to access a route that extends `AuthenticatedRouteMixin`, (s)he would be redirected to the `authenticationRoute`. If the `authenticationRoute` is wrong for whatever reason (the developer forgot to update `config/environment.js` when refactoring things (guilty of that :grimacing:)), the user would not be redirected, the template would not be loaded and there would be no error on the console either.

silent-error

This PR adds a try/catch before redirecting to `Configuration.authenticationRoute` just to make sure the `authenticationRoute` is there.

explicit-error

### DIFF
--- a/addon/mixins/authenticated-route-mixin.js
+++ b/addon/mixins/authenticated-route-mixin.js
@@ -57,7 +57,11 @@ export default Ember.Mixin.create({
       transition.abort();
       this.get('session').set('attemptedTransition', transition);
       Ember.assert('The route configured as Configuration.authenticationRoute cannot implement the AuthenticatedRouteMixin mixin as that leads to an infinite transitioning loop!', this.get('routeName') !== Configuration.authenticationRoute);
-      this.transitionTo(Configuration.authenticationRoute);
+      try {
+        this.transitionTo(Configuration.authenticationRoute);
+      } catch(e) {
+        Ember.Logger.error(e.message);
+      }
     } else {
       return this._super(...arguments);
     }


marcoow:
Hm, this seems to break the build. Also could we check in the router whether the configured route actually exists instead?

romulomachado:
@marcoow That was my first thought and I actually went that way: injecting the `-routing` service and using the `hasRoute` method, but try/catch was, IMO, way simpler. I'll run the tests as soon as I can to fix them.

marcoow:
Hm, `hasRoute` is actually private API. If there is no way doing that with public API I don't think it's an option.

addon/mixins/authenticated-route-mixin.js
```diff
@@ -57,7 +57,11 @@ export default Ember.Mixin.create({
       transition.abort();
       this.get('session').set('attemptedTransition', transition);
       Ember.assert('The route configured as Configuration.authenticationRoute cannot implement the AuthenticatedRouteMixin mixin as that leads to an infinite transitioning loop!', this.get('routeName') !== Configuration.authenticationRoute);
-      this.transitionTo(Configuration.authenticationRoute);
+      try {
+        this.transitionTo(Configuration.authenticationRoute);
+      } catch(e) {
+        Ember.Logger.error(e.message);
```
marcoow: This should re-throw the error.

Issue #884, opened 2016-02-03
Authorization information not added for ajax call outside of Ember Data


adam-knights:
In a particular component I have

```
import Ember from 'ember';
const { service } = Ember.inject;

export default Ember.Component.extend({
  session: service('session'),

  ...

  this.get('session').authorize('authorizer:oauth2', () => {
              // Use ajax as there is no data relationship here
              Ember.$.ajax({
                url: notifyAllUrl,
                method: 'POST',
                contentType: 'application/json; charset=utf-8',
                dataType: 'json',
                data: JSON.stringify({
                  id: orderId
                })
              }).done(function() {
                // do stuff
              });
            });
});
```

Authorization information is added to the headers fine for all our Ember data bits. Having read the frontpage readme it said that this is not done for every request anymore, and suggested adding the `this.get('session').authorize('authorizer:oauth2', () => {`, but this hasn't worked, and I think it should?

Ember data has it by having done

```
export default DS.RESTAdapter.extend(DataAdapterMixin, {
  namespace: 'api',
  authorizer: 'authorizer:oauth2'
});
```

But this particular POST still doesn't have `Authorization: Bearer 893u2ijhfowi...` and I had hoped that would do it.

I imagine I am doing something wrong rather than a bug, but would be happy to propose an addition to the readme if it will help others.

marcoow:
See the [API docs for the session service's `authorize` method](http://ember-simple-auth.com/api/classes/SessionService.html#method_authorize). The block is called with the authorization data created by the authorizer __which can then be inserted into request headers__.

PR #885, opened 2016-02-04
Match example given in readme for session service's authorize method to the api docs

adam-knights:
Following on from https://github.com/simplabs/ember-simple-auth/issues/884.

It was unclear in the readme that headerName and headerValue are available, whereas the api docs shows the method in full.

### DIFF
--- a/README.md
+++ b/README.md
@@ -395,11 +395,11 @@ import OAuth2Bearer from 'ember-simple-auth/authorizers/oauth2-bearer';
 export default OAuth2Bearer.extend();
 ```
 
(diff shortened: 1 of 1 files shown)


marcoow:
The actual arguments that the block is called with actually depend on the used authorizer. I'd rather have a generic example in the README, e.g.

```js
this.get('session').authorize('authorizer:some', (/*authorization data*/) => {
  // use authorization data
});
```

adam-knights:
Agreed, I think that and a further link to the api docs cover both scenarios?

marcoow:
👍🏻

PR #890, opened 2016-02-08
Fix session data handling

### LINKED ISSUE #805: Custom session data only persisted when set method is called on injected session
Related to #708.

I have a component into which I'm injecting the `session` service. I use the session to save some extra data that is persistent across reloads (an action saves the id of an entity in the session data):

``` javascript
export default Ember.Component.extend({
  session: Ember.inject.service('session'),
  actions: {
    manageBusiness: function (businessId) {
      this.get('session').set('data.businessId', businessId); //works fine
    }
  }
});
```

This works fine. However, if I change the `get.set` line to:

``` javascript
this.set('session.data.businessId', businessId);
```

...the data is no longer persistent across reloads. This is because the session service's `set` method (https://github.com/simplabs/ember-simple-auth/blob/2464beca521e0ee63c2833e4083cc3e405fbbb26/addon/services/session.js#L117) changes the `data.businessId` key to `session.businessId`, which causes the `businessId` property to be set on the `internal-session`. The `internal-session`'s `setUnknownProperty` method is then called, which updates the store (https://github.com/simplabs/ember-simple-auth/blob/2464beca521e0ee63c2833e4083cc3e405fbbb26/addon/internal-session.js#L102).

However, if you call `this.set('session.data.businessId', businessId);`, the session's `set` method is not hit and therefore the key is not replaced, and in turn the `internal-session`'s `setUnknownProperty` method is not called and the store isn't updated.

Note that it also works if I do the `data` -> `session` key replacement myself by calling `this.set('session.session.businessId', businessId);` (the `internal-session`'s `setUnknownProperty` is called).

marcoow:
This fixes a few issues with session data not being observed correctly as well as the store not being updated correctly in certain scenarios.

This is WIP and it's still unclear how this will be implemented in a backwards compatible way.

This closes #805

### DIFF
--- a/addon/internal-session.js
+++ b/addon/internal-session.js
@@ -12,11 +12,7 @@ export default Ember.ObjectProxy.extend(Ember.Evented, {
 
   init() {
     this._super(...arguments);
-<<<<<<< d70020c710e1f35560089ec482dba25f3027ffc9
-    this.set('content', { authenticated: {} });
-=======
     this._updateContent({ authenticated: {} });
->>>>>>> use an ember object for session data
   },
 
   authenticate() {
--- a/addon/internal-session.js
+++ b/addon/internal-session.js
@@ -11,7 +11,11 @@ export default Ember.ObjectProxy.extend(Ember.Evented, {
 
   init() {
     this._super(...arguments);
+<<<<<<< d70020c710e1f35560089ec482dba25f3027ffc9
     this.set('content', { authenticated: {} });
+=======
+    this._updateContent({ authenticated: {} });
+>>>>>>> use an ember object for session data
   },
 
   authenticate() {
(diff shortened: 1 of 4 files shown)


psteininger:
@marcoow  Did you guys make any progress internally on this?
It seems that #805 might be causing an issue we see with https://github.com/auth0/auth0-ember-simple-auth. When using the Lock widget, if the first attempt to log in fails, then the second one will result in `authenticationSucceeded` not being called, thus user remains stuck in a weird state on login page.

I did a lot of debugging and stepping through both the `Auth0Lock` and `auth0-ember-simple-auth` code, and everything on their end seems to work as expected.

marcoow:
No, no progress with that yet unfortunately. We hope to be able to spend some time on this this month though.

The problem you're describing doesn't sound like it's related to this though. `authenticationSucceeded` is only called when the session goes from not authenticated to authenticated so maybe the failing authentication attempt you're describing somehow semi-succeeds and makes the session think authentication was actually successful?

marcoow:
An alternative approach to this would actually be to introduce a new API `session.update((data) => { … })` that would allow to update the session data and automatically trigger a store update once the provided callback has completed so that it would ensure that any updated data would get persisted in the session store. That might also make updating the session more explicit.

Of course that would also be backwards compatible as it would be a new API - it would just not fix the existing bugs (which seems to be really hard anyway unless we introduce a huge system of observer as it seems).

psteininger:
@marcoow , this may be a tangent question, but I am trying to figure out what exactly is the mechanism of passing data from authenticator to the session for storing after authentication. I poured over the code this morning and it's not quite clear. Is there a design doc I can look at?

I am digging deeper to understand where the root of the problem lies. I got deep enough to see that `authenticationSucceeded` is not being forwarded to the `authenticator`. I am just trying to figure out why, and which part of the code is at fault here, so I can create a fix and PR it to the right project :)

psteininger:
@marcoow I think I have found what I was looking for: https://github.com/auth0/auth0-ember-simple-auth/issues/15. Looks very likely like an Auth0 problem.

marcoow:
This is not actually what we'll be doing - closing.

Issue #857, opened 2016-01-19
[Ember 2.3]: DEPRECATION: Using the injected `container` is deprecated. 


tschoartschi:
Hi!

I upgraded to Ember 2.3 and I get the following deprecation: 

```
DEPRECATION: Using the injected `container` is deprecated. Please use the `getOwner` helper instead to access the owner of this object. [deprecation id: ember-application.injected-container] See http://emberjs.com/deprecations/v2.x#toc_injected-container-access for more details.
        at HANDLERS.(anonymous function) (http://localhost:4200/app/assets/vendor.js:15523:7)
        at raiseOnDeprecation (http://localhost:4200/app/assets/vendor.js:15433:12)
        at HANDLERS.(anonymous function) (http://localhost:4200/app/assets/vendor.js:15523:7)
        at invoke (http://localhost:4200/app/assets/vendor.js:15539:7)
        at deprecate (http://localhost:4200/app/assets/vendor.js:15492:32)
        at Object.deprecate (http://localhost:4200/app/assets/vendor.js:25501:37)
        at Object.defineProperty.get (http://localhost:4200/app/assets/vendor.js:10953:26)
        at _bindToAuthenticatorEvents (http://localhost:4200/app/assets/vendor.js:173516:31)
```
When I step through the stacktrace it points to the following places:

```
(anonymous function)         application-route-mixin.js:89
(anonymous function)         application-route-mixin.js:83
setupSessionRestoration      setup-session-restoration.js:7
initialize                   ember-simple-auth.js:5
``` 

```
restore                    internal-session.js:53
beforeModel                setup-session-restoration.js:17
setupSessionRestoration    setup-session-restoration.js:7
``` 

```
_bindToAuthenticatorEvents  internal-session.js:150
_setup                      internal-session.js:89
(anonymous function)        internal-session.js:64
``` 

```
authorize        session.js:209
exports.default._ember.default.Mixin.create.ajaxOptions.hash.beforeSend        data-adapter-mixin:76
``` 

Hope this helps. Thanks

marcoow:
PR would be great - should be easily doable with https://github.com/rwjblue/ember-getowner-polyfill

tschoartschi:
@marcoow okay, I'll give it a try :smiley: but I'm not sure if I can find all the appearances since I'm not too familiar with the ```ember-simple-auth``` internals

jcope2013:
[...]

PR #891, opened 2016-02-11
Fix container deprecations, refactor internal-session tests

kbullaughey:
I have refactored the tests of `internal-session` to use `describeModule`, which provides an isolated container so stubbing one isn't necessary. In order to use `describeModule`, I needed to have a resolvable name, and so I put an `internal-session.js` into the dummy app, which allows `describeModule` to use the resolvable name `internal-session:main`. This is only necessary for testing, because in regular use the initializer sets up `InternalSession`. I also needed to change the authenticator's name to something that Ember is happy with, and so chose `authenticator:test` (i.e., it needed a colon). 

These changes appear to solve the container deprecation warnings, and do so in a way that all tests pass.

I believe this solves issues #857 and #822.

### DIFF
--- a/addon/internal-session.js
+++ b/addon/internal-session.js
@@ -6,7 +6,6 @@ const { RSVP, on } = Ember;
 export default Ember.ObjectProxy.extend(Ember.Evented, {
   authenticator:       null,
   store:               null,
-  container:           null,
   isAuthenticated:     false,
   attemptedTransition: null,
   content:             { authenticated: {} },
--- a/addon/internal-session.js
+++ b/addon/internal-session.js
@@ -8,7 +8,7 @@ export default Ember.ObjectProxy.extend(Ember.Evented, {
   store:               null,
   isAuthenticated:     false,
   attemptedTransition: null,
-  content:             { authenticated: {} },
+  content:             null,
 
   authenticate() {
     let args          = Array.prototype.slice.call(arguments);
@@ -182,5 +182,9 @@ export default Ember.ObjectProxy.extend(Ember.Evented, {
         this._clear(true);
       }
     });
+  }),
+
+  _setupContent: on('init', function() {
+    this.set('content', { authenticated: {} });
   })
 });
--- a/tests/dummy/app/internal-session.js
+++ b/tests/dummy/app/internal-session.js
@@ -1,0 +1,3 @@
+import InternalSession from 'ember-simple-auth/internal-session';
+
+export default InternalSession.extend();
--- a/tests/unit/internal-session-test.js
+++ b/tests/unit/internal-session-test.js
@@ -11,8 +11,10 @@ describeModule('internal-session:main', 'InternalSession', {}, () => {
   let session;
   let store;
   let authenticator;
+  let emberTestContext;
 
   beforeEach(function() {
+    emberTestContext = this;
     store         = EphemeralStore.create();
     session       = this.subject({ store });
     authenticator = Authenticator.create();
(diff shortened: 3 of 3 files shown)


marcoow:
@kbullaughey: thanks for the work! However, this actually makes the Internal Session public API which I don't think is great as people will actually start to rely on it.

Also the real problem is actually this line: https://github.com/simplabs/ember-simple-auth/pull/822#discussion_r52180625

If you could put sth. together that fixes that and the tests without making the internal session available to the application that would be great of course. A fix for that really needs to go out soon and I don't really have time for that currently.

kbullaughey:
I'm confused how these changes make the Internal Session API public. I've only added an `internal-session.js` file to the dummy application. I believe from the perspective of someone including the addon, nothing has changed. 

Or are you suggesting that taking out the following line somehow makes the API public?:

    container: null

Could you help me understand how the API is now public? I'd be happy to try and adjust my fix.

I realize I haven't fixed the shared `content` problem. I figured that should be a separate PR.

marcoow:
If the `internal-session.js` fils is in the `addon` folder it's only accessible from the addon itself while if it's in the `app` folder it's merged into the application's tree which doesn't officially make it public API of course but suddenly it's part of the app and people will start to use it.

Also the actuel problem that's causing tests to fail when deleting `container: null` is what I describe here: https://github.com/simplabs/ember-simple-auth/pull/822#discussion_r52180625 so fixing the shared `content` bug should be part of the same PR that removes the container property as it's actually required to get the tests green again after removing the container property.

kbullaughey:
But I didn't put `internal-session.js` in the main `app/` folder, only in `tests/dummy/app`. This was just an artifact of how `describeModule` works. In order to be able to use `describeModule` and have it set up an isolated container, it needs to be able to use the resolver to find `internal-session`. Having a real container allows `getOwner(this).lookup(authenticator)` to work out of the box and thus doesn't require stubbing. 

Are you worried about people taking a cue from the dummy application and assuming that `InternalSession` can be imported and used? Perhaps there's another way to get `describeModule` to work, or set up our own isolated container for testing using the `ember-test-helpers` functionality. I tried to do this, but it was turning into a nightmare, and so I resorted to putting `internal-session.js` in the dummy app.

How is the shared `content` related to our use of the `container`? I still don't see how these have anything to do with each other.

marcoow:
Ah sorry, I see. I thought you moved it into the `app` folder. Using `describeModule` probably makes sense but I wouldn't like to include this into this fix as it is not actually required.

The problem with the shared `content` instance is that when removing the `container: null` from `InternalSession` it will call the [`_updateStore` method](https://github.com/simplabs/ember-simple-auth/blob/master/addon/internal-session.js#L145) which uses `this.content` internally which will still have some data in it that was set in a previous test as it is a globally shared instance. Changing the `content` property so that it is a fresh instance for each instance of `InternalSession` will fix these test failures (a few others will come up though in tests that depend on the globally shared content instance).

kbullaughey:
I was getting test failures that required the `describeModule` change. The unit tests for Internal Session involve the following `beforeEach`:

    let container = { lookup() {} };
    store         = EphemeralStore.create();
    authenticator = Authenticator.create();
    session       = InternalSession.create({ store, container });
    sinon.stub(container, 'lookup').withArgs('authenticator').returns(authenticator);

Without `container: null` subbing the container in this way breaks because the following will return undefined:

    getOwner(this).lookup(authenticator)

So I think my other changes would need to be part of this PR.

marcoow:
#822 removes the container property without switching to `describeModule`. As long as the test sets the container on the internal session instance later it shouldn't be necessary to switch to `describeModule`. 

The problem in #822 is that there's data from previous tests left in the session's `content` as that's a shared instance on the prototype and thus tests have side effects on other tests. That's why it should be fixed so that it is a fresh instance for each instance of `InternalSession`.

kbullaughey:
I've added a commit that makes the content a separate object for each instance. I tried applying this change without the `describeModule` changes and lo and behold it all worked. 

I'll submit those changes as a separate PR, as it seems you prefer that route. Feel free to close this issue of you accept that PR.

marcoow:
Closed in favor of #894

PR #894, opened 2016-02-11
Remove container: null and make authentication content separate for each internal-session

kbullaughey:
This solves #822 (and #857) but without using `describeModule` as was proposed in #891.

### DIFF
--- a/addon/internal-session.js
+++ b/addon/internal-session.js
@@ -6,10 +6,13 @@ const { RSVP, on } = Ember;
 export default Ember.ObjectProxy.extend(Ember.Evented, {
   authenticator:       null,
   store:               null,
-  container:           null,
   isAuthenticated:     false,
   attemptedTransition: null,
-  content:             { authenticated: {} },
+
+  init() {
+    this._super(...arguments);
+    this.set('content', { authenticated: {} });
+  },
 
   authenticate() {
     let args          = Array.prototype.slice.call(arguments);
--- a/tests/unit/internal-session-test.js
+++ b/tests/unit/internal-session-test.js
@@ -12,9 +12,10 @@ describe('InternalSession', () => {
   let session;
   let store;
   let authenticator;
+  let container;
 
   beforeEach(() => {
-    let container = { lookup() {} };
+    container     = { lookup() {} };
     store         = EphemeralStore.create();
     authenticator = Authenticator.create();
     session       = InternalSession.create({ store, container });
@@ -725,5 +726,11 @@ describe('InternalSession', () => {
         });
       });
     });
+  });
+
+  it('does not share the content object between multiple instances', () => {
+    let session2 = InternalSession.create({ store, container });
+
+    expect(session2.get('content')).to.not.equal(session.get('content'));
   });
 });


addon/internal-session.js
```diff
@@ -6,10 +6,9 @@ const { RSVP, on } = Ember;
 export default Ember.ObjectProxy.extend(Ember.Evented, {
   authenticator:       null,
   store:               null,
-  container:           null,
   isAuthenticated:     false,
   attemptedTransition: null,
-  content:             { authenticated: {} },
+  content:             null,
```
marcoow: This can be removed completely

addon/internal-session.js
```diff
@@ -183,5 +181,9 @@ export default Ember.ObjectProxy.extend(Ember.Evented, {
         this._clear(true);
       }
     });
+  }),
+
+  _setupContent: on('init', function() {
+    this.set('content', { authenticated: {} });
   })
```
marcoow: This should better be

```js
init() {
  this._super(...arguments);
  this.set('content', { authenticated: {} });
}
```

and should also be at the top of the class then

tests/unit/internal-session-test.js
```diff
@@ -726,4 +727,20 @@ describe('InternalSession', () => {
...
+    beforeEach(() => {
+      session2 = InternalSession.create({ store, container });
+    });
+
+    it('uses separate content objects', (done) => {
+      Ember.run(function() {
+        session.set('content.secret', 'I see dead people.');
+      });
+      expect(session2.get('content.secret')).to.be.undefined;
+      done();
```
marcoow: This `done` is not necessary

kbullaughey:
Independent of this issue, how do you feel about stubbing containers vs using the built-in `isolatedContainer` functionality of `ember-test-helpers`. Does it make sense to switch to `describeModule` or some other method that uses real containers? IMO this makes the tests more realistic and less likely to fail in the future as the container API evolves.

marcoow:
Using isolated containers instead of stubbed ones definitely sounds like a good idea.

kbullaughey:
Was putting `internal-session.js` in `tests/dummy/app` too hacky? I couldn't figure out any other (reasonably easy) way to get `describeModule` to work.

marcoow:
Not sure actually - would have to investigate.

tests/unit/internal-session-test.js
```diff
@@ -12,9 +12,10 @@ describe('InternalSession', () => {
   let session;
   let store;
   let authenticator;
+  let container;
 
   beforeEach(() => {
-    let container = { lookup() {} };
+    container = { lookup() {} };
```
marcoow: would be nice to keep the aligned assignments

tests/unit/internal-session-test.js
```diff
@@ -726,4 +727,19 @@ describe('InternalSession', () => {
...
+      session2 = InternalSession.create({ store, container });
+    });
+
+    it('uses separate content objects', () => {
+      Ember.run(function() {
+        session.set('content.secret', 'I see dead people.');
+      });
+      expect(session2.get('content.secret')).to.be.undefined;
+    });
+  });
```
marcoow: can be simplified to

```js
describe('multiple internal sessions', () => {
  it('uses separate content objects', () => {
    let session2 = InternalSession.create({ store, container });

    Ember.run(function() {
      session.set('content.secret', 'I see dead people.');
    });

    expect(session2.get('content.secret')).to.be.undefined;
  });
});
```

kbullaughey: Okay. Fixed this, but it actually needs to be `equal` not `eql` because the two separate empty objects are trivially deep-equal, so I use `equal`. I thought explicitly setting something and testing for the result was more illustrative of the desired behavior rather than proof of implementation. But your solution is fine too.

marcoow:
Can you rebase on the latest master and squash the commits?

PR #896, opened 2016-02-11
use separate context objects, which helps solving container deprecation messages

kbullaughey:
See #822 for more discussion. This PR is a rebased, squashed version of #894.


marcoow:
see #894

PR #897, opened 2016-02-12
Let all responses be handled by Ember Data handleResponse

### LINKED ISSUE #858: Error handling in case of 401 / Unauthorized
It seems that the data adapter's error handling in case of a 401 response is quite strange.
It simply [returns true](https://github.com/simplabs/ember-simple-auth/blob/1.0.1/addon/mixins/data-adapter-mixin.js#L100), which leads to the problem that all handlers for failing promises (on [routes](https://github.com/pogopaule/esa-401-handling-demo/blob/master/app/routes/application.js#L13) or on [the promise itself](https://github.com/pogopaule/esa-401-handling-demo/blob/master/app/routes/application.js#L6-L8)) receive `true` as a parameter.

I'd rather expect some form of an error object that I can ask for its HTTP error status.
Or do I miss something?

You can check out [my demo project](https://github.com/pogopaule/esa-401-handling-demo), demonstrating that issue

Note: I think this issue has already been addressed in #768, but IMHO this is still a problem.

jayeff:
When the backend responds with a 401 still invalidate the session but
then let Ember Data handle the response. In the case of a 401 it would
raise an `AdapterError`

fixes #858

### DIFF
--- a/tests/unit/mixins/data-adapter-mixin-test.js
+++ b/tests/unit/mixins/data-adapter-mixin-test.js
@@ -116,10 +116,6 @@ describe('DataAdapterMixin', () => {
 
           expect(sessionService.invalidate).to.have.been.calledOnce;
         });
-
-        it('returns true', () => {
-          expect(adapter.handleResponse(401)).to.be.true;
-        });
       });
 
       describe('when the session is not authenticated', () => {
@@ -132,10 +128,6 @@ describe('DataAdapterMixin', () => {
 
           expect(sessionService.invalidate).to.not.have.been.called;
         });
-
-        it('returns true', () => {
-          expect(adapter.handleResponse(401)).to.be.true;
-        });
       });
     });
 
@@ -145,10 +137,10 @@ describe('DataAdapterMixin', () => {
 
         expect(sessionService.invalidate).to.not.have.been.called;
       });
+    });
 
-      it("returns _super's return value", () => {
-        expect(adapter.handleResponse(200)).to.eq('_super return value');
-      });
+    it("returns _super's return value", () => {
+      expect(adapter.handleResponse(401)).to.eq('_super return value');
     });
   });
 });
--- a/addon/mixins/data-adapter-mixin.js
+++ b/addon/mixins/data-adapter-mixin.js
@@ -95,13 +95,9 @@ export default Ember.Mixin.create({
     @protected
   */
   handleResponse(status) {
-    if (status === 401) {
-      if (this.get('session.isAuthenticated')) {
-        this.get('session').invalidate();
-      }
-      return true;
-    } else {
-      return this._super(...arguments);
+    if (status === 401 && this.get('session.isAuthenticated')) {
+      this.get('session').invalidate();
     }
+    return this._super(...arguments);
   }
 });


dbachet:
Indeed I was able to get `DS.AdapterError` object in the `ApplicationController`.

:+1:

dbachet:
@jayeff not sure if it can help but maybe you would like to see that comment https://github.com/simplabs/ember-simple-auth-example/pull/12/files#r52811342

This comment is related to the issue you opened.

tests/unit/mixins/data-adapter-mixin-test.js
```diff
@@ -116,10 +116,6 @@ describe('DataAdapterMixin', () => {
 
           expect(sessionService.invalidate).to.have.been.calledOnce;
         });
-
-        it('returns true', () => {
-          expect(adapter.handleResponse(401)).to.be.true;
-        });
```
marcoow: Shouldn't this be replaced with below's `it("returns _super's return value", () => {` as well?

jayeff: Unfortunately the diff is slightly difficult to parse. I pulled this test one `describe` up and it's now in the `when the response status is 401` describe block. Maybe it would be appropriate to pull it up another level (as the same test is in response status is 401 and status is not 401).

I suggest to view the [whole file](https://github.com/jayeff/ember-simple-auth/blob/update_401_error_handling/tests/unit/mixins/data-adapter-mixin-test.js#L103-L148) for better readability of the tests for `handleResponse`.

marcoow: Ah ok, thanks. Pulling it up one level more sounds good.

jayeff: Done. I decided to use status 401 on the test for the return value as it represents the only special case in `handleResponse`. Any status code would return the same result.

marcoow:
I'm not sure this is actually a good idea. It will be a breaking change as previously 401 response would not lead to an adapter error being raised. Also this would raise an adapter error although the 401 was already handled by invalidating the session which would usually lead to the page being reloaded so that any other code that handles the adapter error would never run.

jayeff:
> It will be a breaking change as previously 401 response would not lead to an adapter error being raised.

Yes, it is a breaking change but this breaking change would only change the "type" of the raised error. Currently the code raises an unspecific `true` error (check out the [repo from pogopaule](https://github.com/pogopaule/esa-401-handling-demo) to see it yourself; after starting the application open the inspector to see the logged error `true`). Replacing this with a more specific adapter error allows to improve error handling code.

> Also this would raise an adapter error although the 401 was already handled by invalidating the session which would usually lead to the page being reloaded so that any other code that handles the adapter error would never run.

Unfortunately (at least our) error handling code runs before the page reload is happening. We use `Ember.onerror` to react to a bunch of different errors (eg a 404 from the backend) and display error messages to our users. With an 401 error the `true` error will display an red error message for a short timespan before the page reload is triggered and I'm redirected to the login page. This is bad UX. I would suspect that most ember-simple-auth users which use some form of error handling would see similar behavior.

I recorded a short animation to see how the application currently behaves (error messages flashes before the page is reloaded automatically):
error-401

With the proposed change to raise an AdapterError I could inspect the error in our error handling code and ignore all 401 errors (as it's already handled by ember-simple-auth). Currently we detect that a 401 happened by code similar to `if (error === true) { // 401 Unauthorized error - do nothing }` which doesn't feel good. With the AdapterError this could be improved to `if (error.errors[0].status === "401") { [...] }`

dbachet:
@jayeff How do you avoid page refresh? 

Indeed I can't avoid it in my `ApplicationController`

jayeff:
> How do you avoid page refresh?

@dbachet We do not—in our application the page is still reloaded after session invalidation.

dbachet:
@jayeff :+1: ok thx!

marcoow:
The problem with this is mainly that the proposed change now always calls `this._super(...arguments)` - also in the case that the error was already handled and the session was invalidated. This is the breaking change I'm concerned about. Changing the returned error type from `true` to `new DS.AdapterError(status)` should be fine - I'd consider that a bug fix really. I'd like to change the method to

```js
handleResponse(status) {
  if (status === 401 && this.get('session.isAuthenticated')) {
    return new DS.AdapterError(status);
  } else {
  }
}
```

so that it only changes the returned error.

marcoow:
Any updates on this wrt my last comment?

jayeff:
Sorry for the late reply @marcoow. I was on a small skiing holiday and last week was really busy.

I'm fine with updating this PR to return a `DS.AdapterError` directly. I will push an update to this branch (+ rebase) tomorrow.

I'm a little curious why this is better than calling `this._super(...arguments)`. Care to explain the reasoning for this?
I did check the code for `handleResponse` in ember-data back till v1.13.0 and it should always return a `DS.AdapterError` when called with status 401. I suspect the big benefit is returning the `DS.AdapterError` from ember-simple-auth `handleResponse` does guarantee that the behavior will not change because of a update of ember-data. But on the other hand if the behavior of ember-data changes  (eg. a new error) wouldn't I want the same behavior from ember-simple-auth?

marcoow:
The point is that the current implementation does not call `this._super(...arguments)` when it already handled the error. Changing that so that it always calls `this._super(...arguments)` might break things for people. I'd like to change the return value in case that the mixin handled the error by invalidating the session only with this PR and leave everything else unchanged.

jayeff:
@marcoow I just updated the PR. I had to import ember-data into the data-adapter-mixin. I'm not sure if this is a good idea?

FYI: Default ember-data AdapterError on 401 (without ember-simple-auth) is `error.errors === [Object {status: "401", title: "The backend responded with an error", detail: "Unauthorized"}]`; with ember-simple-auth the AdapterError changes to `error.errors === 401`.

jayeff:
@marcoow Did you get a chance to look at the latest change? I'm not sure if raising a `DS.AdapterError` is a good idea. Maybe return a custom error instead (UnauthorizedError)?

marcoow:
Didn't really have time to look into this yet - sorry. Why do you think throwing a `DS.AdapterError` isn't a good idea?

jayeff:
Two reasons:
* Import of `ember-data`. It seems to be the first file of ember-simple-auth which needs this import.
* The ember-simple-auth 401 `DS.AdapterError` isn't the same as the ember-data 401 `DS.AdapterError` (at least with the current implementation)

marcoow:
@jayeff: sorry this took so long (and I'm changing my mind again). I think your original solution was correct - simply invalidate the session on 401 responses and always call `_super`. My original goal was to stop propagation of the error up the chain in Ember Data as it is already handled by invalidating the session. However, as that doesn't seem possible it's the best option to trigger session invalidation and then always call `_super` and let Ember Data do it's normal thing it does for 401 responses.


```js
handleResponse(status) {
  if (status === 401 && this.get('session.isAuthenticated')) {
  } else {
  }
}
```

jayeff:
No problem at all. I restored the old commit and rebased on current master.

I believe that an addon has no real possibility to stop the propagation of the error. Maybe it would be worth opening an issue on ember-data itself to discuss this point further.

marcoow:
> I believe that an addon has no real possibility to stop the propagation of the error. Maybe it would be worth opening an issue on ember-data itself to discuss this point further.

Yes, would be cool if you could go ahead and do that.

jayeff:
Thanks for your patience with my PR and for ember-simple-auth. :smile:

knownasilya:
Is this fix in 1.1?

marcoow:
Yes, this will be released with 1.1.

drauschenbach:
I'm trying to add a 403 error handler using a custom data adapter mixin, but can't seem to avoid a page reload which leads to a loss of error data and inability to display it. Since my data adapter is not invalidating the session, something else is forcing the page reload that I can't seem to control.

PR #898, opened 2016-02-12
Update CONTRIBUTING.md

jayeff:
Remove link to Building section on README as it no longer exists.
Document how to run tests in this file.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -18,8 +18,8 @@ when submitting issues:
 3. Update to the latest releases of Ember.js, jQuery, Ember Data if possible
 4. Include as much information as possible - this includes full stack traces
   etc.
-5. Set up a demo that demonstrates the issue on [JSFiddle](http://jsfiddle.net)
-  or [JSBin](http://jsfiddle.net). If your project is open source, provide a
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -18,8 +18,8 @@ when submitting issues:
 3. Update to the latest releases of Ember.js, jQuery, Ember Data if possible
 4. Include as much information as possible - this includes full stack traces
   etc.
-5. Set up a demo that demonstrates the issue on [JSFiddle](http://jsfiddle.net)
-  or [JSBin](http://jsfiddle.net). If your project is open source, provide a
+5. Set up a demo that demonstrates the issue on [JSFiddle](https://jsfiddle.net/)
+  or [JSBin](https://jsbin.com/). If your project is open source, provide a
```
marcoow: Not sure whether it's actually possible to use ESA on https://ember-twiddle.com but if so that would be an even better alternative of course.

jayeff: I did a Google search about this yesterday. It seems as if they plan to support addons but it's not possible yet: ember-cli/ember-twiddle#22

Otherwise I would suggest to replace jsfiddle/jsbin with ember-twiddle as well.

PR #920, opened 2016-03-02
Basic FastBoot support

rlivsey:
This is the minimum set of changes I had to make for our app to boot & run under FastBoot.

* use ephemeral store if running under FastBoot
* lazily evaluate protocol in cookie store so it doesn’t break under node

### DIFF
--- a/addon/session-stores/adaptive.js
+++ b/addon/session-stores/adaptive.js
@@ -3,6 +3,7 @@ import Ember from 'ember';
 import Base from 'ember-simple-auth/session-stores/base';
 import LocalStorage from 'ember-simple-auth/session-stores/local-storage';
 import Cookie from 'ember-simple-auth/session-stores/cookie';
+import Ephemeral from 'ember-simple-auth/session-stores/ephemeral';
 
 const { computed } = Ember;
 
@@ -82,7 +83,9 @@ export default Base.extend({
     this._super(...arguments);
 
     let store;
-    if (this.get('_isLocalStorageAvailable')) {
+    if (typeof FastBoot !== "undefined") {
+      store = this._createStore(Ephemeral);
+    } else if (this.get('_isLocalStorageAvailable')) {
       const options = { key: this.get('localStorageKey') };
       store = this._createStore(LocalStorage, options);
     } else {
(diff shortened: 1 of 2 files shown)


quaertym:
:+1:

rlivsey:
I'm not sure what the best way to test FastBoot is at the moment, at the least I should add tests around the Ephemeral store fallback. 

If happy with the general direction, I'll flesh it out a bit with more tests.

marcoow:
Thanks for the effort @rlivsey!

I think falling back to the ephemeral store isn't going to work though as the app might also want to change the session when running inside Fastboot in which case it would have to use the cookie store as that's the only way to propagate changes back to the client. Of course the cookie store would work differently in the context of Fastboot though (it would use `document.cookie` when running in the browser and request/response when running in the context of Fastboot (which is running as an express app)).

I think that we will maybe make the cookie store the default with ESA 2.0 as that's the only one that's going to work with Fastboot out of the box.

rlivsey:
@marcoow that makes sense.

I assume we should be using `fastboot.cookies` for this, so the cookie store would check to see if it's running in FastBoot and use that service instead?

marcoow:
Yes, but `fastboot.cookies` is readonly. The `fastboot` service also exposes the request and response as of this pull request: https://github.com/ember-fastboot/ember-fastboot-server/pull/19. I think cookies can be written via the response in express.

rlivsey:
Cool, I'll take a look at that. It would be nice if there was some first class API in Ember (cookies service?) which would paper over this and "do the right thing" in each environment so the addon doesn't have to care, sure that'll shake out over time.

marcoow:
The idea was to add an `ember-cookies` addon or so that abstracts that. I'm planning to implement a first version of that this week.

marcoow:
BTW: another thing that would have to changed for Fastboot is this line: https://github.com/simplabs/ember-simple-auth/blob/master/addon/mixins/application-route-mixin.js#L111. That would have to result in a redirect instead in Fastboot.

pangratz:
> I'm not sure what the best way to test FastBoot is at the moment, at the least I should add tests around the Ephemeral store fallback.

@rlivsey take a look at https://github.com/tomdale/ember-cli-addon-tests

marcoow:
The FastBoot efforts are now handled in #1035

Issue #916, opened 2016-02-29
Explicitly start store syncing


marcoow:
Currently stores start syncing right when they are initialized. This lead to #908 which is solved via #915 but that fix isn't really a nice solution. Instead store syncing should be started explicitly by the session once it has been initially restored.

This can be implemented in a backwards compatible way so that a deprecation would be triggered if the session store doesn't implement the respective method.

marcoow:
An alternative approach to fix this would actually be to implement a state machine in the internal session that ignores store events when the session is currently authenticating or invalidating.

marcoow:
This will be tackled in the next release as it requires some bigger changes that shouldn't be rushed into 1.1 if not necessary.

marcoow:
cleaning up old issues

PR #931, opened 2016-03-16
Fix #927

### LINKED ISSUE #927: {{session.isAuthenticated}} always returns false.
{{session.isAuthenticated}} always returns false when used in a component template in ember-simple-auth 1.1.0-beta.1 through 1.1.0-beta.3 starting with commit 6e55271. 

My set up is as follows (component and template): 
```
import Ember from 'ember';

export default Ember.Component.extend({
  session: Ember.inject.service('session'),
  actions: {
    invalidateSession() {
    },
    authenticateSessionWithGoogle() {
      this.get('session').authenticate('authenticator:torii', 'google-oauth2').catch((reason) => {
        alert(reason.error || reason);
      }); 
    }   
  }
});
```
```
  {{#if session.isAuthenticated}}
    <a {{action 'invalidateSession'}}>Logout</a>
  {{else}}
    <a {{action 'authenticateSessionWithGoogle'}}>Login with Google</a>
  {{/if}}
```
Even with authentication successful, `session.isAuthenticated` does not evaluate true. The commit before 6e55271, that is 1915445, evaluates to true upon successful authentication.

the dummy app works with 1.0.1-beta.3. Not sure what the dummy app is doing differently from my app...the dummy app does some re-routing which I don't have.

I've created a sample app which recreates the issue: https://github.com/bdiz/ember-simple-auth-bug

Initial bug conversation can be seen here: https://github.com/simplabs/ember-simple-auth/issues/874

rmachielse:
This pr prevents the `localStorage` store from triggering the `sessionDataUpdated` event when other `localStorage` keys then the one used by ember-simple-auth have been updated. Should fix #927.

### DIFF
--- a/addon/session-stores/local-storage.js
+++ b/addon/session-stores/local-storage.js
@@ -82,13 +82,15 @@ export default BaseStore.extend({
   },
 
   _bindToStorageEvents() {
-    Ember.$(window).bind('storage', () => {
-      this.restore().then((data) => {
-        if (!objectsAreEqual(data, this._lastData)) {
-          this._lastData = data;
-          this.trigger('sessionDataUpdated', data);
-        }
-      });
+    Ember.$(window).bind('storage', (e) => {
+      if (e.originalEvent.key === this.key) {
+        this.restore().then((data) => {
+          if (!objectsAreEqual(data, this._lastData)) {
+            this._lastData = data;
+            this.trigger('sessionDataUpdated', data);
+          }
+        });
+      }
     });
   }
 });


marcoow:
Can you add a test for this (if it's even possible to trigger `localStorage` events manually)?

marcoow:
Also maybe we actually need sth. like this as well: #916 The fix that I made to the cookie store (https://github.com/simplabs/ember-simple-auth/commit/df20ac5193e7bdd35eabe6dbb1018349ad1a7b46) seems a bit hacky actually and I'm not quite sure it's actually going to work with an authenticator that has a `restore` method that doesn't resolve instantly.

rmachielse:
@marcoow I'll see if I can test this. #916 would be a separate feature right, it shouldn't hold this pr?

marcoow:
@rmachielse: yes, #916 would be separate but I think that might actually fix the root cause for the problems around session stores that we saw in the last days/weeks.

marcoow:
Sorry for the delay!

PR #935, opened 2016-03-23
Upgrade Ember-CLI to 2.4.2

clekstro:
This upgrades app dependencies and renames a few files to conform with Ember-CLI's expectations. Should quiet down some of the upgrade noise in the future.


marcoow:
Thanks for this! There's also an in-progress pull request for the same thing: #918 but unfortunately that depends on a bug in ember-cli-blueprint-test-helpers being fixed - not sure whether that's been done yet. Would be great if you could follow-up on that!

clekstro:
Ha, sorry, should totally should have checked first for something similar. This was just part of looking into the test helpers things we spoke about the other day.

clekstro:
@marcoow Yeah, it looks like [that issue](https://github.com/ember-cli/ember-cli-blueprint-test-helpers/pull/35) is still open. FWIW, my branch passes locally, but is having issues on Travis for some reason. Should I try to investigate and get the build passing (so this could be merged without the helpers changes), or should we wait for that to get fixed downstream and close this in the interim?

marcoow:
@clekstro: it's failing on travis because there we also run the generator tests (which are not run when you only run `ember test`) and those are failing due to the issue in ember-cli-blueprint-test-helpers. I wouldn't really want to disable those so we'll have to wait until a new version of ember-cli-blueprint-test-helpers gets released - of course we might be able to help with that.

clekstro:
Ok, gotcha. I can close this then. No sense in duplicate PR's.

Issue #226, opened 2014-06-27
Add client_id and client_secret to OAuth2 config


dustinfarris:
It seems increasingly difficult to reopen `makeRequest` in ember-cli.  I realize that client_secret is mostly redundant in a browser application, but some OAuth2 servers expect it regardless of whether the client is public or confidential.

Would it be possible to just add these specific items to the payload (or set the appropriate header) if they are present in the user's config?

mpash:
This is part of the spec for OAuth, it needs to be implemented. You're not on the OAuth team so you don't get to cherry pick how you think something should be implemented or what designs jive with you. That's not helping anyone but your own self interests.

PR #937, opened 2016-03-23
use url encoded  instead of setting a header

timini:
from https://aaronparecki.com/2012/07/29/2/oauth2-simplified

`
POST https://api.oauth2server.com/token
    grant_type=password&
    username=USERNAME&
    password=PASSWORD&
    client_id=CLIENT_ID
`

maybe related to #226

### DIFF
--- a/addon/authenticators/oauth2-password-grant.js
+++ b/addon/authenticators/oauth2-password-grant.js
@@ -155,7 +155,11 @@ export default BaseAuthenticator.extend({
   */
   authenticate(identification, password, scope = []) {
     return new RSVP.Promise((resolve, reject) => {
+      const client_id = this.get('clientId');
       const data                = { 'grant_type': 'password', username: identification, password };
+      if (!Ember.isEmpty(client_id)) {
+        data.client_id = client_id;
+      }
       const serverTokenEndpoint = this.get('serverTokenEndpoint');
       const scopesString = Ember.makeArray(scope).join(' ');
       if (!Ember.isEmpty(scopesString)) {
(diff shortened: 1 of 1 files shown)


marcoow:
Is this part of RFC 6749? If so this could be opt-in but the default of sending it in the `Authorization` header cannot be changed of course as that would break all apps out there using the OAuth 2 password authenticator.

timini:
Yep

https://tools.ietf.org/html/rfc6749#section-2.3.1

```
Alternatively, the authorization server MAY support including the
   client credentials in the request-body using the following
   parameters:

   client_id
         REQUIRED.  The client identifier issued to the client during
         the registration process described by Section 2.2.

   client_secret
         REQUIRED.  The client secret.  The client MAY omit the
         parameter if the client secret is an empty string.
```

timini:
I was using https://github.com/evonove/django-oauth-toolkit a popular oauth library, and this is the way it is implemented there.

marcoow:
OK, seems good then. That should be opt-on though so the default behavior doesn't change in the authenticator. I think introducing a new property `sendClientIdInRequestBody` which defaults to `false` would be the way to go.

timini:
Yep, that seems like a good way to go. Ill give it a go.

tylerneu:
I too am using Django OAuth Toolkit and would like to see this change made. I'll work around it until then.

marcoow:
@timini: any updates on this? Would be great to get this in.

marcoow:
superseded by #1165

PR #941, opened 2016-03-29
fix deprecation warning Ember.assign for 2.5.0

gertjanwytynck:


urbany:
This should be enough @gertjanwytynck 
The test fail seems unrelated:
``Cannot find module 'lodash/lang/isFunction'``

PR #943, opened 2016-03-30
seperated transition block as an action on sessionAuthenticated

apokinsocha:
Take this example for reason for PR:

```
    sessionAuthenticated() {
        let _this = this;
        this._loadSomething().catch(function() {
            _this.get('session').invalidate();
        }).finally(function() {
            _this._super(...arguments);
        });

    }
```

The call to _this._super(...arguments) in the finally block is not possible, since Ember does not support calling super on asynchronous blocks, only synchronous blocks. The only thing that is left in this case is to copy and paste the transition code from the ember-simple-auth sessionAuthenticated method, not very convenient for someone who wants to extend while resolving promise/s.

```
    sessionAuthenticated() {
        let _this = this;
        this._loadSomething().catch(function() {
            _this.get('session').invalidate();
        }).finally(function() {
            _this.send('authenticatedTransition');//->much more convenient
        });

    }

```

### DIFF
--- a/addon/mixins/application-route-mixin.js
+++ b/addon/mixins/application-route-mixin.js
@@ -66,23 +66,14 @@ export default Ember.Mixin.create({
   /**
     This method handles the session's
     {{#crossLink "SessionService/authenticationSucceeded:event"}}{{/crossLink}}
-    event. If there is a transition that was previously intercepted by
-    {{#crossLink "AuthenticatedRouteMixin/beforeModel:method"}}the
-    AuthenticatedRouteMixin's `beforeModel` method{{/crossLink}} it will retry
-    it. If there is no such transition, this action transitions to the
-    {{#crossLink "Configuration/routeAfterAuthentication:property"}}{{/crossLink}}.
+    event.
 
     @method sessionAuthenticated
     @public
   */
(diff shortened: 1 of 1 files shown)


marcoow:
You should be able to do what you want as

```js
sessionAuthenticated() {
  let super = this._super;
  this._loadSomething().catch(() => {
  }).finally(() => {
    super(...arguments);
  });
}
```

apokinsocha:
That's great, wasn't familiar with that. I guess I will close the PR in that case.

apokinsocha:
changed from:

`let super = this._super;`

to:

`let _super = this._super;`

got rid of one error, but still does not call to transition.

apokinsocha:
SO related thread: http://stackoverflow.com/questions/25902505/calling-an-ember-super-method-from-promise-handler

marcoow:
Ah sorry, you need to `bind` actually:

```js
sessionAuthenticated() {
  let super = this._super.bind(this);
  this._loadSomething().catch(() => {
  }).finally(() => {
    super(...arguments);
  });
}
```

apokinsocha:
great, thank you (p.s; great library).

Issue #375, opened 2014-12-06
Updated auth token on every api request


solsberg:
I am looking for suggestions on how to support the following scenario. 
My server currently returns a new auth token on every API request which I need to capture from the response headers each time to then use for authorization of the subsequent request. So I can't simply capture the response headers from the initial authentication request and then use the same token for each request in the authorizer. 
The ember-simple-auth authorizer API is able to hook into each API request by providing a hook into the ajax prefilter but it doesn't seem to be able to hook into the ajax response chain in order to capture the updated token and store it for the next request.
Do you see a good way of supporting this? Thanks

marcoow:
Sure, that works as well. I'm not sure whether it's worth adding an abstraction for this to Ember Simple Auth as it's sth. that's rarely used and is easily done via an `.ajaxSuccess` callback in an initializer.

achan:
Any way to do this in 1.0?

marcoow:
@achan: the best way would probably be to do that in your Ember Data adapter's `handleResponse` method.

jpakulski:
@marcoow I am probably missing something but I can not find a way to update the session data as mentioned above.

Our app does not use Ember Data. It does make an ajax call to a Rails API. DeviseTokenAuth is used to secure the API. The token changes on every request.

I added an the following instance initializer (app/instance-initializers/ajax-success.coffee):

```coffeescript
initialize = (applicationInstance)->
  session = applicationInstance.lookup('service:session')

  $(document).ajaxSuccess((event, xhr)=>
    accessToken = xhr.getResponseHeader('access-token')
    client = xhr.getResponseHeader('client')

    if accessToken? and client?
      authenticated = session.get('data.authenticated')
      authenticated.accessToken = accessToken
      authenticated.client = client

      session.set('data.authenticated', authenticated)
      # Assertion Failed: "authenticated" is a reserved key used by Ember Simple Auth!
  )

AjaxSuccessInitializer =
  name: 'ajax-success'
  initialize: initialize

`export {initialize}`
`export default AjaxSuccessInitializer`
```

This fails because data.content is read only.

I can do all sorts of ugly things to get this to work, least of which is: storing the new token in a different property and overwrite the authorizer, but that seems like a lot of work.There must be a better way? Is there a way to do this?

Thank You

marcoow:
You cannot set the `authenticated` section of the session manually. You can set the token though via e.g. `session.set('data.authenticated.<token-property>', accessToken)`.

[...]

PR #946, opened 2016-04-02
Added Authenticator and Authoriser to support devise-token-auth

jaijaz:
I'm not sure what your plans are in terms of extending ember-simple-auth with alternate server side authentication methods, but since you have a devise one currently, and devise gives devise-token-auth as an alternative now that it has removed token auth from it's repo, I thought I would offer you this PR in case you wanted to include it.

Perhaps however you would prefer to have addons that use ember-simple-auth as dependancy?

In case you do wish to use this, it is all relatively straight forward. There is a new mixin that overrides the standard one for the data adaptors to allow for updating the access-token and expiry after every call. It does however have a hack in there to force the update of the session data as documented in issue #375.


addon/authorizers/devise-token-auth.js
```diff
@@ -0,0 +1,135 @@
...
+        && !isEmpty(clientHeaderAttribute)
+        && !isEmpty(expiryHeaderAttribute)
+        && !isEmpty(uidHeaderAttribute)
+        && !isEmpty(tokenHeaderTypeAttribute)
+        && parseInt(expiryHeaderAttribute) > Date.now()/1000) {
+      block(tokenHeaderAttributeName, tokenHeaderAttribute);
+      block(clientHeaderAttributeName, clientHeaderAttribute);
+      block(expiryHeaderAttributeName, expiryHeaderAttribute);
+      block(uidHeaderAttributeName, uidHeaderAttribute);
+      block(tokenHeaderTypeAttributeName, tokenHeaderTypeAttribute);
```
marcoow: The block can only be called once. Otherwise code like this:

```js
this.get('session').authorize('authorizer:devise-token-auth', () => {
  // make request
})
```

would result in 5 requests being sent.

addon/mixins/devise-token-auth-data-adapter-mixin.js
```diff
@@ -0,0 +1,60 @@
...
+    {{#crossLink "SessionService/invalidate:method"}}{{/crossLink}}).
+
+    @method handleResponse
+    @param {Number} status The response status as received from the API
+    @param {Array} headers The response headers as received from the API
+    @protected
+  */
+  handleResponse(status, headers) {
+    if ((status === 401 && this.get('session.isAuthenticated'))
+        || (this.get('session.data.authenticated.uid') != headers['uid'] || this.get('session.data.authenticated.client') != headers['client'] )
```
marcoow: The session data shouldn't be modified from outside of the session/authenticator actually. There would have to be a better way/public API to do things like this if this is required.

marcoow:
Thanks. I'd probably be fine with merging this. However, there is no public API yet for updating the session data from outside the session/authenticator.

Also currently there are a bunch of open issues and 1.1 which is still in beta. These things have to be fixed first before we add new features like this.

jaijaz:
Thanks for your comments. I agree, editing the session the way I am isn't ideal and a proper API would be better.

I have since found a bug in this too where a series of quick API calls can mess up the token state and accidentally log you out.

Going to close this PR and wait for updates. If I get time, I will see if I can't help with some of it.

PR #955, opened 2016-04-10
Stop the baseURL config property from being overridden

JoshTumath:
Ember app authors can currently specify a `baseURL` in the `config/environment.js` file. e.g.

```javascript
ENV['ember-simple-auth'].baseURL = 'http://localhost:4200';
```

However, the configuration is being overridden because of the following line in the `app/initializers/ember-simple-auth.js` file:

```javascript
const config   = ENV['ember-simple-auth'] || {};
config.baseURL = ENV.baseURL;
```

This simple patch removes the overriding expression so that the author's `baseURL` can be used.

### DIFF
--- a/app/initializers/ember-simple-auth.js
+++ b/app/initializers/ember-simple-auth.js
@@ -7,12 +7,8 @@ import setupSessionService from 'ember-simple-auth/initializers/setup-session-se
 export default {
   name:       'ember-simple-auth',
   initialize: function(registry) {
-    const config = ENV['ember-simple-auth'] || {};
-    
-    // We need to use the default Ember baseURL if one hasn't been set
-    if (!(ENV['ember-simple-auth'] && ENV['ember-simple-auth'].baseURL)) {
-      config.baseURL = ENV.baseURL;
-    }
+    const config   = ENV['ember-simple-auth'] || {};
+    config.baseURL = config.baseURL || ENV.baseURL;
     
     Configuration.load(config);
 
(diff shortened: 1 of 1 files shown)


marcoow:
👍🏻

This removes the default though, thus would be a breaking change.

JoshTumath:
> This removes the default though, thus would be a breaking change.
Ah, I assumed you got the default was hardcoded. I have uploaded a corrected patch that falls back to `ENV.baseURL` if the baseURL config isn't present.

marcoow:
I'm not sure how that is different from the original code now.

JoshTumath:
> I'm not sure how that is different from the original code now.

So there are two places where base URLs are being stored in the `config/environment.js` file. We have `ENV.baseURL` and we have `ENV['ember-simple-auth'].baseURL`. The former stores a base URL used by Ember.js and the latter stores a base URL specific to this library.

With the way it worked in the original code, even if you set a value for `ENV['ember-simple-auth'].baseURL`, it was being overridden.

marcoow:
Ah sure, I see. So I guess what we want is this?

```js
const config   = ENV['ember-simple-auth'] || {};
config.baseURL = config.baseURL || ENV.baseURL;
```

JoshTumath:
That seems much simpler. I've updated the patch.

marcoow:
Sorry - just realized the version that's on master is actually correct. The `baseURL` property in Ember Simple Auth's configuration object is actually private API and only used to track the `baseURL` as configured for the complete application in `config/environment.js`.

PR #957, opened 2016-04-12
[WIP] Validate server responses in authenticators

### LINKED ISSUE #942: Validate server responses in authenticators
Currently most (if not all) authenticators simply resolve the returned promise on any successful server response regardless of whether the response actually contains the data that's later being checked for in the `restore` method then. That leads to errors where people seem to be successfully logged in only until they reload when the `restore` method rejects the session as invalid. Validating presence of the respective data in the `authenticate` method already is simple and would solve a whole class of problems that people actually have.

stevenwu:
@marcoow how does this look?

I don't know how to catch the errors from `authenticate` inside tests, if throwing errors is what we want.

Fix #942

### DIFF
--- a/addon/authenticators/devise.js
+++ b/addon/authenticators/devise.js
@@ -104,10 +104,10 @@ export default BaseAuthenticator.extend({
 
       return this.makeRequest(data).then(
         (response) => {
-          if (!this._validate(response)) {
-            run(null, reject, `Check that server response includes ${tokenAttributeName} and ${identificationAttributeName}`);
-          } else {
+          if (this._validate(response)) {
             run(null, resolve, response);
+          } else {
+            run(null, reject, `Check that server response includes ${tokenAttributeName} and ${identificationAttributeName}`);
           }
         },
         (xhr) => run(null, reject, xhr.responseJSON || xhr.responseText)
--- a/addon/authenticators/devise.js
+++ b/addon/authenticators/devise.js
@@ -74,15 +74,7 @@ export default BaseAuthenticator.extend({
     @public
   */
   restore(data) {
-    const { tokenAttributeName, identificationAttributeName } = this.getProperties('tokenAttributeName', 'identificationAttributeName');
-    const tokenAttribute = get(data, tokenAttributeName);
-    const identificationAttribute = get(data, identificationAttributeName);
-
-    if (!isEmpty(tokenAttribute) && !isEmpty(identificationAttribute)) {
-      return Promise.resolve(data);
-    } else {
-      return Promise.reject();
-    }
+    return this._validate(data) ? Promise.resolve(data) : Promise.reject();
   },
 
   /**
--- a/addon/authenticators/devise.js
+++ b/addon/authenticators/devise.js
@@ -105,13 +105,21 @@ export default BaseAuthenticator.extend({
   */
   authenticate(identification, password) {
     return new Promise((resolve, reject) => {
-      const { resourceName, identificationAttributeName } = this.getProperties('resourceName', 'identificationAttributeName');
+      const { resourceName, identificationAttributeName, tokenAttributeName } = this.getProperties('resourceName', 'identificationAttributeName', 'tokenAttributeName');
       const data         = {};
       data[resourceName] = { password };
       data[resourceName][identificationAttributeName] = identification;
 
       return this.makeRequest(data).then(
-        (response) => run(null, resolve, response[resourceName] || response),
+        (response) => {
+          if (isEmpty(response[tokenAttributeName])) {
+            throw new Error(`${tokenAttributeName} is missing in server response`);
+          }
+          if (isEmpty(response[identificationAttributeName])) {
+            throw new Error(`${identificationAttributeName} is missing in server response`);
+          }
+          run(null, resolve, response);
+        },
         (xhr) => run(null, reject, xhr.responseJSON || xhr.responseText)
       );
     });
--- a/addon/authenticators/devise.js
+++ b/addon/authenticators/devise.js
@@ -156,10 +156,8 @@ export default BaseAuthenticator.extend({
     const tokenAttributeName = this.get('tokenAttributeName');
     const identificationAttributeName = this.get('identificationAttributeName');
     const resourceName = this.get('resourceName');
-    if (data[resourceName]) {
-      return !isEmpty(data[resourceName][tokenAttributeName]) && !isEmpty(data[resourceName][identificationAttributeName]);
-    } else {
-      return !isEmpty(data[tokenAttributeName]) && !isEmpty(data[identificationAttributeName]);
-    }
+    const _data = data[resourceName] ? data[resourceName] : data;
+
+    return !isEmpty(_data[tokenAttributeName]) && !isEmpty(_data[identificationAttributeName]);
   }
 });
--- a/addon/authenticators/devise.js
+++ b/addon/authenticators/devise.js
@@ -105,7 +105,9 @@ export default BaseAuthenticator.extend({
       return this.makeRequest(data).then(
         (response) => {
           if (this._validate(response)) {
-            run(null, resolve, response);
+            const resourceName = this.get('resourceName');
+            const _response = response[resourceName] ? response[resourceName] : response;
+            run(null, resolve, _response);
           } else {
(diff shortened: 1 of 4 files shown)


marcoow:
@stevenwu: thanks, definitely goes in the right direction. The `authenticate` methods should not throw errors though but reject the returned promise with an error object instead, e.g.

```js
if (isEmpty(response[tokenAttributeName])) {
  run(null, reject, `${tokenAttributeName} is missing in server response`);
}
```

instead of

```js
if (isEmpty(response[tokenAttributeName])) {
  throw new Error(`${tokenAttributeName} is missing in server response`);
}
```

That way you can test that easily by handling the rejected promise in the test and assert on the value of the error object, e.g.:

```js
it('fails if no token is present', (done) => {
  authenticator.authenticate('username', 'password').catch((error) => {
    expect(error).to.eql('token is missing!');
  });
});
```

Finally the validation logic should be extracted into its own method (e.g. `_validate` or so) that would also be used in the `restore` methods then as the same validation logic should be applied in both `authenticate` and `restore`.

addon/authenticators/oauth2-password-grant.js
```diff
@@ -164,24 +161,19 @@ export default BaseAuthenticator.extend({
...
           const refreshAccessTokens = this.get('refreshAccessTokens');
           if (refreshAccessTokens) {
-            if (isEmpty(response['expires_in'])) {
-              throw new Error('expires_in is missing in server response')
-            }
-            if (isEmpty(response['refresh_token'])) {
-              throw new Error('refresh_token is missing in server response')
-            }
+            this._validate(reject, response, 'expires_in');
+            this._validate(reject, response, 'refresh_token');
```
stevenwu: it looks like 
```
{ error: 'invalid_grant' }
```
is returned elsewhere before `refresh_token` is validated. Maybe we can remove this?

addon/authenticators/oauth2-password-grant.js
```diff
@@ -164,11 +161,19 @@ export default BaseAuthenticator.extend({
       }
       this.makeRequest(serverTokenEndpoint, data).then((response) => {
         run(() => {
+          this._validate(reject, response, 'access_token');
+          const refreshAccessTokens = this.get('refreshAccessTokens');
+          if (refreshAccessTokens) {
+            this._validate(reject, response, 'expires_in');
+            this._validate(reject, response, 'refresh_token');
```
stevenwu:
``` 
{ error: 'invalid_grant' }
```
is returned elsewhere if 'expires_in' and 'access_token' are already in the response. Maybe we can remove this?

addon/authenticators/devise.js
```diff
@@ -81,6 +81,8 @@ export default BaseAuthenticator.extend({
     if (!isEmpty(tokenAttribute) && !isEmpty(identificationAttribute)) {
       return Promise.resolve(data);
     } else {
+      this._validate(Promise.reject, data, tokenAttributeName);
+      this._validate(Promise.reject, data, identificationAttributeName);
       return Promise.reject();
```
marcoow: Might be easier to return a boolean from `_validate` and get `tokenAttributeName` and `identificationAttributeName` internally in the method so that you could just do

```js
if (!this._validate(data)) {
  return Promise.reject();
}
```

addon/authenticators/oauth2-password-grant.js
marcoow: If `{ error: 'invalid_grant' }` nothing else should be in the response actually. But in this case the response code would also be 422 or so so that the request's success handler should never be called anyway. I think we only need to validate successful server responses as erroneous responses will already reject the returned promise anyway.

stevenwu:
I forgot to mention this in the call. Occasionally, I'll see 
```
Error: timeout of 2000ms exceeded. Ensure the done() callback is being called in this test.
```
even though `done()` is called in the tests.

This error showed up after I made the latest refactor.

marcoow:
Not sure what might be causing the timeout actually. Instead of using `done` you can return the promise though - it's a bit more concise anyway, e.g.:

```js
describe('when the server returns incomplete data', () => {
  it('fails when no access_token is present', () => {
    server.respondWith('POST', '/token', [
        200,
        { 'Content-Type': 'application/json' },
        '{}'
    ]);

    return authenticator.authenticate('username', 'password').catch((error) => {
      expect(error).to.eql('access_token is missing in server response');
    });
  });
});
```

addon/authenticators/devise.js
```diff
@@ -105,13 +97,19 @@ export default BaseAuthenticator.extend({
...
       data[resourceName][identificationAttributeName] = identification;
 
       return this.makeRequest(data).then(
-        (response) => run(null, resolve, response),
+        (response) => {
+          if (!this._validate(response)) {
+            run(null, reject, `Check that server response includes ${tokenAttributeName} and ${identificationAttributeName}`);
+          } else {
+            run(null, resolve, response);
+          }
```
marcoow: I'd turn this around for better readability:

```js
if (this._validate(response)) {
  run(null, resolve, response);
} else {
 run(null, reject, `Check that server response includes ${tokenAttributeName} and ${identificationAttributeName}`);
}
```

addon/authenticators/oauth2-password-grant.js
```diff
@@ -164,11 +164,16 @@ export default BaseAuthenticator.extend({
       }
       this.makeRequest(serverTokenEndpoint, data).then((response) => {
         run(() => {
+          if (!this._validate(response)) {
+            run(null, reject, 'access_token is missing in server response');
+          }
```
marcoow: This code is already wrapped in `run`, see line 166.

marcoow:
@stevenwu: this should be ready to merge after rebasing on `master`

stevenwu:
Should we always expect the response to contain a resource name? Is it ever 
`{ "token": "secret token!", "email": "email@address.com" }`

When we validate the data and return it, should it be returned unchanged? I noticed https://github.com/simplabs/ember-simple-auth/blob/2759f6d72b6d5a326e4079b7c882bebbc609c56d/tests/unit/authenticators/devise-test.js#L140 does not contain the resource name.

marcoow:
> Should we always expect the response to contain a resource name? Is it ever 
{ "token": "secret token!", "email": "email@address.com" }

The resource name is optional. Both `{ "token": "secret token!", "email": "email@address.com" }` and `{ user: { "token": "secret token!", "email": "email@address.com" } }` might be valid depending on whether a resource name was configured.

> When we validate the data and return it, should it be returned unchanged? I noticed https://github.com/simplabs/ember-simple-auth/blob/2759f6d72b6d5a326e4079b7c882bebbc609c56d/tests/unit/authenticators/devise-test.js#L140 does not contain the resource name.

Validation should not modify the data but the `authenticate` method will only resolve with the unscoped data (e.g. with `{ "token": "secret token!", "email": "email@address.com" }` both for response `{ "token": "secret token!", "email": "email@address.com" }` as well as for `{ user: { "token": "secret token!", "email": "email@address.com" } }`).

stepantubanov:
Thanks so much for this, I've spent an hour trying to figure out why restoring the session was not working and it turned out that the problem was earlier (when logging in). Failing fast can definitely save a lot of debugging time.

What do you think about adding "_validate" call in the OAuth2 authorizer as well? And what else is left to finish it up? I'd be happy to help.

addon/authenticators/devise.js
```diff
@@ -150,5 +150,16 @@ export default BaseAuthenticator.extend({
...
+
+  _validate(data) {
+    const tokenAttributeName = this.get('tokenAttributeName');
+    const identificationAttributeName = this.get('identificationAttributeName');
+    const resourceName = this.get('resourceName');
+    if (data[resourceName]) {
+      return !isEmpty(data[resourceName][tokenAttributeName]) && !isEmpty(data[resourceName][identificationAttributeName]);
+    } else {
+      return !isEmpty(data[tokenAttributeName]) && !isEmpty(data[identificationAttributeName]);
+    }
```
marcoow: I think this would be simpler as

```js
const resourceName = this.get('resourceName');
const _response = response[resourceName] ? response[resourceName] : response;

return !isEmpty(response[tokenAttributeName]) && !isEmpty(response[identificationAttributeName]);
```

marcoow:
@stevenwu: just one small comment - I swear that will be the last ;)

marcoow:
awesome 🎉

PR #958, opened 2016-04-12
Update docs to show baseURL is internal

JoshTumath:
As discussed in #955, baseURL is an internal config parameter. Therefore, the documentation should be updated to reflect this.


marcoow:
It's publicly readable but not writable so that's all fine actually. The fact that the property is read-only should actually be exposed in the API docs though: http://ember-simple-auth.com/api/classes/Configuration.html#property_baseURL

PR #963, opened 2016-04-14
Add Before Callbacks to Authenticators

BenMorganIO:
While I was programming, I noticed that I wanted to transform some of the data and save it as a record in the data store. To do this, I had to copy the source code and then add then modify the code. It then occurred to me that this project could benefit from having before callbacks.

The purpose of the before callbacks would allow developers to touch the data before a restore, reject, or authentication happens.

I'm just getting started with Ember, so this PR is more of a concept/discussion generator than anything else.

### DIFF
--- a/addon/authenticators/devise.js
+++ b/addon/authenticators/devise.js
@@ -60,6 +60,11 @@ export default BaseAuthenticator.extend({
   */
   identificationAttributeName: 'email',
 
+
+  beforeRestore: (data) => { return data; },
+  beforeReject: (data) => { return data; },
+  beforeAuthenticate: (response) => { return response; },
+
   /**
     Restores the session from a session data object; __returns a resolving
     promise when there are non-empty
@@ -79,8 +84,10 @@ export default BaseAuthenticator.extend({
     const identificationAttribute = get(data, identificationAttributeName);
 
     if (!isEmpty(tokenAttribute) && !isEmpty(identificationAttribute)) {
+      data = beforeRestore(data);
       return Promise.resolve(data);
     } else {
+      data = beforeReject(data);
       return Promise.reject();
     }
   },
@@ -111,7 +118,10 @@ export default BaseAuthenticator.extend({
       data[resourceName][identificationAttributeName] = identification;
 
       return this.makeRequest(data).then(
-        (response) => run(null, resolve, response),
+        (response) => {
+          response = beforeAuthenticate(response);
+          run(null, resolve, response);
+        },
         (xhr) => run(null, reject, xhr.responseJSON || xhr.responseText)
       );
     });


addon/authenticators/devise.js
```diff
@@ -111,7 +118,10 @@ export default BaseAuthenticator.extend({
       data[resourceName][identificationAttributeName] = identification;
 
       return this.makeRequest(data).then(
-        (response) => run(null, resolve, response),
+        (response) => {
+          response = beforeAuthenticate(response);
```
BenMorganIO: Now that I think about it, this is probably more of an `afterAuthenticate()`.

marcoow:
This would probably be solved much easier by extending the default devise authenticator and overriding `authenticate`, `restore` and `invalidate`, e.g.:

```js
```

BenMorganIO:
Hmmm I thought about doing that in my application, but I noticed that I needed to dig into the `data`/`response` information. Which mainly had to happen here: https://github.com/simplabs/ember-simple-auth/pull/963/files#diff-719e859d3a0317bb511a3b40c7ee9598R122.

marcoow:
So you can just override `makeRequest` then:

```js
makeRequest() {
  return this._super(...arguments).then((data) => {
    //modify data
    return data;
  });
}
```

BenMorganIO:
Hmm... This is essentially what I'm doing:

```javascript
  makeRequest(data) {
    return this._super(data).then(
      (response) => {
        response.data.type = 'user';
        this.get('store').pushPayload(response);

        response.email = response.data.attributes.email;
        response.token = response.data.attributes.token;
      },
      (xhr) => run(null, reject, xhr.responseJSON || xhr.responseText)
    );
  }
```

marcoow:
Thanks for the effort but I don't think before/after callbacks really add enough value to justify the increased complexity and growing API surface.

BenMorganIO:
Perhaps we could add some public methods instead? I still feel that "hooking" in is the best in my case. What about:

```javascript
restore(data) {
  ...

  if (!isEmpty(tokenAttribute) && !isEmpty(identificationAttribute)) {
    resolveRestore(data);
  } else {
    rejectRestore();
  }
},

authenticate(identification, password) {
  return new Promise((resolve, reject) => {
    ...

    return this.makeRequest(data).then(
      (response) => runAuthenticate(null, resolve, response),
      (xhr) => run(null, reject, xhr.responseJSON || xhr.responseText)
    );
  });
}

runAuthenticate() {
  Ember.run(null, resolve, response);
},

resolveRestore(data) {
  return Promise.resolve(data);
},

rejectRestore(data) {
  return Promise.reject();
}
```

This way, if you wanted to hook in, you could do:

```javascript
  runAuthenticate(resolve, response) {
    response.data.type = 'user';
    this.get('store').pushPayload(response);

    response.email = response.data.attributes.email;
    response.token = response.data.attributes.token;

    return this._super(resolve, response)
  },

  resolveRestore(data) {
    this.get('store').push(data);
    return this._super(data)
  }
```

marcoow:
What's wrong with overriding, calling `_super` and modifying input arguments before or return values after?

BenMorganIO:
I don't think I actually get the return values I would expect. So, maybe a few questions:

- Is there a way to tell if the promise returned by `restore()` is a `Promise.reject()` or a `Promise.resolve()`?
- Am I able to access the data in the `Promise.resolve()`?
- When I tried the `_super()` for `authenticate()` it broke the session storage. I was setting the `response.email` and `response.token` via the data attributes. It didn't seem to work :/. However, the `pushPayload(response)` worked fine...

marcoow:
You can do this to modify the data that the promise returned by `_super` resolves with:

```js
restore(data) {
  // do stuff before restore
  return this._super(...arguments).then((data) => {
    // do stuff after restore
    return data;
  });
}
```

BenMorganIO:
Since its 1pm now and not 1am for me, here was the work end result of the refactor you suggested @marcoow. Just leaving it here for historical reasons:

```javascript
import Ember from 'ember'
import Devise from 'ember-simple-auth/authenticators/devise';

export default Devise.extend({
  store: Ember.inject.service(),

  restore(data) {
    return this._super(data).then((data) => {
      this.get('store').push(data);
    });
  },

  makeRequest(data) {
    return this._super(data).then((response) => {
      response.data.type = 'user';
      this.get('store').pushPayload(response);

      response.email = response.data.attributes.email;
      response.token = response.data.attributes.token;

      return response;
    });
  }
});
```

PR #964, opened 2016-04-16
Fastboot compatibility	

arjansingh:
This will boot and serve the dummy app for me.

Outstanding issues are:

1. https://github.com/simplabs/ember-cookies/pull/14
2. Integrating fastboot with the dummy apps [server](https://github.com/simplabs/ember-simple-auth/tree/master/server)
3. Fastboot test cases.
4. Updating Fastboott references to be in line with https://github.com/ember-fastboot/ember-fastboot-server/pull/38

I still would like feedback on my implementation in case I am doing something too crazy.

### DIFF
--- a/README.md
+++ b/README.md
@@ -503,7 +503,9 @@ store if `localStorage` is available.
 
 [The Cookie store](http://ember-simple-auth.com/api/classes/CookieStore.html)
 stores its data in a cookie. This is used by the adaptive store if
-`localStorage` is not available.
+`localStorage` is not available. __This store must be used when the
+application uses
+[FastBoot](https://github.com/ember-fastboot/ember-cli-fastboot).__
 
 #### Ephemeral Store
 
@@ -551,6 +553,19 @@ export default Base.extend({
     …
   }
 });
+```
+
+## FastBoot
+
+Ember Simple Auth works with FastBoot out of the box as long as the Cookie
+session store is being used. In order to enable the cookie store, define it as
+the application store:
+
+```js
+// app/session-stores/application.js
+import CookieStore from 'ember-simple-auth/session-stores/cookie';
+
+export default CookieStore.extend();
 ```
 
 ## Testing
--- a/addon/initializers/setup-session.js
+++ b/addon/initializers/setup-session.js
@@ -11,5 +11,7 @@ export default function setupSession(registry) {
     store = 'session-store:test';
     registry.register(store, Ephemeral);
   }
+
+  inject(registry, store, 'cookies', 'service:cookies');
   inject(registry, 'session:main', 'store', store);
 }
--- a/addon/initializers/setup-session.js
+++ b/addon/initializers/setup-session.js
@@ -12,6 +12,5 @@ export default function setupSession(registry) {
     registry.register(store, Ephemeral);
   }
 
-  inject(registry, store, 'cookies', 'service:cookies');
   inject(registry, 'session:main', 'store', store);
 }
--- a/addon/initializers/setup-session.js
+++ b/addon/initializers/setup-session.js
@@ -12,5 +12,6 @@ export default function setupSession(registry) {
     registry.register(store, Ephemeral);
   }
 
+  inject(registry, store, 'cookies', 'service:cookies');
   inject(registry, 'session:main', 'store', store);
 }
--- a/addon/mixins/authenticated-route-mixin.js
+++ b/addon/mixins/authenticated-route-mixin.js
@@ -1,7 +1,8 @@
 import Ember from 'ember';
+import getOwner from 'ember-getowner-polyfill';
 import Configuration from './../configuration';
 
-const { service } = Ember.inject;
+const { inject: { service }, computed } = Ember;
 
 /**
   __This mixin is used to make routes accessible only if the session is
--- a/addon/mixins/unauthenticated-route-mixin.js
+++ b/addon/mixins/unauthenticated-route-mixin.js
@@ -1,4 +1,5 @@
 import Ember from 'ember';
+import getOwner from 'ember-getowner-polyfill';
 import Configuration from './../configuration';
 
 const { service } = Ember.inject;
--- a/addon/session-stores/cookie.js
+++ b/addon/session-stores/cookie.js
@@ -1,6 +1,7 @@
 import Ember from 'ember';
 import BaseStore from './base';
 import objectsAreEqual from '../utils/objects-are-equal';
+import getOwner from 'ember-getowner-polyfill';
 
 const { RSVP, computed, run: { next } } = Ember;
 
@@ -76,8 +77,24 @@ export default BaseStore.extend({
   */
   cookieExpirationTime: null,
 
-  _secureCookies: computed(function () {
-    return window ? window.location.protocol === 'https:' : _fastbootInfo.host().includes('https:');
+  _cookies: computed(function() {
+    let owner = getOwner(this);
+
+    return owner.lookup('service:cookies');
+  }),
+
+  _fastboot: computed(function() {
+    let owner = getOwner(this);
+
+    return owner.lookup('service:fastboot');
+  }),
+
+  _secureCookies: computed(function() {
+    if (this.get('_fastboot.isFastBoot')) {
+      return this.get('_fastboot._fastbootInfo.request.hostname').indexOf('https:') === 0;
+    } else {
+      return window.location.protocol === 'https:'
+    }
   }).volatile(),
 
   _syncDataTimeout: null,
@@ -85,18 +102,26 @@ export default BaseStore.extend({
   _renewExpirationTimeout: null,
 
   _isPageVisible: computed(function() {
-    const visibilityState = document ? document.visibilityState || 'visible' : false;
-    return visibilityState === 'visible';
+    if (this.get('_fastboot.isFastBoot')) {
+      return false;
+    } else {
+      const visibilityState = document.visibilityState || 'visible';
+      return visibilityState === 'visible';
+    }
   }).volatile(),
 
   init() {
     this._super(...arguments);
 
-    next(() => {
-      this._syncData().then(() => {
-        this._renewExpiration();
+    if (!this.get('_fastboot.isFastBoot')) {
+      next(() => {
+        this._syncData().then(() => {
+          this._renewExpiration();
+        });
       });
-    });
+    } else {
+      this._renew();
+    }
   },
 
   /**
@@ -145,7 +170,7 @@ export default BaseStore.extend({
   },
 
   _read(name) {
-    const value = this.get('cookies').read(name) || '';
+    const value = this.get('_cookies').read(name) || '';
     return decodeURIComponent(value);
   },
 
@@ -156,11 +181,11 @@ export default BaseStore.extend({
   },
 
   _write(value, expiration) {
-    this.get('cookies').write(this.cookieName, value, {
+    this.get('_cookies').write(this.cookieName, value, {
       domain:   this.cookieDomain,
-      expires:  new Date(expiration).toUTCString(),
-      path:     '; path=/',
-      secure:   !!this._secureCookies,
+      expires:  expiration ? new Date(expiration) : null,
+      path:     '/',
+      secure:   !!this.get('_secureCookies')
     });
   },
 
--- a/addon/session-stores/cookie.js
+++ b/addon/session-stores/cookie.js
@@ -82,7 +82,7 @@ export default BaseStore.extend({
   _fastboot: computed(function() {
     let owner = getOwner(this);
 
-    return owner.lookup('service:fastboot');
+    return owner && owner.lookup('service:fastboot');
   }),
 
   _secureCookies: computed(function() {
@@ -108,7 +108,7 @@ export default BaseStore.extend({
 
   init() {
     this._super(...arguments);
-/*
+
     if (!this.get('_fastboot.isFastBoot')) {
       next(() => {
         this._syncData().then(() => {
@@ -117,7 +117,7 @@ export default BaseStore.extend({
       });
     } else {
       this._renew();
-    }*/
+    }
   },
 
   /**
--- a/addon/session-stores/cookie.js
+++ b/addon/session-stores/cookie.js
@@ -3,7 +3,7 @@ import BaseStore from './base';
 import objectsAreEqual from '../utils/objects-are-equal';
 import getOwner from 'ember-getowner-polyfill';
 
-const { RSVP, computed, run: { next } } = Ember;
+const { RSVP, computed, inject: { service }, run: { next } } = Ember;
 
 /**
   Session store that persists data in a cookie.
@@ -77,11 +77,7 @@ export default BaseStore.extend({
   */
   cookieExpirationTime: null,
 
-  _cookies: computed(function() {
-    let owner = getOwner(this);
-
-    return owner.lookup('service:cookies');
-  }),
+  _cookies: service('cookies'),
 
   _fastboot: computed(function() {
     let owner = getOwner(this);
--- a/addon/session-stores/cookie.js
+++ b/addon/session-stores/cookie.js
@@ -108,7 +108,7 @@ export default BaseStore.extend({
 
   init() {
     this._super(...arguments);
-
+/*
     if (!this.get('_fastboot.isFastBoot')) {
       next(() => {
         this._syncData().then(() => {
@@ -117,7 +117,7 @@ export default BaseStore.extend({
       });
     } else {
       this._renew();
-    }
+    }*/
   },
 
   /**
@@ -166,8 +166,7 @@ export default BaseStore.extend({
   },
 
   _read(name) {
-    const value = this.get('_cookies').read(name) || '';
-    return decodeURIComponent(value);
+    return this.get('_cookies').read(name) || '';
   },
 
   _calculateExpirationTime() {
--- a/addon/session-stores/cookie.js
+++ b/addon/session-stores/cookie.js
@@ -148,6 +148,7 @@ export default BaseStore.extend({
     if (Ember.isEmpty(data)) {
       return RSVP.resolve({});
     } else {
+      data = decodeURIComponent(data);
       return RSVP.resolve(JSON.parse(data));
     }
   },
--- a/addon/session-stores/cookie.js
+++ b/addon/session-stores/cookie.js
@@ -60,10 +60,10 @@ export default BaseStore.extend({
 
     @property cookieName
     @type String
-    @default ember_simple_auth:session
+    @default ember_simple_auth-session
     @public
   */
-  cookieName: 'ember_simple_auth:session',
+  cookieName: 'ember_simple_auth-session',
 
   /**
     The expiration time for the cookie in seconds. A value of `null` will make
@@ -87,7 +87,7 @@ export default BaseStore.extend({
 
   _secureCookies: computed(function() {
     if (this.get('_fastboot.isFastBoot')) {
-      return this.get('_fastboot._fastbootInfo.request.hostname').indexOf('https:') === 0;
+      return this.get('_fastboot.request.host').indexOf('https:') === 0;
     } else {
       return window.location.protocol === 'https:'
     }
@@ -101,7 +101,7 @@ export default BaseStore.extend({
     if (this.get('_fastboot.isFastBoot')) {
       return false;
     } else {
-      const visibilityState = document.visibilityState || 'visible';
+      const visibilityState = typeof document !== 'undefined' ? document.visibilityState || 'visible' : false;
       return visibilityState === 'visible';
     }
   }).volatile(),
@@ -148,7 +148,6 @@ export default BaseStore.extend({
     if (Ember.isEmpty(data)) {
       return RSVP.resolve({});
     } else {
-      data = decodeURIComponent(data);
       return RSVP.resolve(JSON.parse(data));
     }
   },
--- a/addon/session-stores/cookie.js
+++ b/addon/session-stores/cookie.js
@@ -59,10 +59,10 @@ export default BaseStore.extend({
 
     @property cookieName
     @type String
-    @default ember_simple_auth-session
+    @default ember_simple_auth:session
     @public
   */
-  cookieName: 'ember_simple_auth-session',
+  cookieName: 'ember_simple_auth:session',
 
   /**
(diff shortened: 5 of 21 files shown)


addon/initializers/setup-session.js
```diff
@@ -11,5 +11,7 @@ export default function setupSession(registry) {
     store = 'session-store:test';
     registry.register(store, Ephemeral);
   }
+
+  inject(registry, store, 'cookies', 'service:cookies');
```
arjansingh: One down-side of this approach is that you end up injecting `service:cookies` into whatever `session-store` method the application is using, regardless of whether that method required cookies.

I could not find a cleaner way to inject it though. Open to suggestions.

package.json
```diff
@@ -31,6 +32,7 @@
     "ember-cli-blueprint-test-helpers": "~0.10.0",
     "ember-cli-content-security-policy": "~0.5.0",
     "ember-cli-dependency-checker": "^1.2.0",
+    "ember-cli-fastboot": "ember-fastboot/ember-cli-fastboot",
```
arjansingh: For the final PR I will use a published version of fastboot.

addon/session-stores/cookie.js
```diff
@@ -76,14 +76,16 @@ export default BaseStore.extend({
   */
   cookieExpirationTime: null,
 
-  _secureCookies: window.location.protocol === 'https:',
+  _secureCookies: computed(function () {
+    return window ? window.location.protocol === 'https:' : _fastbootInfo.host().includes('https:');
```
arjansingh: `_fastbootInfo.host()` will be going away with https://github.com/ember-fastboot/ember-fastboot-server/pull/38

I will have to update references accordingly.

marcoow: `FastBootRequest` also has a `protocol` property.

addon/initializers/setup-session.js
marcoow: Hm, why not just use

```js
cookies: service()
```

in the cookie store?

arjansingh: http://stackoverflow.com/questions/34970881/inject-a-service-into-an-ember-object-not-an-ember-controller

Specifically because the `session-stores` are Ember Objects. Ember Objects do not have lifecycle hooks, so `Ember.inject.service()` does not work.

We could refactor the `session-stores` to be services, but that expands the scope of this PR by a lot.

marcoow: Ah, I see. But then we could just add sth. like 

```js
services: computed(function() {
  let owner = getOwner(this);

  return owner.lookup('service:cookies');
}),
```

to the cookie store so you don't have to set the service externally.

arjansingh: I tried something like this first. But not identical. I'll give it another try.

arjansingh:
Status Update:

I'm writing a PR to implement `FastBootResponse` in `ember-fastboot-server`, so that we have a consistent interface for `ember-cookies`. I'll then update `ember-cookies` to work with the new `FastBootRequest` and `FastBootResponse` interfaces. Then, finally I should be able to finish up this PR.

TL;DR: The critical PR path to FasBoot compatibility is now `ember-fastboot-server => ember-cookies => ember-simple-auth`.

marcoow:
I'm getting

```
The Broccoli Plugin: [BroccoliMergeTrees] failed with:
Error: Tree cycle detected
```

when running this

marcoow:
I made some modifications here: https://github.com/simplabs/ember-simple-auth/compare/fastboot

I guess for now we can use the current FastBoot API and switch to using `FastBootRequest` and `FastBootResponse` as those become available as that shouldn't change anything significantly.

arjansingh:
Things should not change significantly so we can go ahead with that. My only suggestion then would be to keep it in beta until we get the official APIs working.

arjansingh:
@marcoow I'm working off of your new branch. A lot of it is looking good! 😄 

The big problem I'm hitting right now is getting the mock server for the dummy app to work. Fastboot server is not registering it's routes. Looking at workarounds right now.

mike-north:
I have a working fastboot app with ember-simple-auth now :) From my hacking through things, I have some more info for you guys:

* ember-cookies needs a new release cut, including https://github.com/simplabs/ember-cookies/pull/16 . Without this, ember-cookies will never successfully detect that it's in fastboot mode, and will attempt to access the "document" global (which of course doesn't exist in fastboot context!)
* `UnauthenticatedRouteMixin` and `AuthenticatedRouteMixin` need some adjustment made (or fastboot needs a fix) to handle `transition.abort()` properly. I'm not sure if "the right way" is even defined yet, but it appears that fastboot considers its job to be done (or impossible to continue) upon the transition being aborted, and simple-auth is about to continue with a new transition. I believe a `this.transitionTo` within `beforeModel` will abort the current transition anyway (see documentation on `Route#redirect`, which makes a point about saying that in the redirect hook, the transition will NOT be aborted -- implying it would be aborted in other hooks).

marcoow:
> ember-cookies needs a new release cut, including simplabs/ember-cookies#16 . Without this, ember-cookies will never successfully detect that it's in fastboot mode, and will attempt to access the "document" global (which of course doesn't exist in fastboot context!)

Done, see https://github.com/simplabs/ember-cookies/releases/tag/0.0.6

> UnauthenticatedRouteMixin and AuthenticatedRouteMixin need some adjustment made (or fastboot needs a fix) to handle transition.abort() properly. I'm not sure if "the right way" is even defined yet, but it appears that fastboot considers its job to be done (or impossible to continue) upon the transition being aborted, and simple-auth is about to continue with a new transition. I believe a this.transitionTo within beforeModel will abort the current transition anyway (see documentation on Route#redirect, which makes a point about saying that in the redirect hook, the transition will NOT be aborted -- implying it would be aborted in other hooks).

I think the way to go here might just be to respond  with a  302 redirect via the response that FastBoot exposes: http://expressjs.com/en/4x/api.html#res.redirect

mike-north:
> I think the way to go here might just be to respond with a 302 redirect via the response that FastBoot exposes: http://expressjs.com/en/4x/api.html#res.redirect

I wonder if this is something that should be handled outside of ESA? It seems like, if the desired result here is to have a redirect, the desired result whenever you have a `this.transitionTo` or `this.replaceWith` in the route that is active for the server-side rendered stuff, it should be treated the same way.

marcoow:
> I wonder if this is something that should be handled outside of ESA? It seems like, if the desired result here is to have a redirect, the desired result whenever you have a this.transitionTo or this.replaceWith in the route that is active for the server-side rendered stuff, it should be treated the same way.

Sure, that's sth. that could be handled in FastBoot. But then the question is whether FastBoot would patch `Route.transitionTo` somehow or whether we add an abstraction on top of transitionTo and http://expressjs.com/en/4x/api.html#res.redirect that works for both cases (like ember-cookies)? The latter would of course mean that using FastBoot requires to use more and more things on top of it that are not standard Ember.js. I guess for now it might be the best option to just do 2 different things in ESA depending on whether the app runs in FastBoot or the browser and see whether we find a good way of solving these things in FastBoot itself (also might make sense to wait a bit and see which problems of this kind people actually have before tackling them in FastBoot).

.gitignore
```diff
@@ -17,3 +18,17 @@
 /libpeerconnection.log
 npm-debug.log
 testem.log
+
+# Created by https://www.gitignore.io/api/vim
```
maletor: You should probably just add this to your global gitignore. I don't think it belongs here.

README.md
```diff
@@ -503,7 +503,9 @@ store if `localStorage` is available.
 
 [The Cookie store](http://ember-simple-auth.com/api/classes/CookieStore.html)
 stores its data in a cookie. This is used by the adaptive store if
-`localStorage` is not available.
+`localStorage` is not available. __This store must be used when the
+application uses
+[FastBoot](https://github.com/ember-fastboot/ember-cli-fastboot).__
```
mike-north: I wonder if we could log a warning message whenever localstorage session-store is used in conjunction with fastboot?

addon/mixins/unauthenticated-route-mixin.js
```diff
@@ -34,6 +35,12 @@ export default Ember.Mixin.create({
   */
   session: service('session'),
 
+  _isFastBoot: computed(function() {
+    const fastboot = getOwner(this).lookup('service:fastboot');
+
+    return fastboot ? fastboot.get('isFastBoot') : false;
+  }),
```
mike-north: same code as in `authenticated-route-mixin.js` -- Potentially worth creating either a base mixin, or a util containing things like this?

README.md
marcoow: That should probably be an assertion instead so that the app crashes in FastBoot as it wouldn't be going to work correctly anyway.

addon/mixins/authenticated-route-mixin.js
```diff
@@ -33,6 +34,12 @@ export default Ember.Mixin.create({
   */
   session: service('session'),
 
+  _isFastBoot: computed(function() {
+    const fastboot = getOwner(this).lookup('service:fastboot');
+
+    return fastboot ? fastboot.get('isFastBoot') : false;
+  }),
```
marcoow: this doesn't seem to be used?

addon/mixins/unauthenticated-route-mixin.js
marcoow: I think the one in `authenticated-route-mixin.js` is actually no longer needed.

marcoow: actually this one isn't used anymore neither ;)

addon/mixins/unauthenticated-route-mixin.js
```diff
@@ -50,7 +57,9 @@ export default Ember.Mixin.create({
   beforeModel() {
     if (this.get('session').get('isAuthenticated')) {
       Ember.assert('The route configured as Configuration.routeIfAlreadyAuthenticated cannot implement the UnauthenticatedRouteMixin mixin as that leads to an infinite transitioning loop!', this.get('routeName') !== Configuration.routeIfAlreadyAuthenticated);
-      return this.transitionTo(Configuration.routeIfAlreadyAuthenticated);
+
+      this.set('session.attemptedTransition', transition);
```
marcoow: we don't need to remember the transition in this case

addon/mixins/unauthenticated-route-mixin.js
```diff
@@ -50,7 +57,9 @@ export default Ember.Mixin.create({
...
+      return this.transitionTo(Configuration.authenticationRoute);
     } else {
       return this._super(...arguments);
     }
```
marcoow: actually I don't think this whole file needs to be changed at all

addon/session-stores/cookie.js
```diff
@@ -76,25 +77,51 @@ export default BaseStore.extend({
   */
   cookieExpirationTime: null,
 
-  _secureCookies: window.location.protocol === 'https:',
+  _cookies: computed(function() {
+    let owner = getOwner(this);
+
+    return owner.lookup('service:cookies');
+  }),
```
marcoow: This can be injected as a normal service as ember-cookies will be a dependency of ember-simple-auth and thus always be present.

addon/session-stores/cookie.js
```diff
@@ -76,25 +77,51 @@ export default BaseStore.extend({
...
+    return owner.lookup('service:fastboot');
+  }),
+
+  _secureCookies: computed(function() {
+    if (this.get('_fastboot.isFastBoot')) {
+      return this.get('_fastboot._fastbootInfo.request.hostname').indexOf('https:') === 0;
+    } else {
+      return window.location.protocol === 'https:'
+    }
+  }).volatile(),
```
marcoow: this should be transparently handled within ember-cookies now

addon/session-stores/cookie.js
```diff
@@ -76,25 +77,51 @@ export default BaseStore.extend({
...
+      next(() => {
+        this._syncData().then(() => {
+          this._renewExpiration();
+        });
       });
-    });
+    } else {
+      this._renew();
+    }*/
   },
```
marcoow: this seems to be still in progress?

tests/dummy/app/authenticators/oauth2.js
```diff
@@ -1,5 +1,6 @@
 import OAuth2PasswordGrant from 'ember-simple-auth/authenticators/oauth2-password-grant';
 
 export default OAuth2PasswordGrant.extend({
-  serverTokenRevocationEndpoint: '/revoke'
+  serverTokenEndpoint: 'http://localhost:4200/token',
+  serverTokenRevocationEndpoint: 'http://localhost:4200/revoke'
```
marcoow: Can we use the `apiHost` setting from the config here?

arjansingh:
thanks @marcoow just rebased this branch. working on it this week.

arjansingh:
fyi the dummy app now works 😄

addon/session-stores/cookie.js
arjansingh: Yeah it looks like you commented out those line. What are we trying to do here? Happy to make the changes.

tests/dummy/app/authenticators/oauth2.js
arjansingh: sure thing!

arjansingh:
@marcoow ok so a couple tests are failing. I'm going to try to debug them, but if you could as well, I'd appreciate it. I'm pretty sure it's just updating the tests to work with the new cookie interface.

addon/mixins/authenticated-route-mixin.js
```diff
@@ -55,8 +56,6 @@ export default Ember.Mixin.create({
   beforeModel(transition) {
     if (!this.get('session.isAuthenticated')) {
       Ember.assert('The route configured as Configuration.authenticationRoute cannot implement the AuthenticatedRouteMixin mixin as that leads to an infinite transitioning loop!', this.get('routeName') !== Configuration.authenticationRoute);
-
-      this.set('session.attemptedTransition', transition);
       return this.transitionTo(Configuration.authenticationRoute);
     } else {
       return this._super(...arguments);
```
marcoow: looks like all changes to this file can be reverted

addon/mixins/unauthenticated-route-mixin.js
```diff
@@ -1,7 +1,8 @@
 import Ember from 'ember';
+import getOwner from 'ember-getowner-polyfill';
 import Configuration from './../configuration';
 
-const { service } = Ember.inject;
+const { inject: { service }, computed } = Ember;
 
 /**
   __This mixin is used to make routes accessible only if the session is
```
marcoow: looks like all changes to this file can be reverted

addon/session-stores/cookie.js
```diff
@@ -59,10 +60,10 @@ export default BaseStore.extend({
 
     @property cookieName
     @type String
-    @default ember_simple_auth-session
+    @default ember_simple_auth:session
     @public
   */
-  cookieName: 'ember_simple_auth-session',
+  cookieName: 'ember_simple_auth:session',
```
marcoow: we shouldn't change the cookie name in the PR if not absolutely necessary

addon/session-stores/cookie.js
```diff
@@ -76,25 +77,47 @@ export default BaseStore.extend({
...
+
+  _fastboot: computed(function() {
+    let owner = getOwner(this);
+
+    return owner.lookup('service:fastboot');
+  }),
+
+  _secureCookies: computed(function() {
+    if (this.get('_fastboot.isFastBoot')) {
+      return this.get('_fastboot._fastbootInfo.request.hostname').indexOf('https:') === 0;
```
marcoow: Doesn't FastBoot provide a public interface to the request and response now?

addon/session-stores/cookie.js
marcoow: Oh, that was me ;)

I guess I did that probably only because cookie access wasn't working in FastBoot back then. Should be good to uncomment that now.

addon/session-stores/cookie.js
```diff
@@ -125,6 +148,7 @@ export default BaseStore.extend({
     if (Ember.isEmpty(data)) {
       return RSVP.resolve({});
     } else {
+      data = decodeURIComponent(data);
```
marcoow: I think this should not be necessary anymore

addon/session-stores/cookie.js
```diff
@@ -143,26 +167,22 @@ export default BaseStore.extend({
...
 
   _read(name) {
-    let value = document.cookie.match(new RegExp(`${name}=([^;]+)`)) || [];
-    return decodeURIComponent(value[1] || '');
+    return this.get('_cookies').read(name) || '';
   },
 
   _calculateExpirationTime() {
-    let cachedExpirationTime = this._read(`${this.cookieName}-expiration_time`);
+    let cachedExpirationTime = this._read(`${this.cookieName}:expiration_time`);
```
marcoow: same as above - we shouldn't change the cookie name here unless absolutely necessary

marcoow:
> @marcoow ok so a couple tests are failing. I'm going to try to debug them, but if you could as well, I'd appreciate it. I'm pretty sure it's just updating the tests to work with the new cookie interface.

Great!

One of the tests is failing due to the changed default cookie name which should be reverted anyway if possible.

addon/mixins/authenticated-route-mixin.js
arjansingh: sorry this was a bad merge resolve.

arjansingh:
I tired all day to get these tests passing. I can't do it. I think it's because the tests should be using [ember-mocha](https://github.com/switchfly/ember-mocha) instead of mocha. Can you give this a shot @marcoow?

marcoow:
> I tired all day to get these tests passing. I can't do it. I think it's because the tests should be using ember-mocha instead of mocha. Can you give this a shot @marcoow?

everything seems to be fine already?

arjansingh:
Oh that's so weird. They were failing locally for me and they were failing before I pushed. 😛 

I'm going to integrate this into my app to see how it works beyond the dummy app. Can you think of anything else the addon needs?

addon/mixins/authenticated-route-mixin.js
```diff
@@ -1,7 +1,8 @@
 import Ember from 'ember';
+import getOwner from 'ember-getowner-polyfill';
 import Configuration from './../configuration';
 
-const { service } = Ember.inject;
+const { inject: { service }, computed } = Ember;
 
 /**
   __This mixin is used to make routes accessible only if the session is
```
marcoow: I think these changes can be reverted?

addon/mixins/unauthenticated-route-mixin.js
```diff
@@ -1,4 +1,5 @@
 import Ember from 'ember';
+import getOwner from 'ember-getowner-polyfill';
```
marcoow: I think this is not needed?

tests/unit/session-stores/shared/cookie-store-behavior.js
```diff
@@ -3,54 +3,69 @@ import Ember from 'ember';
...
       store.persist({ key: 'value' });
-      renew(store);
+      renew(store).then(done);
     });
 
-    it('stores the expiration time in a cookie named "test-session-expiration_time"', () => {
-      expect(document.cookie).to.contain(`${store.cookieName}-expiration_time=60`);
-    });
+    // TODO: the "…:expiration_time" never actually seems to get written
+    it('stores the expiration time in a cookie named "test:session:expiration_time"');
```
marcoow: This should be enabled again

arjansingh:
The tests won't pass without those changes. See the comments in the code.

On Monday, June 20, 2016, Marco Otte-Witte <notifications@github.com> wrote:

> There's a few open issues still - I hope to find time this week to help
> finishing it.
>
> —
> You are receiving this because you authored the thread.
> Reply to this email directly, view it on GitHub
> <https://github.com/simplabs/ember-simple-auth/pull/964#issuecomment-227070132>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/ABhksVVbiPBGoOSjlOR-1W4gYgcFyHC4ks5qNkMHgaJpZM4II0ii>
> .
>


-- 
Arjan

josemarluedke:
Is there anything I can help with to get this merged soon?

marcoow:
@josemarluedke: best way to get involved is to join our call at 6:30pm CEST today where we discuss the status of things, assign tasks etc. Check the #dev-simple-auth channel on the community slack for details.

josemarluedke:
Sent out a PR to @arjansingh with some basic stuff, such as rabase, upgrade fastboot and more. I will be working more on this tomorrow.

arjansingh:
@josemarluedke @marcoow Thanks for taking a look and rebasing with https://github.com/arjansingh/ember-simple-auth/pull/3

I have a proposal. Can we create the feature branch on `simplabs/ember-simple-auth#fastboot`?

Then we can merge my code into that and use it for mainline development. I don't want to become a bottleneck for this work.

marcoow:
@arjansingh, @josemarluedke: makes sense. Can you guys create a new PR against the [`fastboot` branch](https://github.com/simplabs/ember-simple-auth/tree/fastboot)?

Thanks for your efforts!

marcoow:
replaced with #1035

tests/unit/session-stores/shared/cookie-store-behavior.js
josemarluedke: I tried to make this pass, but no luck on that.

addon/session-stores/cookie.js
josemarluedke: It seems that this has been changed in the new feature branch.

addon/mixins/unauthenticated-route-mixin.js
josemarluedke: This is been used in the new feature branch.

addon/initializers/setup-session.js
josemarluedke: Fixed in https://github.com/simplabs/ember-simple-auth/pull/1039

tests/unit/session-stores/shared/cookie-store-behavior.js
marcoow: I'll try to fix that

addon/session-stores/cookie.js
marcoow: yeah, I reverted these back to the old names

PR #968, opened 2016-04-20
Reference well written blog post 

ssured:
https://github.com/ryanlabouve/ember-jwt-esa-blog
@ryanlabouve : ping

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,6 +65,8 @@ npm install && bower install && ember serve
 
 and go to [http://localhost:4200](http://localhost:4200).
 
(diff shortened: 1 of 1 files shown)


marcoow:
While I think @ryanlabouve's posts are indeed very good I don't think blog posts in general are sth. that belongs in the README. They tend to get out of date at some point and usually (and also in this case) cover a very specific aspect or use case of the library so are only relevant for some of the people reading the README.

@ryanlabouve suggested having a cookbooks collection or so somewhere and I think that would be a much better place for referring to blog posts like his.

ryanlabouve:
@ssured thanks for the recommendation! I agree with @marcoow 100% on not adding it for the reasons he mentioned. 

However, I'll update you on any progress that's made towards a cookbook!

Issue #904, opened 2016-02-19
Add the ability to make all routes use AuthenticatedRouteMixin by default (except the auth route)


mdentremont:
It would be nice if there was an option to automatically make _all_ routes require authentication (except for the authentication route).

This would prevent issues where a dev forgets to add the Authentication mixin. One could then use the UnauthenticationRouteMixin to explicitly not require authentication on certain routes.

marcoow:
Would you have sth. in mind how that might look? Maybe you can come up with a PR?

mdentremont:
I'll preface by saying the following is a nasty hack, but it demonstrates one way of implementing the desired functionality:

```python
import Ember from 'ember';

import Configuration from 'ember-simple-auth/configuration';
import UnauthenticatedRouteMixin from 'ember-simple-auth/mixins/unauthenticated-route-mixin';

export default Ember.Route.reopen({
  session: Ember.inject.service('session'),

  beforeModel(transition) {
    if (Configuration.requireAuthenticateByDefault) {
      if (this.routeName != 'application' && (!UnauthenticatedRouteMixin.detect(this)) &&
          !this.get('session.isAuthenticated')) {
        Ember.assert(
            'The route configured as Configuration.authenticationRoute cannot implement the AuthenticatedRouteMixin mixin as that leads to an infinite transitioning loop!',
            this.get('routeName') !== Configuration.authenticationRoute);

        transition.abort();
        this.set('session.attemptedTransition', transition);
        this.transitionTo(Configuration.authenticationRoute);
        return;
      }
    }
    return this._super(...arguments);
  }
};
```

mdentremont:
We might be able to add the following code to the `ApplicationRouteMixin` in order to add the desired functionality:

```
  beforeModel(transition) {
    if (!config['simple-auth'].authenticatedRoutes || this.get('session.isAuthenticated')) {
      return this._super(...arguments);
    }

    if (transition.targetName === Configuration.authenticationRoute) {
      Ember.assert(
          'The route configured as Configuration.authenticationRoute must not implement the AuthenticatedRouteMixin mixin, as that leads to an infinite transitioning loop!',
          !AuthenticatedRouteMixin.detect(this));

      return this._super(...arguments);
    }

    for (let handlerInfo of transition.handlerInfos) {
      if (handlerInfo.name === 'application') {
        // Skip the application route
        continue;
      }

      if (!UnauthenticatedRouteMixin.detect(handlerInfo.handler)) {
[...]

PR #971, opened 2016-04-25
WIP Extend Router DSL  #904

ryanlabouve:
Opening a WIP PR since this issue had quite a bit of activity today to give some transparency with where things are at.

I'm definitely welcome to suggestions! Here's my current list of questions:
- How should we handle `authenticatedRoute` name conflict?
- How should we bring over the container utilities?
- Any suggestions on tests?

Will make some more progress on this tonight. After we figure out how to handle the name conflict, expanding to handle authenticatedRoute will be easy.
#### There is a name conflict with `authenticatedRoute` via torii

Since torii is currently using `this.unauthenticatedRoute` on Router, we have no way to distinguish whether they want torii or ESA mixins.

Should we somehow namespace our new routes to Ember Simple Auth? (That's why I started with `unauthenticatedRoute`.)
#### Should we combine container utilities into one file?

Several container utilities are needed for registering the mixin on a given route. `addon/utils/lookup.js` is an example of one that ESA already had. Grabbed `register` and `lookup-factory` from troii. Should we consolidate this into one file in ESA, pull out into independent addon, or leave in different files?

`addon/utils/lookup.js`
`addon/utils/register.js` from https://github.com/Vestorly/torii/blob/master/addon/lib/container-utils.js

`addon/utils/lookup-factory.js` from https://github.com/Vestorly/torii/blob/master/addon/lib/container-utils.js
#### Still working on the tests.

Suggestions welcome. I still want to verify that `addon/bootstrap/routing.js` is properly applying a mixin via test. Will probalby break this out into its own unit test.

May rethink some of the tests based off of reading https://github.com/Vestorly/torii/blob/master/tests/acceptance/routing-test.js. (Probably should have done that much earlier)

/cc @mdentremont

### DIFF
--- a/addon/bootstrap/routing.js
+++ b/addon/bootstrap/routing.js
@@ -1,0 +1,21 @@
+import lookup from '../utils/lookup';
+import { lookupFactory } from '../utils/lookup-factory';
+import { register } from '../utils/register';
+
+function reopenOrRegister(applicationInstance, factoryName, mixin) {
+  const factory = lookup(applicationInstance, factoryName);
+  if (factory) {
+    factory.reopen(mixin);
+  } else {
+    const basicFactory = lookupFactory(applicationInstance, 'route:basic');
+    const routeWithMixin = basicFactory.extend(mixin);
+    register(applicationInstance, factoryName, routeWithMixin);
+  }
+}
+
+export default function(applicationInstance, unauthenticatedRoutes, mixin) {
+  unauthenticatedRoutes.forEach((routeName) => {
+    const factoryName = `route:${routeName}`;
+    reopenOrRegister(applicationInstance, factoryName, mixin);
+  });
+}
--- a/addon/instance-initializers/setup-router.js
+++ b/addon/instance-initializers/setup-router.js
@@ -1,0 +1,21 @@
+import '../router-dsl-ext';
+import bootstrapRouting from '../bootstrap/routing';
+import UnauthenticatedRouteMixin from 'ember-simple-auth/mixins/unauthenticated-route-mixin';
+
+export function initialize(applicationInstance) {
+  const router = applicationInstance.get('router');
+  const setupRoutes = function() {
+    const unauthenticatedRoutes = router.router.unauthenticatedRoutes;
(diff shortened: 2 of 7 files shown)


ryanlabouve:
(This relates to #904)

marcoow:
The conflicting router DSL method names are a really a bit unfortunate. Given that the mixin's names are `AuthenticatedRouteMixin` and `UnauthenticatedRouteMixin` it would be great to be able to use `authenticatedRoute` and `unauthenticatedRoute` accordingly of course.

It might be a good idea to make the router DSL opt-in (and ask the torii project to do so as well) as changes of Ember.js core APIs will always have that potential for conflicts and thus should maybe not switched on by default anyway (/cc @mixonic).

An alternative API we could use that would be conflict-free would be sth. like this:

```js
Router.map(function() {
  this.route('index');
  this.authenticated(function() {
    this.route('projects');
  });
  this.unauthenticated(function() {
    this.route('login');
  });
})
```

marcoow:
@ryanlabouve: an alternative might be to use `this.route('projects', { authenticated: true });`/`this.route('projects', { unauthenticated: true });`

mtangoo:
Interested in this. Commented so that I get notification on the proceeding of this PR

marcoow:
OT: @mtangoo: you can always click the subscribe button on the right on anything.

mtangoo:
Thanks,  @marcow. Didn't know that

marcoow:
Moving this to 1.3 to make it easier for us to finally get 1.2 out.

marcoow:
I'm not even so sure anymore we need this at all as the only case where this is really useful is if you have a totally flat routing structure with a lot of routes that should all be authenticated but are not below a common parent route that the `AuthenticatedRouteMixin` could be applied to which seems to be a pretty rare case.

At the same time making this so that it works in all supported Ember versions (which actually go back all the way to 1.2) would be a huge effort, not to speak about future Ember versions…

ryanlabouve:
👍 closing.

marcoow:
Didn't want to imply it has to be closed - but if you think the above points make sense that's cool.

PR #978, opened 2016-05-13
change cookie default key names to be rfc2616 compliant

### LINKED ISSUE #977: Default keys for cookies are not RFC standard compliant
The default key names of this plugin are `ember_simple_auth:session` and `ember_simple_auth:session:expiration_time` those are not standard compliant b/c of the colon in them.
Have a read of this gist that gives a good summary: https://gist.github.com/jeremiahlee/ff7dfb60572211dce306

This causes problems as soon as you use cookies and your backend is itself standard compliant and validates cookies or also if the backend validates the headers, one example would be `hapi.js` you will be denied to access those applications with an error like 

    {"statusCode":400,"error":"Bad Request","message":"Invalid cookie value"}

And the behavior of `hapi.js` is not wrong, they do deny the request right, but ember-simple-auth's default key names are wrong, so I would suggest to change those.

wzrdtales:
fixes #977

### DIFF
--- a/addon/session-stores/adaptive.js
+++ b/addon/session-stores/adaptive.js
@@ -28,10 +28,10 @@ export default Base.extend({
 
     @property localStorageKey
     @type String
-    @default 'ember_simple_auth:session'
+    @default 'ember_simple_auth-session'
     @public
   */
-  localStorageKey: 'ember_simple_auth:session',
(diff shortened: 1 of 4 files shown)


wzrdtales:
@marcoow Please review if I have missed something.

PR #981, opened 2016-05-16
Added authenticator for authentization based on session cookies

marxsk:
In order to support authorization based on session cookies, the simple authenticator is required. 

The proposed solution should be general enough to support most of the common scenarios. I would like you to consider merging this authenticator into upstream. I would happy to improve my code when it is required.

### DIFF
--- a/addon/authenticators/cookie-authenticator.js
+++ b/addon/authenticators/cookie-authenticator.js
@@ -1,0 +1,134 @@
+import Ember from 'ember';
+import Base from 'ember-simple-auth/authenticators/base';
+
+const { RSVP, Logger } = Ember;
+
+/**
+  Authenticator that works with session cookies
+
+  Authenticator is working on premise that cookie is received after authorization
+  and cookie is attached to every request automatically by browser, so no other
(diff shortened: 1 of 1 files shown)


marcoow:
Thanks for the contribution!

This seems very generic though and would probably not work for most people out of the box. We'd probably have to add a lot of customization/configuration options to this at which point it would probably get quite complex for actually very little value.

I don't think something like this should be added to the core library itself.

marxsk:
Hi, 

I agree that this is very very generic but fortunately for session cookies authentication, there is not much to do. I don't know how often is this authentication used, in general, we are using this code with Sinatra/Rack::Session::Cookies. We have tested it also with others backends and did not found more customizing required.

You have to customize:
* endpoints (/login, /logout works mostly out of the box; /login-status is strange but it can be replaced by any page available only for authorized users)
* structure of login information is generalized 
* it might happen that you have to send /login via GET but we did not meet this yet

I completely agree that building a big customization engine around this does not make sense. As most of the code is just documentation.

PR #983, opened 2016-05-18
[WIP] Persist to cookie when relevant attributes change

### LINKED ISSUE #854: Session cookie expiration issues
Hi,

I have a problem with setting my cookie expiration. I'm attempting to set up a 'remember me' functionality and going from the documentation I implemented this:

```
const expirationTime = this.get('rememberMe') ? (14 * 24 * 60 * 60) : null;
this.set('session.store.cookieExpirationTime', expirationTime);
```

The problem is if I set 'remember me' with an expiration of 14 days later and log out/log in again without 'remember me', passing in null seems to leave the cookie at the original date of 14 days later instead of switching it to a session cookie. From reading through issue #395 it sounds like that is because the default when the window is reopened will be that original 14 days later. 

I personally got passed this by just passing 'Session' into the expiration time (even though the documentation expresses that it expects an `int` or `null`) like so:

```
const expirationTime = this.get('rememberMe') ? (14 * 24 * 60 * 60) : 'Session';
this.set('session.store.cookieExpirationTime', expirationTime);
```

Whether this is intended or not, I don't know, but it is working perfectly so far. Is this a bug or is the documentation lacking? Or am I potentially doing something wrong?

stevenwu:
There's quite a bit I'm unsure about, but it appears to work.

Fix #854 .

### DIFF
--- a/addon/session-stores/adaptive.js
+++ b/addon/session-stores/adaptive.js
@@ -5,8 +5,7 @@ import LocalStorage from 'ember-simple-auth/session-stores/local-storage';
 import Cookie from 'ember-simple-auth/session-stores/cookie';
 
 const {
-  computed,
-  observer
+  computed
 } = Ember;
 
 const LOCAL_STORAGE_TEST_KEY = '_ember_simple_auth_test_key';
@@ -54,7 +53,10 @@ export default Base.extend({
     },
     set(key, value) {
       this.set('_cookieDomain', value);
-      this.get('_store').set('cookieDomain', value);
+      let _store = this.get('_store');
+      if (_store) {
+        _store.set('cookieDomain', value);
+      }
       return value;
     }
   }),
@@ -68,13 +70,16 @@ export default Base.extend({
     @public
   */
   _cookieName: 'ember_simple_auth-session',
-  cookieName: computed({
+  cookieName: computed('cookieName', {
     get(key) {
       return this.get('_cookieName');
     },
     set(key, value) {
       this.set('_cookieName', value);
-      this.get('_store').set('cookieName', value);
+      let _store = this.get('_store');
+      if (_store) {
+        _store.set('cookieName', value);
+      }
       return value;
     }
   }),
@@ -96,7 +101,10 @@ export default Base.extend({
     },
     set(key, value) {
       this.set('_cookieExpirationTime', value);
-      this.get('_store').set('cookieExpirationTime', value);
+      let _store = this.get('_store');
+      if (_store) {
+        _store.set('cookieExpirationTime', value);
+      }
       return value;
     }
   }),
--- a/addon/session-stores/adaptive.js
+++ b/addon/session-stores/adaptive.js
@@ -4,7 +4,10 @@ import Base from 'ember-simple-auth/session-stores/base';
 import LocalStorage from 'ember-simple-auth/session-stores/local-storage';
 import Cookie from 'ember-simple-auth/session-stores/cookie';
 
-const { computed } = Ember;
+const {
+  computed,
+  observer
+} = Ember;
 
 const LOCAL_STORAGE_TEST_KEY = '_ember_simple_auth_test_key';
 
@@ -137,5 +140,10 @@ export default Base.extend({
   */
   clear() {
     return this.get('_store').clear();
-  }
+  },
+
+  _cookieDataChanged: observer('cookieDomain', 'cookieName', 'cookieExpirationTime', function() {
+    const data = this.getProperties('cookieDomain', 'cookieName', 'cookieExpirationTime');
+    return this.get('_store').persist(data);
+  })
 });
--- a/addon/session-stores/adaptive.js
+++ b/addon/session-stores/adaptive.js
@@ -47,7 +47,7 @@ export default Base.extend({
     @default null
     @public
   */
-  cookieDomain: null,
+  cookieDomain: computed.alias('_store.cookieDomain'),
 
   /**
     The name of the cookie to use if `localStorage` is not available.
@@ -57,7 +57,9 @@ export default Base.extend({
     @default ember_simple_auth-session
     @public
   */
-  cookieName: 'ember_simple_auth-session',
+  cookieName: computed('_store.cookieName', function() {
+    return this.get('_store.cookieName') || 'ember_simple_auth-session';
+  }),
 
   /**
     The expiration time for the cookie in seconds if `localStorage` is not
@@ -69,7 +71,7 @@ export default Base.extend({
     @type Integer
     @public
   */
-  cookieExpirationTime: null,
+  cookieExpirationTime: computed.alias('_store.cookieExpirationTime'),
 
   _isLocalStorageAvailable: computed(function() {
     try {
@@ -140,10 +142,5 @@ export default Base.extend({
   */
   clear() {
     return this.get('_store').clear();
-  },
-
-  _cookieDataChanged: observer('cookieDomain', 'cookieName', 'cookieExpirationTime', function() {
-    const data = this.getProperties('cookieDomain', 'cookieName', 'cookieExpirationTime');
-    return this.get('_store').persist(data);
-  })
+  }
 });
--- a/addon/session-stores/adaptive.js
+++ b/addon/session-stores/adaptive.js
@@ -47,7 +47,17 @@ export default Base.extend({
     @default null
     @public
   */
-  cookieDomain: null,
+  _cookieDomain: null,
+  cookieDomain: computed({
+    get(key) {
+      return this.get('_cookieDomain');
+    },
+    set(key, value) {
+      this.set('_cookieDomain', value);
+      this.get('_store').set('cookieDomain', value);
+      return value;
+    }
+  }),
 
   /**
     The name of the cookie to use if `localStorage` is not available.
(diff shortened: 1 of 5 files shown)


addon/session-stores/adaptive.js
```diff
@@ -137,5 +140,10 @@ export default Base.extend({
   */
   clear() {
     return this.get('_store').clear();
-  }
+  },
+
+  _cookieDataChanged: observer('cookieDomain', 'cookieName', 'cookieExpirationTime', function() {
+    const data = this.getProperties('cookieDomain', 'cookieName', 'cookieExpirationTime');
+    return this.get('_store').persist(data);
+  })
```
marcoow: Could we do that with getters and setters instead of an observer? E.g.

```js
cookieDomain:  computed({
  get() {
    return this._cookieDomain;
  },
  set(_, value) {
    this._cookieDomain = value;
    this.rewriteCookie();
  }
}),
```

Also it should be sufficient to only define that on the cookie store. We just need to make sure that the cookie settings are defined as aliases on the `_store` then.

stevenwu:
@marcoow I'm unable to get the adaptive store tests to pass.

When should `rewriteCookie` be called? I think putting it in the setters for `cookieDomain`, `cookieName`, `cookieExpiration` is buggy because getters can be called before setters complete (I believe setters must return a value).

addon/session-stores/adaptive.js
```diff
@@ -66,7 +69,7 @@ export default Base.extend({
     @type Integer
     @public
   */
-  cookieExpirationTime: null,
+  cookieExpirationTime: computed.alias('_store.cookieExpirationTime'),
```
marcoow: I think these properties have to be actual properties on the adaptive store itself as they are read from that and assigned to the cooke store when that's created (see [line 92](https://github.com/simplabs/ember-simple-auth/pull/983/files#diff-8ccd5433965bb44e3dbb5f054de89699R92))

addon/session-stores/cookie.js
```diff
@@ -107,6 +143,21 @@ export default BaseStore.extend({
...
+    if (data.hasOwnProperty('cookieName')) {
+      this.set('cookieName', data.cookieName);
+      delete data.cookieName;
+    }
+
+    if (data.hasOwnProperty('cookieExpirationTime')) {
+      this.set('cookieExpirationTime', data.cookieExpirationTime);
+      delete data.cookieExpirationTime;
+    }
+
```
marcoow: When would `data` contain these options (`cookieDomain` etc.)? The idea would be that `data` only contains the actual payload to write to the cookie and the options for `cookieDomain` etc. are just read from the instance itself (`this.get('cookieDomain')` etc.).

addon/session-stores/cookie.js
```diff
@@ -198,5 +257,14 @@ export default BaseStore.extend({
     } else {
       return RSVP.resolve();
     }
+  },
+
+  // FIXME: maybe be causing getter bugs with rewriteCookie called before
+  // setter can return value
```
marcoow: Can you explain what this is about? I'm not sure I understand what you mean.

stevenwu: According to the guides, the setter should return the value you want to assign
```
fullName: Ember.computed('firstName', 'lastName', {
    set(key, value) {
      var [firstName, lastName] = value.split(/\s+/);
      this.set('firstName', firstName);
      this.set('lastName',  lastName);
      this.rewriteCookie();
      return value;
    }
  })
```
An example buggy situation is where there's `this.get('fullName')` inside `rewriteCookie()`. It's a non issue if this code is wrong.

stevenwu:
@marccow I think I'm closer now. I'm wondering if the tests are thorough enough.

addon/session-stores/cookie.js
marcoow: I see what you mean - good point. We should maybe wrap `rewriteCookie` in `Ember.run.scheduleOnce` anyway so we don't rewrite unnecessarily often which should also solve this problem.

marcoow:
Can you rebase with the latest master?

stevenwu:
I'm unsure about `Ember.run.scheduleOnce` and the extra `run`s I added for 
```
Uncaught Error: Assertion Failed: You have turned on testing mode, which disabled the 
run-loop's autorun. You will need to wrap any code with asynchronous side-effects in an 
Ember.run
```

addon/session-stores/cookie.js
```diff
@@ -242,8 +242,10 @@ export default BaseStore.extend({
...
 
   rewriteCookie() {
-    const data = this._read(this._oldCookieName);
-    const expiration = this._calculateExpirationTime();
-    this._write(data, expiration);
+    run.scheduleOnce('actions', this, function() {
+      const data = this._read(this._oldCookieName);
+      const expiration = this._calculateExpirationTime();
+      this._write(data, expiration);
+    });
```
marcoow: The above `run.scheduleOnce('actions', this, this.rewriteCookie);` should be sufficient. I think it's easier if we keep the actual method a synchronous operation.

marcoow:
> I'm unsure about Ember.run.scheduleOnce and the extra runs I added for

>> Uncaught Error: Assertion Failed: You have turned on testing mode, which disabled the 
run-loop's autorun. You will need to wrap any code with asynchronous side-effects in an 
Ember.run

Yeah, that's fine. That message you're getting is one of the major weirdnesses in Ember testing currently - both hard to understand and hard to explain. Wrapping the code that causes it into a `run` is generally the right thing to do though.

marcoow:
Sorry for my inactvitiy - was super busy recently. Let's discuss this in this week's call.

addon/session-stores/adaptive.js
```diff
@@ -44,7 +46,21 @@ export default Base.extend({
     @default null
     @public
   */
-  cookieDomain: null,
+  _cookieDomain: null,
+  cookieDomain: computed({
+    get() {
+      return this.get('_cookieDomain');
+    },
+    /*jshint unused: true */
```
marcoow: Does this actually report an unused error for `key`? I don't think it should as we cannot prevent having it in the arguments list here. You could maybe replace it with `_` to make clear it's unused though.

addon/session-stores/adaptive.js
```diff
@@ -66,7 +96,21 @@ export default Base.extend({
...
+      let _store = this.get('_store');
+      if (_store) {
+        _store.set('cookieExpirationTime', value);
+      }
+      return value;
+    }
+  }),
 
   _isLocalStorageAvailable: computed(function() {
     try {
```
marcoow: couldn't we just define these as `alias`es for the properties on the `_store`?

addon/session-stores/cookie.js
```diff
@@ -52,7 +52,18 @@ export default BaseStore.extend({
     @default null
     @public
   */
-  cookieDomain: null,
+  _cookieDomain: null,
+  cookieDomain: computed('_cookieDomain', {
+    get() {
+      return this.get('_cookieDomain');
+    },
+    /* jshint unused: true */
```
marcoow: don't think this jshint note should be necessary

addon/session-stores/cookie.js
```diff
@@ -137,7 +171,10 @@ export default BaseStore.extend({
     @public
   */
   clear() {
-    this._write(null, 0);
+    document.cookie.split(';').forEach((c) => {
+      document.cookie = c.replace(/^ +/, '')
+        .replace(/=.*/, `=;expires=${new Date().toUTCString()};path=/`);
+    });
```
marcoow: Is this necessary for the change in this PR? It seems unrelated?

addon/session-stores/cookie.js
```diff
@@ -148,20 +185,24 @@ export default BaseStore.extend({
...
     let path        = '; path=/';
-    let domain      = isEmpty(this.cookieDomain) ? '' : `; domain=${this.cookieDomain}`;
     let expires     = isEmpty(expiration) ? '' : `; expires=${new Date(expiration).toUTCString()}`;
     let secure      = !!this._secureCookies ? ';secure' : '';
-    document.cookie = `${this.cookieName}=${encodeURIComponent(value)}${domain}${path}${expires}${secure}`;
-    if (expiration !== null) {
-      let cachedExpirationTime = this._read(`${this.cookieName}-expiration_time`);
-      document.cookie = `${this.cookieName}-expiration_time=${encodeURIComponent(this.cookieExpirationTime || cachedExpirationTime)}${domain}${path}${expires}${secure}`;
+    let cookieName = this.get('_cookieName');
+    let cookieDomain = this.get('_cookieDomain');
```
marcoow: I think you can just use

```js
let cookieName = this._cookieName;
let cookieDomain = this._cookieDomain;
```

here

tests/unit/session-stores/adaptive-test.js
```diff
@@ -45,5 +50,19 @@ describe('AdaptiveStore', () => {
         store.get('_store')._syncData();
       }
     });
+
+    it('persists to cookie when cookie attributes change', () => {
+      let store;
```
marcoow: this could be moved to line 57, e.g. `let store = Adaptive.create({…`

marcoow:
Looks good - sorry for my slow turnaround time again. I left a few comments regarding smaller things. We should add some more tests for the cookie rewrite behavior though - especially the fact that it deletes the old cookie and writes one with the new name when the name property changes and the fact that it will only rewrite the cookie once per run loop when multiple properties are changed. These should be added to the https://github.com/simplabs/ember-simple-auth/blob/master/tests/unit/session-stores/shared/cookie-store-behavior.js so that they are executed both for the cookie store as well as for the adaptive store when that uses the cookie store internally. Let me know if you need help with that!

stevenwu:
Great, will make these fixes soon.

marcoow:
@stevenwu: I fixed the cookie store rewrite tests and pushed the changes here: https://github.com/simplabs/ember-simple-auth/tree/cookie-store-rewrite

marcoow:
The cookie store rewriting efforts are now handled in #1056

PR #985, opened 2016-05-22
move route mixin transition routes into computed properties

### LINKED ISSUE #975: Allow override of config variables authenticationRoute, routeAfterAuthentication, routeIfAlreadyAuthenticated in mixin.
In the 3 `simple-auth` route mixins `AuthenticatedRouteMixin`, `UnauthenticatedRouteMixin` and `ApplicationRouteMixin`, the redirect routes defined in `config/environment.js` are used to redirect the user to the correct routes based on if they are logged in, not logged in and should be etc.
I am finding that for an app with multiple user roles (ie. a `User` login and an `Admin` login) I am wanting to override these definitions.
eg. if someone who is not logged in went to `user/dashboard`, I think it should be redirected to `user/login`, which can be defined in the `config` file, BUT if that same person hit `admin/dashboard`, I would like that to go to `admin/login`, however it will also redirect to `user/login` which could be confusing.

My thought is that the 3 mixins defined above had an overridable variable that defaulted to the config definition, eg `AuthenticatedRouteMixin` might define: (see related code [here](https://github.com/simplabs/ember-simple-auth/blob/9d5a37b1ece0b82ee504e6cd6f367458b1d3120d/addon/mixins/authenticated-route-mixin.js#L55-L65))
```javascript
authenticationRoute: Ember.computed(function() {
  return Configuration.authenticationRoute;
}),
```
which I could then redefine in my admin routes to `admin/login` (I'd probably extend the mixin and include that override so I can easily add to all admin routes).

Would love to hear if people agree with this approach, or if others have any other alternatives, I'll have a look into a PR over the next little while.

timjcook:
- [x] Ember CP's can be overrided in the route implementation of the mixin.
- [x] Review tests
- [x] Review documentation changes if required

closes #975


addon/mixins/authenticated-route-mixin.js
```diff
@@ -54,11 +66,11 @@ export default Ember.Mixin.create({
   */
   beforeModel(transition) {
     if (!this.get('session.isAuthenticated')) {
-      Ember.assert('The route configured as Configuration.authenticationRoute cannot implement the AuthenticatedRouteMixin mixin as that leads to an infinite transitioning loop!', this.get('routeName') !== Configuration.authenticationRoute);
+      Ember.assert('The route configured as Configuration.authenticationRoute cannot implement the AuthenticatedRouteMixin mixin as that leads to an infinite transitioning loop!', this.get('routeName') !== this.get('authenticationRoute'));
 
       transition.abort();
       this.set('session.attemptedTransition', transition);
-      this.transitionTo(Configuration.authenticationRoute);
+      this.transitionTo(this.get('authenticationRoute'));
```
marcoow: extract `authenticationRoute` into a local variable instead of getting it twice

marcoow:
We should also deprecate the configuration values. The best way to do that is probably when loading the config - if custom settings are provided, trigger a deprecation.

The mixin tests should also have tests that assert that the properties can be overridden.

timjcook:
@marcoow ok great, I can fix that up this week. Just to confirm, you are saying that we should show deprecations if they set some routes in the config but then override them in the route, but if they just use the config, then no deprecations?
I'll also add some tests for the mixin as per your comments.

marcoow:
@timjcook: no, I'm suggesting there should be deprecations when someone changes the route names in the config as the new way of changing those would be by overriding the respective properties in routes.

marcoow:
@timjcook: any updates on this?

timjcook:
oops, sorry about that @marcoow, slipped off my radar. Will try and get it ship shape as soon as I can :)

marcoow:
@timjcook: awesome!

timjcook:
@marcoow done. let me know if there is anything more you'd like me to change.

marcoow:
Awesome - just 2 minor things:

- [ ] Change `Configuration` example in README to not use the deprecated setting
- [ ] Add description text to `@deprecated` flag on deprecated config settings

timjcook:
@marcoow ok cool, updated the Configuration example to set the `baseURL` config option instead and added deprecation messages for each of the deprecated Configuration options.

dknutsen:
@timjcook @marcoow does this change mean that if we want to override the CPs with a custom, say 'authenticationRoute', then we have to do it in every route that extends the 'AuthenticatedRouteMixin'? I like this change overall but if that's the case it means a lot of duplication of the exact same value.

marcoow:
@dknutsen: yes, that's correct actually - in order to avoid duplication you could just create a custom base route though that all authenticated routes extend though.

Floby:
Hello
I arrived here googling the deprecation warning that I have, and I must say I'm not entirely sure what it is I have to do.
Can someone help?

dknutsen:
@Floby essentially what you used to be able to configure in `config/environment.js` now has to be configured on the routes that use the AuthenticatedRouteMixin. For example if you're specifying 'login' as your 'authenticationRoute' you used to be able to do that in `config/environment.js' but now it needs to go on each route that extends the `AuthenticatedRouteMixin`. When I asked my question above I found it weird that I'd have to specify that in a ton of different places, but I guess it makes sense, and I essentially used a twist on @marcoow's suggestion. Instead of making a base route, I extended the AuthenticatedRouteMixin to define those properties, and then in my routes, instead of using the AuthenticatedRouteMixin from ESA, I used the one that I had extended.

So I created a mixin in my app like this:
```
import Ember from 'ember';
import AuthenticatedRouteMixin from 'ember-simple-auth/mixins/authenticated-route-mixin';

export default Ember.Mixin.create(AuthenticatedRouteMixin, {
  // NOTE: this used to be something we could set in config/environment.js but
  //       that usage was removed in ESA recently and now it must be set
  //       individually on all routes that extend AuthenticatedRouteMixin, so
  //       instead of duplicating that prop in every single one of our routes
  //       I am extending the ESA mixin with our own (of the same name) and
  //       adding the authenticationRoute property here.
  //
  //       (see https://github.com/simplabs/ember-simple-auth/pull/985)


  // these can either be static strings or computed properties (they are CPs
  // in the base ESA mixin which we're extending)

  authenticationRoute: 'callback',

  //routeAfterAuthentication: '',

  //routeIfAlreadyAuthenticated: '',
});
```

Then in my routes I do this:
```
...
import AuthenticatedRouteMixin from "../mixins/authenticated-route-mixin";

export default Ember.Route.extend(AuthenticatedRouteMixin, {
...
```

That's just my approach though. Essentially all you need to do is remove any of these ('authenticationRoute', 'routeAfterAuthentication' and 'routeIfAlreadyAuthenticated') from your config/environment.js file and instead set them on your routes. My method was just a way of doing it conveniently since all of my authenticated routes would be pointing back to the same login route.

Sorry for the long rant, hope that helps!

Floby:
Thanks ! @dknutsen

PR #992, opened 2016-06-04
Cleanup transition usage in authenticated and unauthenticated route mixins

acburdine:
no issue
- fixes potential test timing issue
- removes unecessary abort call

@marcoow This became necessary as I was working on some ember-simple-auth refactoring for Ghost. In this particular use case, a parent route implemented the unauthenticated route mixin, and in the tests there was several checks to ensure each child route was redirected correctly (see the tests [here](https://github.com/TryGhost/Ghost-Admin/blob/master/tests/acceptance/setup-test.js#L26-L46))

Basically what was happening is the `transition.abort()` call in the unauthenticated route mixin was causing the transition to abort, which then allowed the `andThen` method in the tests to run, before the second transition was started. This only occurred when more than one `visit`/`andThen` combination was used in the same `it` block.

iirc, `transitionTo` calls `abort` internally, so those calls in each mixin were unnecessary. (see [here](https://gist.github.com/machty/5723945#abort-and-retry-transitions))


marcoow:
👍🏻

PR #993, opened 2016-06-06
[BUGFIX] Remove Ember.Logger

Serabe:
Ember.Logger is not substituted by noops in production.

More info in emberjs/guides#1467


Serabe:
Locally, I get two failing tests, but those are failing in master as well. Submitting this and, if travis fails, I'll try to get a deeper look at that.

marcoow:
👍🏻

PR #1001, opened 2016-06-18
Add documentation on how to add current user

### LINKED ISSUE #966: Document an example of how to set/get the current user
There seem to be many examples, with wildly different approaches, as to how to set and get the current user (for displaying the name of the user in the nav bar, or displaying an account page for the current user to modify their user record). The problem is that all of them are outdated and the code samples do not work anymore. It would be great if the readme provided an example that works with the current version of Ember Simple Auth since this would be a common request.

kylemellander:
This adds a separate readme with instructions on how to add a current user.  It is kept separate because it is not core functionality. 

Closes #966

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,7 +51,7 @@ requests such as Ember Data requests__.
 ## How do I use it?
 
 __Ember Simple Auth comes with a
-[dummy app](https://github.com/simplabs/ember-simple-auth/tree/master/tests/dummy)
+[dummy app](tests/dummy)
 that implementes a complete auth solution__ including authentication against
 the application's own server as well as Facebook, authorization of Ember Data
 requests and error handling. __Check out that dummy app for reference.__ To
@@ -572,6 +572,10 @@ module in the application's namespace:
 import { currentSession, authenticateSession, invalidateSession } from '<app-name>/tests/helpers/ember-simple-auth';
 ```
 
+## Other guides
+
+* [Managing Current User](guides/set-current-user.md)
+
 ## Configuration
 
 Ember Simple Auth is configured via the `'ember-simple-auth'` section in the
@@ -592,7 +596,7 @@ for the available settings.
 Ember Simple Auth is developed by and &copy;
 [simplabs GmbH/Marco Otte-Witte](http://simplabs.com) and contributors. It is
 released under the
-[MIT License](https://github.com/simplabs/ember-simple-auth/blob/master/LICENSE).
+[MIT License](LICENSE).
 
 Ember Simple Auth is not an official part of [Ember.js](http://emberjs.com) and
 is not maintained by the Ember.js Core Team.
--- a/guides/set-current-user.md
+++ b/guides/set-current-user.md
@@ -1,0 +1,134 @@
+[Back to Main README](README.md)
+
+## Managing a Current User
+
+Although there are various ways to load currentUser data, this is the canonical
+way to do so. Your authentication route will need to identify how it can access
+the data for the current user. There are 2 examples: one loading through a
+userId that is sent through the auth payload and another accessing a specific
+endpoint (i.e. '/users/me').
+
+### CurrentUser Service
+
+In order for the app to keep the user data, we will store it in its own service:
+
+#### Example with user/me endpoint
+
+```js
+// app/services/session-account.js
+import Ember from 'ember';
+
+const { inject: { service }, isEmpty, RSVP } = Ember;
+
+export default Ember.Service.extend({
+  store: service(),
+
+  account: null,
+
+  loadCurrentUser() {
+    return new RSVP.Promise((resolve, reject) => {
+      return this.get('store').find('user', 'me').then(
+        user => resolve(this.set('user', user)),
+        reject
+      );
+    });
+  }
+});
+```
+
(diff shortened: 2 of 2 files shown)


marcoow:
Looks good!

I'd call the new folder `guides` though.

readme/set-current-user.md
```diff
@@ -0,0 +1,92 @@
+[Back to Main README](https://github.com/simplabs/ember-simple-auth/blob/master/README.md)
+
+## How to Set currentUser Data
+
+Although there are various ways to load currentUser data, this is the canonical
+way to do so. Your authentication route will need to send the id of the user
+account. In this example, it is sent as the user_id.
```
marcoow: This is not actually required. The API could also expose an endpoint `/accounts/me` that returns the current user it identifies based on the authorization header.

readme/set-current-user.md
```diff
@@ -0,0 +1,92 @@
...
+
+  _loadCurrentUser() {
+    return this.get('sessionAccount').loadCurrentUser();
+  }
+});
+```
+
+This sets uses the service to set the when the application route loads or when
+a session is authenticated.  In this instance, if you can't load the current
+user, then it will not log you in.
```
marcoow: This should also explain why the current user needs to be loaded at 2 different places in the code - one if the application already starts up authenticated, one if the session is authenticated at a later point in time.

README.md
```diff
@@ -553,6 +553,10 @@ export default Base.extend({
 });
 ```
 
+## Setting Current User as a part of the session
+
+[README for setting current user](https://github.com/simplabs/ember-simple-auth/blob/master/readmes/set-current-user.md)
+
```
marcoow: can we make this a relative link so it would link to the respective file in a branch when looking at that branch?

guides/set-current-user.md
```diff
@@ -0,0 +1,97 @@
+[Back to Main README](https://github.com/simplabs/ember-simple-auth/blob/master/README.md)
+
+## How to Set currentUser Data
+
+Although there are various ways to load currentUser data, this is the canonical
+way to do so. Your authentication route will need to identify how it can access
+the data for the current user. In this example, it is sent as the user_id on the
+authentication payload.
```
marcoow: I think we should also include the scenario where instead of having the actual user id at hand you have a dedicated endpoint for loading the user record that belongs to the token (or whatever authentication info you have), e.g. `GET /users/me`. This would actually be what I'd advise most users to use as e.g. Auth 2.0 defines the response pretty strictly and wouldn't allow an additional `user_id` in the response at all.

guides/set-current-user.md
```diff
@@ -0,0 +1,97 @@
...
+
+  account: null,
+
+  loadCurrentUser() {
+    return new RSVP.Promise((resolve, reject) => {
+      const userId = this.get('session.data.authenticated.user_id');
+      if (!isEmpty(userId)) {
+        return this.get('store').find('user', userId).then((user) => {
+          this.set('user', user);
+          resolve();
```
marcoow: this should resolve with the user record

guides/set-current-user.md
```diff
@@ -0,0 +1,97 @@
...
+      }
+    });
+  }
+});
+```
+
+In this example, we grab the user id and then make an ember-data request to the
+API. After a successful request, it sets the user and is easily access anywhere
+by injecting this service.  Note: If no id is sent, then the user will not be
+set and this implementation will log you out.
```
marcoow: that's not true - the above method will not log the user out but simply not do anything.

guides/set-current-user.md
```diff
@@ -0,0 +1,97 @@
...
+  }
+});
+```
+
+This uses the session-account service you created to set the currentUser when
+the application route loads and there is a session that has successfully been
+restored (using the beforeModel() hook), or when a session is authenticated
+(using the sessionAuthenticated() hook). Notice that in the
+sessionAuthenticated() hook, if we can't load the current user, then it will
+invalidate our session.
```
marcoow: This should also explain why it's necessary to call that method from 2 different places - `beforeModel` for that case that the application already starts up authenticated and `sessionAuthenticated` if the session becomes authenticated via the tab/windows syncing mechanism.

marcoow:
Looks pretty cool - I added a few comments. I think we should call this _"managing a current user"_ or so though and add a section for guides like this to the table of contents introduced with #1002

kylemellander:
@marcoow I've addressed your comments.

marcoow:
@kylemellander: I created a branch `docs-refactor`. Can you submit a new PR targeting that branch? That would make collaborating on these docs things easier.

marcoow:
this has been replaced by #1040

jrhe:
If the app tries accessing the user property before the user is loaded, an error will likely occur. Do you have any suggestions around this? Resolving in the find request in load fixes this for beforeModel but there doesn't seem to be a way to get sessionAuthenticated to wait on a promise. Perhaps there is a different hook that can be used?

marcoow:
Right, `sessionAuthenticated` will fire immediately when the session becomes authenticated. If you need the current user in that method you can just call the current user service's `load` method there.

PR #1002, opened 2016-06-18
Add table of contents to readme

kylemellander:
This gives a table of contents to help make information organized and to help people to access the information they need.  This also allows us to create additional, less core documentation in the readme subfolder without cluttering the main readme.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,6 +16,43 @@ minimal requirements with respect to application structure, routes etc. With
 its pluggable strategies it __can support all kinds of authentication and
 authorization mechanisms__.
 
+# Table of Contents
+
+**Information**
+
+* [What does it do?](#what-does-it-do)
+* [How does it work?](#how-does-it-work)
+* [Example App](#example-app)
+
+**Usage**
+
+* [Installation](#installation)
+* [Walkthrough](#walkthrough)
+
+**Core Feature Guides**
+
+* [The Session Service](#the-session-service)
+* [Authenticators](#authenticators)
+  * [Customizing an Authenticator](#customizing-an-authenticator)
+  * [Implementing a custom Authenticator](#implementing-a-custom-authenticator)
+* [Authorizers](#authorizers)
+  * [Customizing an Authorizer](#customizing-an-authorizer)
+  * [Implementing a custom Authorizer](#implementing-a-custom-authorizer)
+* [Session Stores](#session-stores)
+  * [Store Types](#store-types)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -16,6 +16,44 @@ minimal requirements with respect to application structure, routes etc. With
...
+**Other Resources**
+
+* [Configuration](https://github.com/simplabs/ember-simple-auth/blob/master/README.md#configuration)
+* [Upgrading from Pre-1.0 versions](http://log.simplabs.com/post/131698328145/updating-to-ember-simple-auth-10)
+* [Implementing Current User Data](https://github.com/simplabs/ember-simple-auth/blob/master/readme/set-current-user.md)
+* [API Documentation](http://ember-simple-auth.com/api/)
+
+**License & Contributing**
+
+* [License](https://github.com/simplabs/ember-simple-auth/blob/master/README.md#license)
```
marcoow: I don't think we need this in the TOC

marcoow:
Can we make the links so they only use e.g. `#what-does-it-do` so that they would also work on different branches etc.?

README.md
```diff
@@ -16,6 +16,40 @@ minimal requirements with respect to application structure, routes etc. With
 its pluggable strategies it __can support all kinds of authentication and
 authorization mechanisms__.
 
+# Table of Contents
+
+**Information**
+
+* [What does it do?](#what-does-it-do)
+* [How does it work?](#how-does-it-work)
+* [Working example](#how-do-i-use-it)
```
marcoow: I think we should rename the header to _"Example app"_ or so well. _"How do I use it?"_ doesn't seem like a great header for that…

README.md
```diff
@@ -16,6 +16,40 @@ minimal requirements with respect to application structure, routes etc. With
...
+
+**Usage**
+
+* [Installation](#installation)
+* [Basic Usage](#basic-usage)
```
marcoow: Maybe _"Walkthrough"_ or so might be better than _"Basic Usage"_?

README.md
```diff
@@ -16,6 +16,40 @@ minimal requirements with respect to application structure, routes etc. With
...
+
+**Core Building Blocks**
```
marcoow: Maybe we should rename this to _"Guide"_ or so as it actually talks about testing, implementing custom authenticators etc. as well?

marcoow:
this has been replaced by #1040

PR #1010, opened 2016-06-24
Disable 'sessionDataUpdated' trigger if window is active for IE

jevanlingen:
According to the spec the `storage-event` should only be fired **for other** tabs and windows, see https://html.spec.whatwg.org/multipage/webstorage.html#the-localstorage-attribute.

However, IE does fail to do so: https://connect.microsoft.com/IE/feedback/details/798684/ie-localstorage-event-misfired. To bring this behaviour to IE we can emulate it by using the `document.visibilityState`. Pity enough this works only for IE10 en IE11, so older versions of Internet Explorer still suffer the extra unneeded session-data update.

### DIFF
--- a/addon/session-stores/local-storage.js
+++ b/addon/session-stores/local-storage.js
@@ -83,7 +83,7 @@ export default BaseStore.extend({
 
   _bindToStorageEvents() {
     jQuery(window).bind('storage', (e) => {
-      if (e.originalEvent.key === this.key) {
+      if (e.originalEvent.key === this.key && (document.visibilityState typeof undefined || document.visibilityState === 'hidden')) {
         this.restore().then((data) => {
           if (!objectsAreEqual(data, this._lastData)) {
             this._lastData = data;
--- a/addon/session-stores/local-storage.js
+++ b/addon/session-stores/local-storage.js
@@ -83,7 +83,7 @@ export default BaseStore.extend({
 
   _bindToStorageEvents() {
     jQuery(window).bind('storage', (e) => {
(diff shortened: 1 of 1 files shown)


marcoow:
What does this fix specifically?

jevanlingen:
1. No need to call the restore data function for open tab (because it did initiate the storage change)
2. I got a custom keycloak adaption. Things go weird in IE if restore is called upon when logging in

marcoow:
> 2. I got a custom keycloak adaption. Things go weird in IE if restore is called upon when logging in

Can you provide more precise info on this please, it's a bit vague actually ;)

jevanlingen:
In the end it was more an implementation problem then a bug within Ember Simple Auth. So closed this pull.

**Explanation**
 it was a combination of things:
1. Keycloak restore: do a reload
2. After reload, set `_bindToStorageEvents`
3. IE always executed the `storage-event` function, even if it's the setter is within the same window
4.. I saved an Ember Data model in the session, so the if-statement `!objectsAreEqual(data, this._lastData)` in the `storage-event` was always true (serialized model against 'real' model).
5. Trigger the `sessionDataUpdated` flag.
6. If  `sessionDataUpdated` is triggerd, call restore
7. Go to step 1...

To fix this endless loop, I added the `document.hidden` check. But this does not fix the problem for multiple tabs. So I choose to save the id of the model in the session only, to prevent this loop from happening.

P.S. If you wonder why other browers do not have the same problem if multiple tabs are used, it's because the `storage-event` is not fired when the localstorage is updated but the localstorage data is not changed at all.

PR #1012, opened 2016-06-24
Add rejectWithXhr to optionally reject with XHR vs response body

f3ndot:
Allows ember apps using ember-simple-auth to receive the whole XHR object if the backend fails, instead of the response body, if they so choose.

In the case of OAuth 2.0 backends, it's been a pattern in the wild to use X- headers to send context as to why a grant has failed. Examples include API throttling, brute force lockouts, and [OTP/two-factor authentication](https://developer.github.com/v3/auth/#working-with-two-factor-authentication) information.

Selfishly, I require this change so my application can be notified when the API has locked out an account due to suspicious activity via an X- header.

The decision to expose it as an option was chosen so backwards compatibility is maintained and keeps the addon simple for those who need not be concerned with complex backends.

### DIFF
--- a/tests/unit/authenticators/devise-test.js
+++ b/tests/unit/authenticators/devise-test.js
@@ -104,13 +104,31 @@ describe('DeviseAuthenticator', () => {
 
     describe('when the authentication request fails', () => {
       beforeEach(() => {
-        server.post('/users/sign_in', () => [400, { 'Content-Type': 'application/json' }, '{ "error": "invalid_grant" }']);
+        server.post('/users/sign_in', () => [400, { 'Content-Type': 'application/json', 'X-Custom-Context': 'foobar' }, '{ "error": "invalid_grant" }']);
       });
 
       it('rejects with the correct error', (done) => {
         authenticator.authenticate('email@address.com', 'password').catch((error) => {
           expect(error).to.eql({ error: 'invalid_grant' });
           done();
+        });
+      });
+
+      describe('when reject with XHR is enabled', () => {
+        beforeEach(() => {
+          authenticator.set('rejectWithXhr', true);
+        });
+
+        it('rejects with xhr object', () => {
+          return authenticator.authenticate('username', 'password').catch((error) => {
+            expect(error.responseJSON).to.eql({ error: 'invalid_grant' });
+          });
+        });
+
+        it('provides access to custom headers', () => {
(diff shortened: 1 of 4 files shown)


tests/unit/authenticators/devise-test.js
```diff
@@ -113,6 +113,26 @@ describe('DeviseAuthenticator', () => {
...
+        beforeEach(() => {
+          authenticator.set('rejectWithXhr', true);
+        });
+
+        it('rejects with xhr object', (done) => {
+          authenticator.authenticate('username', 'password').catch((error) => {
+            expect(error.responseJSON).to.eql({ error: 'invalid_grant' });
+            done();
+          });
+        });
```
marcoow: can you change this to

```js
it('rejects with xhr object', (done) => {
  authenticator.authenticate('username', 'password').catch((error) => {
    expect(error.responseJSON).to.eql({ error: 'invalid_grant' });
    done();
  });
});
```

tests/unit/authenticators/devise-test.js
```diff
@@ -113,6 +113,26 @@ describe('DeviseAuthenticator', () => {
...
+
+        it('provides access to custom headers', (done) => {
+          authenticator.authenticate('username', 'password').catch((error) => {
+            expect(error.getResponseHeader('X-Custom-Context')).to.eql('foobar');
+            done();
+          });
+        });
```
marcoow: see above

tests/unit/authenticators/oauth2-password-grant-test.js
```diff
@@ -199,6 +199,26 @@ describe('OAuth2PasswordGrantAuthenticator', () => {
...
+            done();
+          });
+        });
+
+        it('provides access to custom headers', (done) => {
+          authenticator.authenticate('username', 'password').catch((error) => {
+            expect(error.getResponseHeader('X-Custom-Context')).to.eql('foobar');
+            done();
+          });
+        });
```
marcoow: see above for these as well - we should generally not use `done` when returning the promise works as well

marcoow:
Sorry for the late reply - was super busy recently.

This looks really good - just a few comments wrt test style.

f3ndot:
@marcoow excellent. Thank you! I've made the amendments as you've requested.

PR #1017, opened 2016-07-08
fix a bug preventing users from signing into another account

fivetwentysix:
when I logout, and log back in, content is sometimes undefined and I have to fetched it out of restoredContent.

Is anyone having the same issue?


marcoow:
Can you add a bit more info about what problem this is fixing specifically?

fivetwentysix:
I have no idea but looks like my auth works without this change now.

PR #1026, opened 2016-07-23
Use Object.keys since Ember.keys is deprecated

f3ndot:
This removes a deprecation warning in Ember 1.13 and later that is generated every time the `authenticate` function is called.

screen shot 2016-07-23 at 12 35 52 am

### DIFF
--- a/addon/authenticators/oauth2-password-grant.js
+++ b/addon/authenticators/oauth2-password-grant.js
@@ -14,9 +14,10 @@ const {
   $: jQuery,
   testing,
   warn,
-  keys
+  keys: emberKeys
 } = Ember;
 const assign = emberAssign || merge;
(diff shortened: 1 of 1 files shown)


f3ndot:
I realize `Object.keys` is not available in IE8 but that's going to be/is dropped in Ember 2.x

So dunno what you want to do there

marcoow:
ESA 1.x still supports Ember 1.13 so we'd have to make sure the library is compatible with IE 8 as well. I guess we could simply do

```js
const keys = Object.keys || Ember.keys;
```

f3ndot:
Alright, I'm clearly too tired to YOLO author this on GitHub. I'll squash and push up a real fix in time.

f3ndot:
@marcoow All set!

PR #1035, opened 2016-07-28
FastBoot Compatibility

marcoow:
This makes Ember Simple Auth compatible with FastBoot.

The goal of this is for Ember Simple auth to support FastBoot out of the box without the need for anyone to change anything to their ESA setup (except for switching to the cookie session store as `localStorage` doesn't work in FastBoot obviously).
### TODO
- [x] ~~replace [`window.btoa`](https://github.com/simplabs/ember-simple-auth/blob/master/addon/authenticators/oauth2-password-grant.js#L125) with sth. that works in FastBoot~~ (FastBoot exposes a `window` global (which is the sandbox) which means this works already)
- [x] replace [`window.location`](https://github.com/simplabs/ember-simple-auth/blob/master/addon/mixins/application-route-mixin.js#L123) with sth. that works in FastBoot
- [x] don't try [binding to `localStorage` events with jQuery](https://github.com/simplabs/ember-simple-auth/blob/master/addon/session-stores/local-storage.js#L85) when running in FastBoot
- [x] inject the FastBoot service into the cookie store manually in an initializer
- [x] write integration tests using ember-cli-addon-tests
- [x] use `ember-network` instead of jQuery's AJAX methods, see #1066 
- [x] replace occurences of jQuery, `window` etc. in the docs with sth. that works in FastBoot
- [x] fix tests (FastBoot tests seem to fail randomly - must be some sort of race condition or so)


marcoow:
rebased on `master`

addon/session-stores/cookie.js
```diff
@@ -143,8 +166,7 @@ export default BaseStore.extend({
   },
 
   _read(name) {
-    let value = document.cookie.match(new RegExp(`${name}=([^;]+)`)) || [];
-    return decodeURIComponent(value[1] || '');
+    return this.get('_cookies').read(name) || '';
```
mhluska: This is throwing an error in FastBoot because recent versions of Ember don't like injected properties in objects without a container:

```
Error: Assertion Failed: Attempting to lookup an injected property on an object without a container, ensure that the object was instantiated via a container.
```

arjansingh:
Forgot to rename some of the commits but let's make sure to get rid of the `WIP` and `DRAFT` commit messages when we do the final rebase. Sorry about that.

addon/session-stores/cookie.js
arjansingh: I ran into this... the solution (which might have been rebased out) was to manually inject the cookies service. Did the rebasing mess that up?

https://github.com/simplabs/ember-simple-auth/blob/2d49133ec8543ae53af713d7551072ba49f65ce7/addon/initializers/setup-session.js#L17

marcoow: @arjansingh: yeah, I might have messed that up during rebasing 😳

asleroid:
Can you please take a look at this issue regarding this Pull Request's latest commit:

https://github.com/simplabs/ember-simple-auth/issues/1043

addon/session-stores/cookie.js
josemarluedke: I can confirm that this is a problem.

marcoow: I'm going to fix that. Sorry for that - I really shouldn't be breaking great work of people who want to contribute 😔

josemarluedke:
@marcoow I have a open PR that adds integration testing for fastboot in #1045. These tests are basic and can be improved over time, but I think they are a good start.

I think with that, we can say we have a basic support for fastboot. 🎉

marcoow:
@josemarluedke: awesome - thanks for the hard work! Will review and merge as soon as possible.

briangonzalez:
@marcoow Any idea when this will land?

marcoow:
@briangonzalez: we're hoping soon ;)

A bunch of people are working on it and also related changes that are necessary in ember-cookies. I just returned from vacation and hope we manage to release a beta this month.

addon/session-stores/cookie.js
marcoow: This is actually only generating an error when using the adaptive store. Of course that should work as well as the adaptive store should simply fallback to the cookie store and work out of the box (although the session would not actually be transported between the client and server then as the client would use `localStorage`). It should certainly not lead to an error though.

simonihmig:
Would you consider using [ember-fastboot-addon-tests](https://github.com/kaliber5/ember-fastboot-addon-tests) for the Fastboot tests? It would dry up the tests and remove some boilerplate code. Should be fairly easy to do, as the implementation is pretty similar. I could put a PR together if there are no objections! :)

marcoow:
@simonihmig: not sure what ember-fastboot-addon-tests would simplify, can you elaborate? At this point I'm a bit hesitant to add more changes onto this as this has been open for quite some time already. that said, adding ember-fastboot-addon-tests could always be done after the initial 1.2 beta release of course.

simonihmig:
It would probably remove the need for these files or parts of them:
* https://github.com/simplabs/ember-simple-auth/blob/fastboot/node-tests/fastboot-runner.js
* https://github.com/simplabs/ember-simple-auth/blob/fastboot/node-tests/fastboot/basic-test.js#L3-L30
* https://github.com/simplabs/ember-simple-auth/blob/fastboot/node-tests/fastboot/basic-test.js#L67-L72

This is just a minor optimization, and should certainly not block the release of 1.2 in any way. So if you feel better doing this afterwards, that would be fine as well!

marcoow:
Yeah, let's look at that again after the 1.2 beta release

simonihmig:
@marcoow No problem. Just added #1097 to not forget about this. Looking forward to 1.2 myself! :)

lolmaus:
Hi!

I'm building a FastBoot app and I successfully added GitHub auth with ESA, Torii and Gatekeeper. But I discovered that ESA 1.1.0 completely prevents static HTML from generating! The HTML file ends up just like it would without FastBoot.

So I tried ESA `fastboot` branch and ran into issues.

First of all, it crashed on [this line] with `_fastboot.request.host` being `undefined`.

Also, looking at this code I saw another problem: it's incorrect to search for protocol name in `request.host` because according to FastBoot docs that property only contains the hostname.

I only want HTML for anonymous users to be generated, so I tried switching to the localStorage session store. When I updated the `restore` method to return a prerejected promise in case of `isFastBoot`, FastBoot started generating HTML correctly.

But on the `fastboot` branch there appeared another crazy problem: in a controller, `model` comes as `null`! I return an `RSVP.hash` from the route's `model` hook, but it ends up `null`. Simply removing ESA-fastboot from `package.json` resolves that issue.

lolmaus:
OK, I figured out the `null` model.

It happens because I short-circuit my `localStorage` adapter in FastBoot mode, having it return a prerejected projmise. According to documentation, this is valid behavior:

> @return {Ember.RSVP.Promise} A promise that resolves with the data currently persisted in the store when the data has been restored successfully and rejects otherwise.

When I replaced the short circuit to returning a preresolved model instead of prerejected, the problem is gone: model resolves just fine.

This issue may be related to [this code], even though on a second thought it shouldn't cause this effect:

```js
  applicationRoute.reopen({
    beforeModel() {
      return session.restore().then(
        () => applyOriginalBeforeModel(...arguments),
        () => applyOriginalBeforeModel(...arguments)
      );
    }
  });
```

marcoow:
@lolmaus: what does the `model` method look like for the route that the model is `null` for? Are you using the `DataAdapterMixin` in the respective adapter?

lolmaus:
@marcoow 

1. https://github.com/lolmaus/lolma.us/blob/1cf778c53d2ee1601843cf7452d985e8e608aaf4/app/pods/locale/route.js#L28-L55
2. No, I'm using the base `ember-data/adapter` driven by `ember-network/fetch`.

It's very weird that the model returns as `null` without error. I return `RSVP.hash` from the model hook, so the model should always be a promise even if there was a crash inside the promise.

marcoow:
[`RSVP.hash` only fulfill when _all_ of the individual promises fulfill](https://github.com/tildeio/rsvp.js#hash-of-promises).

lolmaus:
@marcoow So what? If any (or even all) child promises fail, the resulting promise will still be a promise. It will never turn into `null`.

marcoow:
[This](https://github.com/lolmaus/lolma.us/blob/1cf778c53d2ee1601843cf7452d985e8e608aaf4/app/pods/application/route.js#L41) will reject when you're not running on FastBoot. I'm not sure what a subsequent call to `modelFor` will result in.

lolmaus:
@marcoow Everything is fine in non-FastBoot mode. The problem happens in FastBoot mode.

And again: even if it rejects, I should see a (rejected) promise in controller's `model` property. This is true when there's no ESA in the project. Adding ESA `fastboot` branch turns the `model` in controller into `null`.

lolmaus:
PS I'm available in Ember Slack in ESA channel.

lolmaus:
Filed #1104 to address the cookie store trying to find protocol name in the `host` property.

lolmaus:
:scream_cat: :sparkles: :champagne:

josemarluedke:
🎉 🍾 :tada::tada::tada:

marcoow:
Thanks for the great work everyone! Will publish and announce tomorrow.

addon/mixins/authenticated-route-mixin.js
```diff
@@ -72,7 +79,11 @@ export default Mixin.create({
       let authenticationRoute = this.get('authenticationRoute');
       assert('The route configured as Configuration.authenticationRoute cannot implement the AuthenticatedRouteMixin mixin as that leads to an infinite transitioning loop!', this.get('routeName') !== authenticationRoute);
 
-      this.set('session.attemptedTransition', transition);
+      if (!this.get('_isFastBoot')) {
+        transition.abort();
```
juanazam: Submitted: https://github.com/simplabs/ember-simple-auth/pull/1126 😄

PR #1040, opened 2016-08-01
Docs refactor

marcoow:
This combines several improvements around documentation.

## TODO

- [ ] ~~I think we should rename the `session-account` service in the dummy app to `current-user` as well (and its methods and properties) so that it matches the example in the guide. Otherwise the 2 implementation might lead to confusion as they'd use different naming for essentially the same things making it harder to see that both are actually the same.~~ doesn't really work as the model is called `Account` instead of `User` so that would probably be even more confusing.


marcoow:
@kylemellander: can you review again?

guides/managing-current-user.md
```diff
@@ -23,24 +61,19 @@ const { inject: { service }, isEmpty, RSVP } = Ember;
...
 
-  account: null,
-
   loadCurrentUser() {
-    return new RSVP.Promise((resolve, reject) => {
-      return this.get('store').find('user', 'me').then(
-        user => resolve(this.set('user', user)),
-        reject
-      );
+    return this.get('store').find('user', 'me').then((account) => {
```
kylemellander: maybe we should keep the user variable name here to keep it consistent with the find action to keep things clear.

kylemellander:
@marcoow - Just one little thing with variable names, but overall, solid.

guides/managing-current-user.md
marcoow: Yeah, makes sense - I'll rename the service to `currentUser` or so as well.

marcoow:
@kylemellander: fixed

marcoow:
oops - wrong button

PR #1041, opened 2016-08-02
Better route config docs

marcoow:
This improves the docs and deprecations for the routes that ESA transitions to in different scenarios.

### DIFF
--- a/addon/configuration.js
+++ b/addon/configuration.js
@@ -47,7 +47,7 @@ export default {
     route is not authenticated.
 
     @property authenticationRoute
-    @deprecated 'deprecated in favor of overriding the authenticationRoute method on the authenticatedRouteMixin'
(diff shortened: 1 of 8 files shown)


addon/configuration.js
```diff
@@ -47,7 +47,7 @@ export default {
     route is not authenticated.
 
     @property authenticationRoute
-    @deprecated 'deprecated in favor of overriding the authenticationRoute method on the authenticatedRouteMixin'
+    @deprecated AuthenticatedRouteMixin/authenticationRoute:property
```
marcoow: YUIDoc doesn't allow crosslinks or any formatting in the deprecation message. So instead of providing an actual text here we simply only provide the reference to the things that this elements has been deprecated in favor of so that we can link to that in the docs.

PR #1047, opened 2016-08-06
Small fixes

stevenwu:
@marcoow the remaining failing tests concern cookie domain. I don't think it's possible to see the domain in a `document.cookie` string.

### DIFF
--- a/addon/session-stores/cookie.js
+++ b/addon/session-stores/cookie.js
@@ -189,7 +189,7 @@ export default BaseStore.extend({
       document.cookie = `${this._oldCookieName}=`;
     }
 
-    if (isBlank(value)) {
+    if (isBlank(value) && expiration !== 0) {
       return;
     }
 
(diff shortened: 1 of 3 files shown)


addon/session-stores/cookie.js
```diff
@@ -189,7 +189,7 @@ export default BaseStore.extend({
       document.cookie = `${this._oldCookieName}=`;
     }
 
-    if (isBlank(value)) {
+    if (isBlank(value) && expiration !== 0) {
```
marcoow: I don't think this is correct. If the expiration is a number other than 0 you'd still want to write the cookie - it would just expire after `expiration` seconds.

stevenwu: This is for detecting a cookie clear which runs `this._write(null, 0)`. What would be a better way of handling this?

marcoow: I think the point is that in order to clear a cooke you need to write it with an expiration time of 0 or am I missing something?

PR #1053, opened 2016-08-16
Update managing-current-user.md

maletor:


marcoow:
What's the motivation for this?

maletor:
If you use the second implementation of `load()` where it returns `Ember.RSPV.Promise`, it doesn't work when the app first boots. Maybe that code needs to be rewritten instead of the `beforeModel()` code but I couldn't figure out why.

marcoow:
Hm, actually you'll want to return the promise so that Ember would enter/remain in the loading state while the current user is being loaded. Does the promise you're returning reject maybe?

maletor:
It's possible. I was using this code you wrote but could only get a blank page. Then I switched to code under

> Using a dedicated endpoint

Finally, I decided I would forget loading the current user information before the route itself loaded by setting the handlebars template to the return value of the store's `findRecord` method. 

Since the model can't be resolved without a valid key, loading the user information such as a profile photo or a name can be deferred until the route itself loads. At least for me, and probably the majority of cases, the current user information is not important enough to be required to load before the route is entered.

```javascript
// app/services/current-user.js
import Ember from 'ember';

const { inject: { service }, isEmpty, RSVP } = Ember;

export default Ember.Service.extend({
  session: service('session'),
  store: service(),

  load() {
      let userId = this.get('session.data.authenticated.user_id');
      if (!isEmpty(userId)) {
          this.set('user', user);
        }, reject);
      } else {
        resolve();
      }
    });
  }
});
```

marcoow:
Making sure the loading state is active while the current user is being loaded is actually an integral part of the strategy described in the guide. It both ensures a consistent UI state and lets people depend on the current user being present once the route has loaded.

Of course you can always opt out of that by not returning the promise from the `beforeModel` but I don't think it should be in the guide like that.

maletor:
What about the current route not being able to load because the auth token returned 401 and already rejected the session? 

It's not really possible for an `AuthenticatedRoute` to load that way.

Even if it does, when it's trying to render the full name of a user it will trigger a 401 at the users route which will again logout the session. 

So making it asynchronous should always work, don't you think.

marcoow:
Logging the user out when the server responds with a 401 is intended behavior really. I'm not sure what the problem with that is.

> So making it asynchronous should always work, don't you think?

I assume you're referring to not returning the promise from the `beforeModel` method? The problem with that is that the current user would be loaded and thus not accessible when the `model` method runs which is what we want though. A lot of people expect the current user to be loaded when the session is authenticated and want to use it in e.g. their `model` methods. The solution described in this guide offers a way of doing so, thus solving the way the current user is handled for a majority of users.

PR #1055, opened 2016-08-17
Remove top-level Ember imports from generated authenticators

balinterdi:
`Ember` is not used in any of the generated authenticators and only makes jshint sad.

### DIFF
--- a/blueprints/authenticator/index.js
+++ b/blueprints/authenticator/index.js
@@ -15,25 +15,25 @@ module.exports = {
 
     if (baseClass === 'torii') {
       return {
-        imports: 'import Ember from \'ember\';' + EOL + 'import Torii from \'ember-simple-auth/authenticators/torii\';',
+        imports: 'import Torii from \'ember-simple-auth/authenticators/torii\';',
         baseClass: 'Torii',
         body: EOL + '  torii: Ember.inject.service(\'torii\')'
       };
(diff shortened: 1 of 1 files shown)


blueprints/authenticator/index.js
```diff
@@ -15,25 +15,25 @@ module.exports = {
 
     if (baseClass === 'torii') {
       return {
-        imports: 'import Ember from \'ember\';' + EOL + 'import Torii from \'ember-simple-auth/authenticators/torii\';',
+        imports: 'import Torii from \'ember-simple-auth/authenticators/torii\';',
         baseClass: 'Torii',
         body: EOL + '  torii: Ember.inject.service(\'torii\')'
```
mike183: I've not actually tested this, but just happened to notice that this authenticator does appear to use the imported Ember module.

balinterdi: I removed that line because I received a jslint error about the unused import. Not sure what the difference is between your setup and mine is, then.

marcoow: That line is actually necessary as the authenticator cannot access torii otherwise but the import should be removed. PR would be awesome!

mike183: @balinterdi I didn't actually see an error, it was just an observation I made when looking through the contents of recent commits.

PR #1056, opened 2016-08-17
Cookie store rewrite

marcoow:
This rewrites the cookie store so that when any of the store's attributes change the cookie will be rewritten as well so that the cookie is actually consistent with the store's attributes.

This replaces #983


marcoow:
There seems to be a problem with clearing cookies. Somehow cookies are still there even after they are cleared with e.g.

```js
document.cookie = 'ember_simple_auth-session=; path=/; expires=Thu, 01 Jan 1970 00:00:00 GMT';
```

I'm not really sure what might be causing that but that needs to be fixed.

Issue #1054, opened 2016-08-17
OAuth2 Implicit Grant Flow Authenticator


256dpi:
Hi! I'm currently working on a OAuth2 Implicit Grant Flow authenticator for ember-simple-auth as part of the [fire](https://github.com/256dpi/fire) framework and the [fire-auth](https://github.com/256dpi/fire-auth) add-on:

Source: https://github.com/256dpi/fire-example/blob/master/ui/app/authenticators/oauth2.js

The authenticator can be used in conjunction with the already built-in OAuth2 Bearer authorizer to authenticate and authorizer users using the OAuth2 Implicit Grant Flow. As I plan to use the implementation in multiple projects I was wondering if you would be interested in adding the implementation to the ember-simple-auth repository?

I thought about creating an extension module, but it seems to me a little too much boilerplate for just this one class...

Implementation Details: The Authenticator must be called in the callback route that receives the access token via a HTTP redirection: https://github.com/256dpi/fire-example/blob/master/ui/app/routes/auth/callback.js#L9.

Based on that, the authenticator should support many established providers (google, facebook, ...) as well as custom implementations. It even allows multiple login strategies (normal, pop-up, iframe): https://github.com/256dpi/fire-example/blob/master/ui/app/templates/application.hbs#L9.

If you're interested I would love to submit a PR with some tests.

marcoow:
Looks good, yeah. I remember discussing adding sth. like this to the library with someone else a few months ago. There might be an open PR with sth. similar already. Adding support for the functionality you're describing would definitely be cool. Besides the authenticator we should definitely also include the route mixin for the callback route that triggers authentication with the new authenticator.

256dpi:
I started working on the PR. I'm almost finished, but there is one test case that seems a little complicated: https://github.com/256dpi/ember-simple-auth/blob/implicit-grant/tests/unit/mixins/callback-route-mixin-test.js

The tests currently break as the fake InternalSession seems to have no `lookup` method defined and therefore `authenticate` cannot locate the authenticator to authenticate the request. Any ideas?

Here is the simplified trace:

```
undefined is not an object (evaluating '(0, _emberGetownerPolyfill['default'])(this).lookup')

_lookupAuthenticator@http://localhost:7357/assets/vendor.js:88131:58
authenticate@http://localhost:7357/assets/vendor.js:87915:52
activate@http://localhost:7357/assets/vendor.js:88430:39
```

marcoow:
[...]

PR #1057, opened 2016-08-19
OAuth2 Implicit Grant Authenticator

256dpi:
This PR implements the new implicit grant authenticator as discussed in #1054.

@marcoow Ready to review?

### DIFF
--- a/addon/authenticators/oauth2-implicit-grant.js
+++ b/addon/authenticators/oauth2-implicit-grant.js
@@ -1,0 +1,103 @@
+import Ember from 'ember';
+import BaseAuthenticator from './base';
+
+const {
+  RSVP,
+  isEmpty
+} = Ember;
+
+/**
+ Authenticator that conforms to OAuth 2
+ ([RFC 6749](http://tools.ietf.org/html/rfc6749)), specifically the _"Implicit
+ Grant Type"_.
+
+ @class OAuth2ImplicitGrantAuthenticator
+ @module ember-simple-auth/authenticators/oauth2-implicit-grant
+ @extends BaseAuthenticator
+ @public
+ */
+export default BaseAuthenticator.extend({
+  /**
+   Restores the session from a session data object; __will return a resolving
+   promise when there is a non-empty `access_token` in the session data__ and
+   a rejecting promise otherwise.
+
+   @method restore
+   @param {Object} data The data to restore the session from
+   @return {Ember.RSVP.Promise} A promise that when it resolves results in the session becoming or remaining authenticated
+   @public
+   */
+  restore(data) {
+    return new RSVP.Promise((resolve, reject) => {
+      if (!this._validateData(data)) {
+        return reject();
+      }
+
+      return resolve(data);
+    });
+  },
+
+  /**
+   Authenticates the session using the specified location `hash`
+   (see https://tools.ietf.org/html/rfc6749#section-4.2.2).
+
+   __If the access token is valid and thus authentication succeeds, a promise that
+   resolves with the access token is returned__, otherwise a promise that rejects
+   with the error code as returned by the server is returned
+   (see https://tools.ietf.org/html/rfc6749#section-4.2.2.1).
+
+   @method authenticate
+   @param {Object} hash The optional location hash
+   @return {Ember.RSVP.Promise} A promise that when it resolves results in the session becoming authenticated
+   @public
+   */
+  authenticate(hash) {
+    return new RSVP.Promise((resolve, reject) => {
+      let response = this._parseResponse(hash);
+
+      if (response.error) {
+        reject(response.error);
+      } else if (!this._validateData(response)) {
+        reject('invalid_token');
+      } else {
+        resolve(response);
+      }
+    });
+  },
+
+  /**
+   This method simply returns a resolving promise.
+
+   @method invalidate
+   @return {Ember.RSVP.Promise} A promise that when it resolves results in the session being invalidated
+   @public
+   */
+  invalidate() {
+    return RSVP.Promise.resolve();
+  },
+
+  _validateData(data) {
+    // see https://tools.ietf.org/html/rfc6749#section-4.2.2
+
+    // jscs:disable requireCamelCaseOrUpperCaseIdentifiers
+    return !isEmpty(data) &&
+      !isEmpty(data.access_token) &&
+      !isEmpty(data.state) &&
+      !isEmpty(data.token_type);
+    // jscs:enable requireCamelCaseOrUpperCaseIdentifiers
+  },
+
+  _parseResponse(locationHash) {
+    let params = {};
+    const query = locationHash.substring(locationHash.indexOf('?'));
+    const regex = /([^#?&=]+)=([^&]*)/g;
+    let match;
+
+    // decode all parameter pairs
+    while ((match = regex.exec(query)) !== null) {
+      params[decodeURIComponent(match[1])] = decodeURIComponent(match[2]);
+    }
+
+    return params;
+  }
+});
--- a/addon/mixins/callback-route-mixin.js
+++ b/addon/mixins/callback-route-mixin.js
@@ -1,0 +1,57 @@
+import Ember from 'ember';
+
+const { inject: { service }, Mixin } = Ember;
+
+export default Mixin.create({
+  /**
+   The session service.
+
+   @property session
+   @readOnly
+   @type SessionService
+   @public
+   */
+  session: service('session'),
+
+  /**
+   The authenticator that should be used to authenticate the callback.
+
+   @property authenticator
+   @type String
+   @default null
+   @public
+   */
+  authenticator: null,
+
+  /**
+   The error that ocurred during the authentication attempt.
+
+   @property error
+   @type String
+   @default null
+   @public
+   */
+  error: null,
+
+  /**
+   Authenticates the passed authentication server response and sets the error
+   property if the authentication failed.
+
+   @method activate
+   @public
+   */
+  activate() {
+    let authenticator = this.get('authenticator');
+
+    let hash = '';
+
+    // test if fastboot is available without requiring the fastboot service to be present
+    if (typeof FastBoot === 'undefined') {
+      hash = window.location.hash;
+    }
+
+    return this.get('session').authenticate(authenticator, hash).catch((err) => {
+      this.set('error', err);
+    });
+  }
+});
--- a/tests/unit/authenticators/oauth2-implicit-grant-test.js
+++ b/tests/unit/authenticators/oauth2-implicit-grant-test.js
@@ -1,0 +1,80 @@
+/* jshint expr:true */
+/* jscs:disable requireDotNotation */
+import Ember from 'ember';
+import { it } from 'ember-mocha';
+import { describe, beforeEach, afterEach } from 'mocha';
+import { expect } from 'chai';
+import Pretender from 'pretender';
+import OAuth2ImplicitGrant from 'ember-simple-auth/authenticators/oauth2-implicit-grant';
+
+const { tryInvoke } = Ember;
+
+describe('OAuth2ImplicitGrantAuthenticator', () => {
+  let authenticator;
+  let server;
+
+  let accessTokenHash = '#access_token=secret-token&state=abcd&token_type=bearer';
+
+  const accessToken = {
+    'access_token': 'secret-token',
+    'state': 'abcd',
+    'token_type': 'bearer'
+  };
+
+  beforeEach(() => {
+    authenticator = OAuth2ImplicitGrant.create();
+    server        = new Pretender();
+  });
+
+  afterEach(() => {
+    tryInvoke(server, 'shutdown');
+  });
+
+  describe('#restore', () => {
+    describe('when the data contains an access_token', () => {
+      it('resolves with the correct data', () => {
+        return authenticator.restore(accessToken).then((_data) => {
+          expect(_data).to.eql(accessToken);
+        });
+      });
+
+      describe('when the data does not contain an access_token', () => {
+        it('returns a rejecting promise', (done) => {
+          authenticator.restore().catch(() => {
+            done();
+          });
+        });
+      });
+    });
+  });
+
+  describe('#authenticate', () => {
+    describe('with hash provided', () => {
+      it('authentication should succeed on a proper access token', () => {
+        return authenticator.authenticate(accessTokenHash).then((_accessToken) => {
+          expect(_accessToken).to.eql(accessToken);
+        });
+      });
+
+      it('authentication should fail on an invalid access token', () => {
+        return authenticator.authenticate('#foo=bar').catch((_err) => {
+          expect(_err).to.eql('invalid_token');
+        });
+      });
+
+      it('authentication should fail on a proper error object', () => {
+        return authenticator.authenticate('#error=access_denied').catch((_err) => {
+          expect(_err).to.eql('access_denied');
+        });
+      });
+    });
+  });
+
+  describe('#invalidate', () => {
+    it('returns a resolving promise', (done) => {
+      authenticator.invalidate().then(() => {
+        done();
+      });
+    });
+  });
+});
--- a/README.md
+++ b/README.md
@@ -357,6 +357,7 @@ this.get('session').authenticate('authenticator:some');
 Ember Simple Auth comes with 3 authenticators:
 
 * [`OAuth2PasswordGrantAuthenticator`](http://ember-simple-auth.com/api/classes/OAuth2PasswordGrantAuthenticator.html): an OAuth 2.0 authenticator that implements the _"Resource Owner Password Credentials Grant Type"_
+* [`OAuth2ImplicitGrantAuthenticator`](http://ember-simple-auth.com/api/classes/OAuth2ImplicitGrantAuthenticator.html): an OAuth 2.0 authenticator that implements the _"Implicit Grant Type"_
 * [`DeviseAuthenticator`](http://ember-simple-auth.com/api/classes/DeviseAuthenticator.html): an authenticator compatible with the popular Ruby on Rails authentication plugin [devise](https://github.com/plataformatec/devise)
 * [`ToriiAuthenticator`](http://ember-simple-auth.com/api/classes/ToriiAuthenticator.html): an authenticator that wraps the [torii library](https://github.com/Vestorly/torii)
 
--- a/tests/unit/mixins/callback-route-mixin-test.js
+++ b/tests/unit/mixins/callback-route-mixin-test.js
@@ -1,0 +1,48 @@
+/* jshint expr:true */
+import Ember from 'ember';
+import { it } from 'ember-mocha';
+import { describe, beforeEach } from 'mocha';
+import { expect } from 'chai';
+import sinon from 'sinon';
+import CallbackRouteMixin from 'ember-simple-auth/mixins/callback-route-mixin';
+
+const { Object: EmberObject, RSVP, Route } = Ember;
+
+describe('CallbackRouteMixin', () => {
+  let route;
+  let session;
+  let ok;
+
+  describe('#activate', () => {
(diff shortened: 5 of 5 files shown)


marcoow:
Cool - CI is failing though.

I'll review after my vacation - sorry for the wait.

256dpi:
I fixed the tests and rebased on master. Enjoy your holidays!

addon/authenticators/oauth2-implicit-grant.js
```diff
@@ -0,0 +1,122 @@
...
+
+      if (response.error) {
+        return reject(response);
+      }
+
+      if (!this._validateToken(response)) {
+        return reject({
+          error: 'invalid_token',
+          // jscs:disable requireCamelCaseOrUpperCaseIdentifiers
+          error_description: 'The received token is invalid'
```
marcoow: Is this actually needed? I guess just rejecting with `'invalid_token'` should be sufficient, e.g.

```js
reject('invalid_token');
```

addon/authenticators/oauth2-implicit-grant.js
```diff
@@ -0,0 +1,122 @@
...
+          // jscs:enable requireCamelCaseOrUpperCaseIdentifiers
+        });
+      }
+
+      return resolve(response);
```
marcoow: no need to return here

addon/authenticators/oauth2-implicit-grant.js
```diff
@@ -0,0 +1,122 @@
...
+      }
+
+      if (!this._validateToken(response)) {
+        return reject({
+          error: 'invalid_token',
+          // jscs:disable requireCamelCaseOrUpperCaseIdentifiers
+          error_description: 'The received token is invalid'
+          // jscs:enable requireCamelCaseOrUpperCaseIdentifiers
+        });
+      }
```
marcoow: We should replace the `return`s here with `if`/`else if`, e.g.

```js
if (response.error) {
  reject(response);
} else if (!this._validateToken(response)) {
  …
  reject('invalid_token');
} else {
  resolve(response);
}
```

addon/authenticators/oauth2-implicit-grant.js
```diff
@@ -0,0 +1,122 @@
...
+   This method simply returns a resolving promise.
+
+   @method invalidate
+   @return {Ember.RSVP.Promise} A promise that when it resolves results in the session being invalidated
+   @public
+   */
+  invalidate() {
+    return new RSVP.Promise((resolve) => {
+      return resolve();
+    });
```
marcoow:
```js
return RSVP.Promise.resolve();
```

addon/authenticators/oauth2-implicit-grant.js
```diff
@@ -0,0 +1,122 @@
...
+  },
+
+  _validateToken(token) {
```
marcoow: this should maybe be renamed to `_validateResponse(response)` or so

addon/mixins/callback-route-mixin.js
```diff
@@ -0,0 +1,69 @@
...
+   @method activate
+   @public
+   */
+  activate() {
+    let authenticator = this.get('authenticator');
+
+    this.get('session').authenticate(authenticator).then(() => {
+      if (window.opener) {
+        return window.close();
+      }
```
marcoow: This won't result in any notification to the opener, meaning the opener would never know what the outcome of the authentication operation was. Implementing this in a proper way would require us to rebuild large parts of torii which I don't think is sth. we want to do.

I'd leave this out and leave it to people using this mixin to override the `activate` method if they want to have that behavior.

addon/mixins/callback-route-mixin.js
```diff
@@ -0,0 +1,69 @@
...
+
+      return this.transitionTo(this.get('routeOnSuccess'));
+    }).catch((err) => {
+      return this.transitionTo(this.get('routeOnError'), { queryParams: err });
+    });
```
marcoow: This should not actually transition. If authentication succeeds, ESA will already transition to the `routeAfterAuthentication`. In the case that authentication fails we should maybe set an error property instead or so.

tests/unit/authenticators/oauth2-implicit-grant-test.js
```diff
@@ -0,0 +1,121 @@
...
+    tryInvoke(server, 'shutdown');
+  });
+
+  describe('#restore', () => {
+    describe('when the data contains an access_token', () => {
+      it('resolves with the correct data', (done) => {
+        authenticator.restore(accessToken).then((_data) => {
+          expect(_data).to.eql(accessToken);
+          done();
+        });
```
marcoow: When testing promise based behavior you can return the promise from the test instead of using `done`, e.g.:

```js
describe('when the data does not contain an access_token', () => {
  it('resolves with the correct data', () => {
    return authenticator.restore(accessToken).then((_data) => {
      expect(_data).to.eql(accessToken);
    });
  });
```

marcoow:
This looks great! Just left a few comments. I'd probably not include this in the 1.2 release as that's already quite big but if everything goes well with that we should be able to include it in 1.3.

This should also be listed in the [_"Authenticators"_ section in the README](https://github.com/simplabs/ember-simple-auth#authenticators).

addon/authenticators/oauth2-implicit-grant.js
256dpi: An authentication server should always respond with an error object, that might directly be displayed to the user. Rejecting with just `'invalid_token'` would require the user to check for that single case.

An other option would be to just return the error identifier and require the user to bring a human readable text that can be displayed.

addon/authenticators/oauth2-implicit-grant.js
256dpi: The functions is also used during `restore` where there is no actual response involved. As it only checks if the passed object is a valid token I can't think of a better name.

addon/mixins/callback-route-mixin.js
256dpi: I'll remove the unnecessary success transition, but in most cases the user wants to transition back to the route that initiated the authentication process. This is the use case this mixin should cover. I'll make the error transition optional and also set the error in a property to cover both cases.

256dpi:
Ok I added some more commits based on your suggestions. I'll squash them when we're finished with the review.

addon/mixins/callback-route-mixin.js
marcoow: I'd say let's leave both transitions out for now. If anyone wants to have a transition in the error case they could always override the method, e.g.

```js
activate() {
  this._super().catch(() => {
    …
  });
}
```

addon/authenticators/oauth2-implicit-grant.js
```diff
@@ -0,0 +1,118 @@
...
+
+      if (response.error) {
+        reject(response);
+      } else if (!this._validateToken(response)) {
+        reject({
+          error: 'invalid_token',
+          // jscs:disable requireCamelCaseOrUpperCaseIdentifiers
+          error_description: 'The received token is invalid'
+          // jscs:enable requireCamelCaseOrUpperCaseIdentifiers
+        });
```
marcoow: I think this should just be

```js
reject('invalid_token');
```

addon/authenticators/oauth2-implicit-grant.js
```diff
@@ -0,0 +1,118 @@
...
+
+   @method invalidate
+   @return {Ember.RSVP.Promise} A promise that when it resolves results in the session being invalidated
+   @public
+   */
+  invalidate() {
+    return RSVP.Promise.resolve();
+  },
+
+  _validateToken(token) {
```
marcoow: As this validates a data set and not just the token I think `_validateData(data)` might be a better name?

addon/authenticators/oauth2-implicit-grant.js
```diff
@@ -0,0 +1,118 @@
...
+
+    // parse eventual error description
+    // jscs:disable requireCamelCaseOrUpperCaseIdentifiers
+    if (params.error_description) {
+      params.error_description = params.error_description.replace(/\+/g, ' ');
+    }
+    // jscs:enable requireCamelCaseOrUpperCaseIdentifiers
+
+    return params;
+  }
```
marcoow: I'm not sure whether the query param processing should be moved to the route instead where we could also use Ember's query param's functionality instead of having to process ourselves. What do you think?

addon/mixins/callback-route-mixin.js
```diff
@@ -0,0 +1,67 @@
...
+   If the routeOnError property is set it will transition to the set route with
+   the error specified as query parameters.
+
+   @method activate
+   @public
+   */
+  activate() {
+    let authenticator = this.get('authenticator');
+
+    this.get('session').authenticate(authenticator).catch((err) => {
```
marcoow: This should return the promise so it can be handled in an overridden method.

addon/authenticators/oauth2-implicit-grant.js
256dpi: Would this mean we also just return the error code from the response?

See: https://tools.ietf.org/html/rfc6749#section-5.2.

addon/authenticators/oauth2-implicit-grant.js
256dpi: To be secure authentication servers must append the token after the fragment e.g. `/callback#access_token=1234`. In that sense we cannot rely on Embers abstractions here.

See: https://tools.ietf.org/html/rfc6749#section-4.2.2.

I'd rather suggest to keep it in the authenticator so that the route callback mixin is not mandatory for using the authenticator.

addon/authenticators/oauth2-implicit-grant.js
marcoow: oh right, that would make sense - yes

256dpi:
Ok I made the discussed changes and squashed the commits. Any chance to see this 🚢 with 1.2? I have some projects I would really like to migrate to this solution. 😄

addon/authenticators/oauth2-implicit-grant.js
```diff
@@ -0,0 +1,106 @@
...
+   @method authenticate
+   @param {Object} response The response
+   @return {Ember.RSVP.Promise} A promise that when it resolves results in the session becoming authenticated
+   @public
+   */
+  authenticate(response) {
+    return new RSVP.Promise((resolve, reject) => {
+      if (isEmpty(response)) {
+        response = this._parseResponse(window.location.hash);
+      }
```
marcoow: Sorry - just realizing this now: in what case would `authenticate` be called with a `response`? Wouldn't we always want the authenticator to process the location hash?

I also think we should pass in the location hash when calling authenticate:

```js
this.get('session').authenticate('autenticator:oauth2-implicit-grant', window.location.hash);
```

marcoow: Also we have to check whether `window` is actually defined so this would not throw an error when running in FastBoot.

256dpi: In most cases the authenticator is run within the window that received the access token via the the hash from the authentication server. However, in some implementations (custom pop-up stuff maybe) an implementor wants to pass an already extracted hash (from another window) to `authenticate`. IMHO the authenticator should support that kind of customization.

Do you see an issue with that?

I agree on checking for the existence of window. A patch will follow.

marcoow: > In most cases the authenticator is run within the window that received the access token via the the hash from the authentication server. However, in some implementations (custom pop-up stuff maybe) an implementor wants to pass an already extracted hash (from another window) to authenticate. IMHO the authenticator should support that kind of customization.

Allowing that kind of flexibility is exactly the reason why I'd like to pass the hash in as an argument ;)

256dpi: Oh I see. You'd rather want to pass an actual hash string to authenticate (to be parsed) rather than an already parsed access token? Hmm, would it make sense to parse a string and directly use a passed object?

marcoow: Yes, I'd always pass the hash string as it appears in the URL the user is redirected to.

256dpi:
I updated the PR accordingly.

addon/authenticators/oauth2-implicit-grant.js
```diff
@@ -0,0 +1,108 @@
...
+   @method authenticate
+   @param {Object} hash The optional location hash
+   @return {Ember.RSVP.Promise} A promise that when it resolves results in the session becoming authenticated
+   @public
+   */
+  authenticate(hash) {
+    return new RSVP.Promise((resolve, reject) => {
+      if (isEmpty(hash) && !!window) {
+        hash = window.location.hash;
+      }
```
marcoow: I'd leave this out now and treat `hash` as a required argument.

addon/mixins/callback-route-mixin.js
```diff
@@ -0,0 +1,67 @@
...
+   @public
+   */
+  activate() {
+    let authenticator = this.get('authenticator');
+
+    return this.get('session').authenticate(authenticator).catch((err) => {
+      this.set('error', err);
+
+      if (!isEmpty(this.get('routeOnError'))) {
+        return this.transitionTo(this.get('routeOnError'), { queryParams: { error: err } });
```
marcoow: I'd really leave this out and leave it to implementors to redirect on errors by overriding the `activate` method:

```js
activate() {
  this._super(...arguments).catch(() => {
    return this.transitionTo(this.get('routeOnError'), { queryParams: { error: err } });
  });
});
```

256dpi: I find it a little ugly to use the Mixin (as a simplification) and then override the method each time which looks a little ugly and adds complexity again. If the mixin just calls the authenticator then it might be better to convert it to an example and let use call the authenticator in his own route.

Again, IMHO a mixin should provide functionality that is commonly used.

marcoow: The point is it's easy to add a transition in an override if you need it but it's harder to remove it if you don't need it.

I think what we want to achieve with the mixin is to encapsulate the code that **everybody** who uses this would have to use but leave everything else out. The problem is that once you start implementing configurable behavior there will be people demanding configuration options for other things as well and if you actually implement all these configuration options you end up with a huge and complex mess. The general strategy of ESA always has been to not add a lot of configurable behavior but make it easy to add it if you need it by overriding methods.

256dpi:
I updated the PR again. The callback route should now also not crash when used with Fastboot: https://github.com/simplabs/ember-simple-auth/pull/1057/files#diff-c499bd5af1dfdaca5e66b49844aede27R49.

However, the current implementation will totally break in conjunction with fastboot as the redirect by the authentication server would always trigger a fastboot server side render that has no access to `window.location.hash`.  From a quick scan through the docs I'm not aware of a way to disable server-side rendering for a particular route. FYI: A browser would never submit the fragment on a GET request. Do you have an idea?

What are you plans regarding fastboot suport?

marcoow:
Awesome, thanks for the effort 👍

Ember Simple Auth 1.2 will support FastBoot out of the box. That is mainly about transporting the session state between the client and the FastBoot server seamlessly. As the hash of a URL is purely client-side and will never be included in `GET` requests sent to the server, when rendering the callback route in FastBoot authentication would just always fail and lead to an error message being rendered or a transition to a different route being triggered. I'd say this is exactly what we want in this case as it's the _"correct"_ behavior actually.

256dpi:
Do you think it might be worth to discuss with the fastboot team a way to disable fastboot for a single URL and let it render on the client side? That way an app can fully leverage fastboot except for the auth callback part.

marcoow:
Hm, just realized FastBoot is actually a problem in this case ;)

When the user is being redirected back to the Ember app from the auth server with the access token in the hash, that request would always result in FastBoot rendering the route. However, trying to authenticate the user with FastBoot would always fail as it doesn't have access to the hash as you correctly said.

I guess the easiest option here would be to simply not execute the code that triggers the authentication in FastBoot but on the client side only.

marcoow:
> Do you think it might be worth to discuss with the fastboot team a way to disable fastboot for a single URL and let it render on the client side?

Do you want to bring this up in the #fastboot channel on the community slack?

256dpi:
I'm not a slack user, I'll raise a GH issue instead. 😄

256dpi:
The solution to the problem is to wrap the route callback logic with:

```js
if(!this.get('fastboot.isFastBoot')) {
  // this will only execute in the browser
}
```

That way the parsing of the access token is done in the browser when the Ember app is executed again after being pre-rendered in the server.

A proper implementation would require stuff that has ben already added in  #1035. Should I rebase the branch on the fastboot branch and implement the missing pieces?

marcoow:
We'll hopefully merge the `fastboot` branch soon. You can rebase against master and add the missing pieces then as this won't be release with 1.2 anyway.

essmd:
Hi guys! Any news or updates on when this PR will or can be merged?

marcoow:
We'll most likely release 1.2 final this week. After that this will be good to merge.

256dpi:
@marcoow Please ping me when you're ready.

addon/authenticators/oauth2-implicit-grant.js
```diff
@@ -0,0 +1,103 @@
...
+
+   @method restore
+   @param {Object} data The data to restore the session from
+   @return {Ember.RSVP.Promise} A promise that when it resolves results in the session becoming or remaining authenticated
+   @public
+   */
+  restore(data) {
+    return new RSVP.Promise((resolve, reject) => {
+      if (!this._validateData(data)) {
+        return reject();
```
marcoow: Can you modify this so it rejects with a proper error message (like https://github.com/simplabs/ember-simple-auth/blob/master/addon/authenticators/oauth2-password-grant.js#L239)?

addon/authenticators/oauth2-implicit-grant.js
```diff
@@ -0,0 +1,103 @@
...
+   @public
+   */
+  authenticate(hash) {
+    return new RSVP.Promise((resolve, reject) => {
+      let response = this._parseResponse(hash);
+
+      if (response.error) {
+        reject(response.error);
+      } else if (!this._validateData(response)) {
+        reject('invalid_token');
```
marcoow: https://github.com/simplabs/ember-simple-auth/blob/master/addon/authenticators/oauth2-password-grant.js#L239

addon/mixins/callback-route-mixin.js
```diff
@@ -0,0 +1,57 @@
...
+    // test if fastboot is available without requiring the fastboot service to be present
+    if (typeof FastBoot === 'undefined') {
+      hash = window.location.hash;
+    }
+
+    return this.get('session').authenticate(authenticator, hash).catch((err) => {
+      this.set('error', err);
+    });
+  }
+});
```
marcoow: This mixin should maybe be renamed so it's a bit more descriptive, e.g. `OAuth2ImplicitGrantCallbackRouteMixin`.

dknutsen:
Hey guys, this is looking awesome. Any updates on when it will be merged? Excited to use it! 🎉

256dpi:
I'm very sorry, but I won't really have time in the next couple of months. Feel free to pick it up and finish it!

dknutsen:
@256dpi I think I could handle that if you could give me access to the repo at some point

256dpi:
@dknutsenAFAIK the cleanest would be to fork the project yorself, pull in my changes in your own branch and submit another pull request. @marcoow Would that be ok for your?

dknutsen:
@256dpi ah right, that works for me

marcoow:
Anything that works for you guys works for me 😄 Would be great to get this merged for 1.3!

jamesdixon:
Any update on this? 👍

marcoow:
@dknutsen: this is almost ready to merge except for a few not-yet-adressed review comments.

marcoow:
closing in favor of #1252

PR #1065, opened 2016-09-04
Add resolve() call to load currentUser docs

### LINKED ISSUE #1064: Problem with Current User docs and implementation
It appears that a vitally important `resolve()` is missing from the [current user documentation](https://github.com/simplabs/ember-simple-auth/blob/master/guides/managing-current-user.md).  It does appear in the [dummy code](https://github.com/simplabs/ember-simple-auth/blob/master/tests/dummy/app/services/session-account.js#L15);

Additionally it appears as if this solution cannot be used reliably due to its async nature.  Obviously the `beforeModel` hook is not async since Ember waits for all the Promises to resolve before continuing, but the `sessionAuthenticated` event _is_ async, so if it takes longer for the promise to resolve than it takes to render the next route, the transition will blow up.

Specifically what I'm doing is transitioning to `dashboard` after authentication.  In the `dashboard` route I have a `beforeModel` that checks to see if the logged in user has taken the tour yet, if they have, I show them the dashboard, if they haven't I `transitionTo` the tour.

Because the current user promise hasn't resolved, the service doesn't have a `user` and the transition blows up.

kylemellander:
Closes #1064

### DIFF
--- a/guides/managing-current-user.md
+++ b/guides/managing-current-user.md
@@ -91,6 +91,7 @@ export default Ember.Service.extend({
       if (!isEmpty(userId)) {
         return this.get('store').find('user', userId).then((user) => {
           this.set('user', user);
+          resolve();
         }, reject);
       } else {
         resolve();


guides/managing-current-user.md
```diff
@@ -91,6 +91,7 @@ export default Ember.Service.extend({
       if (!isEmpty(userId)) {
         return this.get('store').find('user', userId).then((user) => {
           this.set('user', user);
+          resolve();
```
marcoow: I think we should change the complete method to

```js
load() {
  let userId = this.get('session.data.authenticated.user_id');
   if (!isEmpty(userId)) {
       this.set('user', user);
     });
   } else {
     return Ember.RSVP.resolve();
  }
}
```

for simplicity.

marcoow:
closing this as it's essentially the same thing as #1176.

PR #1069, opened 2016-09-13
Add explanatory comment to application route

davewasmer:
The setup-session-restoration initializer needs the application route to be explicitly defined in the consuming app, so simple auth must ship with `app/routes/application.js` in case the consuming app hasn't defined their own. However, if the consuming app has another addon that similarly provides an application route via `app/routes/application.js`, simple auth may override that addon's file with it's own empty version, depending on the load order.

This is easily fixed by specifying an explicit load order (i.e. `after: 'ember-simple-auth'`). However, because the contents of simple auth's `app/routes/application.js` are completely generic, and Ember's build process merges this file into the app folder without retaining source information, it's difficult to know from the consuming app's perspective that simple-auth is the addon providing this generic file (and therefore the `after` option needs to specify "ember-simple-auth").

This adds an explanatory comment, hopefully making it a bit clearer in case anyone (_cough_ me five minutes ago _cough_) is trying to figure out where this empty application route file is coming from.

### DIFF
--- a/app/routes/application.js
+++ b/app/routes/application.js
@@ -1,3 +1,4 @@
 import Ember from 'ember';
 
+// Ensure the application route exists for ember-simple-auth's `setup-session-restoration` initializer
 export default Ember.Route.extend();


marcoow:
Thanks for that!

It should eventually be possible to get rid of this completely - we'd just have to figure out how we can get either the actual application route or the one that Ember will generate on the fly if none has been defined [here](https://github.com/simplabs/ember-simple-auth/blob/master/addon/instance-initializers/setup-session-restoration.js#L4).

PR #1070, opened 2016-09-14
Ensure correct root url handling

### LINKED ISSUE #1048: baseURL and rootURL problem
I recently updated ember to 2.7 and changed the baseURL to rootURL and added the baseURL to the simple auth config. Unfortunately ember simple auth goes into a login/logout loop although the session does authenticate successfully. When adding the baseURL: '/' back to environment.js, simple auth does exactly what it is supposed to. Would like to continue with Ember's recommended deprecation but this issue is causing us to stick with baseURL and ignore the warning. Any suggestions?

My main issue is that there are no assertions, no errors, no logs. Everything appears to work correctly to login until is automatically just logs out.

kylemellander:
- [x] Updates docs to show readonly even if its an empty string.
- [x] Read the default value for config.baseURL from rootURL first, then baseURL as a back.
- [x] Raises deprecation error when running config.get('baseURL').
~~ - [ ] Update tests in https://github.com/simplabs/ember-simple-auth/blob/master/tests/unit/configuration-test.js~~

Replaces #1051 

Closes #1048

### DIFF
--- a/docs/theme/partials/props.handlebars
+++ b/docs/theme/partials/props.handlebars
@@ -11,9 +11,9 @@
     {{#if access}}
       <span class="label label-access-{{access}} pull-right">{{access}}</span>
     {{/if}}
-    {{#if readOnly}}
+    {{#exists readonly}}
       <span class="label label-warning pull-right">readonly</span>
-    {{/if}}
+    {{/exists}}
     <h3 class="panel-title signature">
       <strong class="name">`{{name}}`</strong>: <code>{{#crossLink type}}{{/crossLink}}</code>
     </h3>
--- a/app/initializers/ember-simple-auth.js
+++ b/app/initializers/ember-simple-auth.js
@@ -8,7 +8,7 @@ export default {
   name:       'ember-simple-auth',
   initialize: function(registry) {
     const config   = ENV['ember-simple-auth'] || {};
-    config.baseURL = ENV.baseURL;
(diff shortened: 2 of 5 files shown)


docs/theme/partials/props.handlebars
```diff
@@ -11,7 +11,7 @@
     {{#if access}}
       <span class="label label-access-{{access}} pull-right">{{access}}</span>
     {{/if}}
-    {{#if readOnly}}
+    {{#if readOnly includeZero=true}}
```
marcoow: what's this?

addon/configuration.js
```diff
@@ -97,5 +97,17 @@ export default {
...
+  get(keyName) {
+    if (keyName === 'baseURL') {
+      deprecate(
+        `Ember Simple Auth: Reading the property 'baseURL' is deprecated. It will become private beginning version 2.0.0.`,
+        false,
+        { id: 'ember-simple-auth.configuration.baseURL', until: '2.0.0' }
+      );
+    }
+
+    return this._super(...arguments);
```
marcoow: We should actually make the property defined in line 41 a [deprecating alias](http://emberjs.com/api/classes/Ember.computed.html#method_deprecatingAlias) for a to-be-added rootURL property.

marcoow:
Awesome!

I left a few review notes and added a TODO above.

PR #1077, opened 2016-09-29
Should `resolve` after fetching user.

dhilipsiva:
Else the promise remains unresolved.

### DIFF
--- a/guides/managing-current-user.md
+++ b/guides/managing-current-user.md
@@ -94,6 +94,7 @@ export default Ember.Service.extend({
       if (!isEmpty(userId)) {
         return this.get('store').find('user', userId).then((user) => {
           this.set('user', user);
+          resolve();
         }, reject);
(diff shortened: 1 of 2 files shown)


guides/managing-current-user.md
```diff
@@ -94,6 +94,7 @@ export default Ember.Service.extend({
       if (!isEmpty(userId)) {
         return this.get('store').find('user', userId).then((user) => {
```
marcoow: this `return` can be removed as well actually

dhilipsiva: Done 👍

marcoow:
Can you remove the [`return` here](https://github.com/simplabs/ember-simple-auth/blob/master/tests/dummy/app/services/session-account.js#L13) as well?

dhilipsiva:
Done! 👍

PR #1081, opened 2016-10-11
Detect version of EmberData to conditionally define ajaxOptions

cibernox:
In Ember-data >= 2.7 there `ajaxOptions` is deprecated in favour of
several more fineGrained methods, one of them being `headersForRequest`.

Ember-Simple-Auth already uses `headersForRequest`, but also for backward
compatibility, continued to define `ajaxOptions` for order versions of ED,
which raises annoying deprecation warnings in newer versions.

This PR makes ESA to conditionally define ajaxOptions ONLY for versions of ED
that do not have the newer API

### DIFF
--- a/addon/mixins/data-adapter-mixin.js
+++ b/addon/mixins/data-adapter-mixin.js
@@ -1,6 +1,8 @@
 import Ember from 'ember';
+import RESTAdapter from 'ember-data/adapters/rest';
 
 const { inject: { service }, Mixin, assert, isPresent } = Ember;
+const RESTAdapterPrototype = RESTAdapter.proto();
 
 /**
   __This mixin can be used to make Ember Data adapters authorize all outgoing
@@ -30,7 +32,7 @@ const { inject: { service }, Mixin, assert, isPresent } = Ember;
   @public
 */
 
-export default Mixin.create({
+let config = {
   /**
     The session service.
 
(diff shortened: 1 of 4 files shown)


cibernox:
@marcoow Feature wise this is done. I'm not sure how to properly notify users that in EmberData >= 2.7 they should use `headersForRequest` instead. 
Docs? Release notes?

pangratz:
Since the `ds-improved-ajax` feature is still somewhat in flux and the [RFC](https://github.com/emberjs/rfcs/pull/171) needs to be flushed out first before the feature can be enabled, I would suggest not to mention the `headersForRequest` for now.

Also, maybe it makes sense to only add custom `headersForRequest` function only if the `rest`-adapter defines it (because only then it is actually needed)?

cibernox:
@pangratz If the RFC is in flux, why does ED already throw deprecations that point to those methods?

Is it possible to inspect the target in which a mixing is about to be applied to conditionally execute logic? What is the advantage, not running a noop method on each request?

pangratz:
> If the RFC is in flux, why does ED already throw deprecations that point to those methods?

Yep, that has been an error on our side 😔 . Somehow that feature got into ember-data without going through an RFC process first, and more unfortunately it has been enabled in the beta cycle. The assertion has been added after we realized that many apps would have custom `ajax` and `ajaxOptions` and the new feature would suddenly break their apps. To address that we added the deprecation warning but the overwritten behavior of `ajax` and `ajaxOptions` is kept.

cibernox:
In that case, should we freeze this PR until the RFC is merged? 

Really the only important bit for this feature is the `headersForRequest` method. If the level of confidence that this specific part of the RFC is stable, even if there is discussions about other parts like the `bodyForRequest/dataForRequest`, this could go ahead.

marcoow:
I'd say we just ignore the deprecation for now until ED has decided on the future of `ds-improved-ajax`.

cibernox:
Fair enough. Let's label this PR as blocked and unblock it once the RFC gets merged.

marcoow:
The ED RFC has been closed - closing this as well. Seems like we'll have to do something different if we now if we want to tackle this again.

PR #1104, opened 2016-11-23
Cookie store `_secureCookies`: use correct protocol property in FastBoot mode

lolmaus:
According to [the guide](https://ember-fastboot.com/docs/user-guide), the `host` property only contains the host. For protocol, the `protocol` property should be used and it contains protocol name without the colon.

This edit also prevents a crash if the property is nully for [some reason](https://github.com/robwebdev/ember-cli-staticboot).

### DIFF
--- a/addon/session-stores/cookie.js
+++ b/addon/session-stores/cookie.js
@@ -108,10 +108,10 @@ export default BaseStore.extend({
 
   _secureCookies: computed(function() {
     if (this.get('_fastboot.isFastBoot')) {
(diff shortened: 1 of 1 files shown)


marcoow:
Yep, this seems good - just the tests fail. I think you'll have to adapt the mock so it includes the correct data.

lolmaus:
I've updated my PR, tests are green locally.

One of CI jobs failed randomly, please restart.

PR #1106, opened 2016-11-25
Fix OAuth2 authenticator tokenRefreshOffset

sagaban:
Just a minor fix on how a random number is generated.

### DIFF
--- a/addon/authenticators/oauth2-password-grant.js
+++ b/addon/authenticators/oauth2-password-grant.js
@@ -113,7 +113,7 @@ export default BaseAuthenticator.extend({
     const min = 5;
     const max = 10;
 
-    return (Math.floor(Math.random() * (max - min)) + min ) * 1000;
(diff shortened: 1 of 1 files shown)


marcoow:
Hm, was this actually broken before? Also this change breaks the build.

sagaban:
Sorry, it was a rookie mistake

marcoow:
I'm still not sure this was actually broken before - can you elaborate what this fixes?

sagaban:
Sure: As I understand the method, `tokenRefreshOffset` should return a value between the min and the max defined values. Am I right? With the previous code it was not that way. It was working properly just because `min` was equal to `max - min`.

marcoow:
hm: 
bildschirmfoto 2016-11-25 um 12 56 02

sagaban:
Yes, with 10 and 5 will work, but with different values won't.
image

PR #1116, opened 2016-12-06
WIP: Add a DeviseTokenAuth authenticator/authorizer

olleolleolle:
This WIP PR adds a new Authenticator and Authorizer for use with the Ruby gem [devise_token_auth](https://github.com/lynndylanhurley/devise_token_auth).

It is an extraction from working code [at eCraft](https://github.com/ecraft/).

## Checklist

- [x] Write API documentation, introducing devise_token_auth
- [ ] Write tests
- [x] DeviseAuthenticator - accommodate #1118
- [ ] Accommodate the Mixin, presented as a new file `addon/mixins/devise-token-auth-data-adapter-mixin.js` here
- [ ] ajaxOptions to be compatible with ember-network/fetch

## Questions

- [ ] Can we deprecate the Devise authenticator, authorizer? The proposed backend patch which comes linked in  the DeviseAuthorizer API docs is from 2013.
- [ ] devise_token_auth gem has a setting for using Legacy-style Devise (non-Bearer Token authentication) [README link to section where this is explained](https://github.com/lynndylanhurley/devise_token_auth/blob/master/README.md#configuration-cont) - I should investigate that to see if that could be a way to get the Devise parts EmberSimpleAuth to _have_ a backend done

### DIFF
--- a/addon/authenticators/devise-token-auth.js
+++ b/addon/authenticators/devise-token-auth.js
@@ -1,7 +1,7 @@
 import Ember from 'ember';
 import DeviseAuthenticator from './devise';
 
-const { RSVP, assert, getProperties, Array: { isEvery }, isPresent, run } = Ember;
+const { RSVP: { Promise }, isEmpty, run } = Ember;
 
 /**
   Authenticator that works with the Ruby gem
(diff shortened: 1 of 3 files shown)


ValentinTrinque:
Hi, I was planning to add the support for DTA to this package. Since you already started it, what about co-operating on this integration ?

So, where should we start / continue ?

ValentinTrinque:
@olleolleolle Ping ?

olleolleolle:
@ValentinTrinque If you want to "investigate that to see if that could be a way to get the Devise parts EmberSimpleAuth to have a backend done" to see whether it's possible to _not_ do all of this?

ValentinTrinque:
@olleolleolle So what you want me to do is, to see if it's possible to integrate devise_token_auth using Devise legacy support ?

olleolleolle:
@ValentinTrinque The value would be: create a documented workaround for those that want to keep using Devise.

DeviseTokenAuth users still need something like this authenticator/authorizer.

It also needs tests - investigate what the other authenticator/authorizer adapters have.

ValentinTrinque:
@olleolleolle Ok so, you mean integrate Devise-legacy through DTA instead of the current Devise implementation. And then continue the integration of DTA and its different adapters right ?

henrixapp:
The travis build is failing, with no visible error message. Can you explain this?

olleolleolle:
The test failures are about `cookieExpirationTime`, perhaps I've missed setting something.

marcoow:
closing - lots left to be done but no progress. Thanks for the effort though, @olleolleolle!

PR #1123, opened 2016-12-08
YUIDoc configuration and npm script command

Artmann:
Ember-simple-auth has a theme and helpers for YUIDoc but it's not added as a dependency nor does it contain the correct config for YUIDoc. This PR fixes the mentioned issues.

- Add YUIDoc as a dev-dependency
- Update the YUIDoc config
- Add a NPM command to run `ember-cli-yuidoc`

ping @olleolleolle

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,7 +12,8 @@
     "start": "ember server",
     "test": "ember try:each",
     "nodetest": "node node-tests/nodetest-runner.js",
-    "fastboot-nodetest": "node node-tests/fastboot-runner.js"
+    "fastboot-nodetest": "node node-tests/fastboot-runner.js",
+    "yuidoc": "ember ember-cli-yuidoc"
   },
   "repository": "https://github.com/simplabs/ember-simple-auth",
   "engines": {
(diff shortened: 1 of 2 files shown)


olleolleolle:
Proposed title of PR: "YUIDoc configuration and npm script command"

marcoow:
Currently documentation is built as part of the production build already: https://github.com/simplabs/ember-simple-auth/blob/master/ember-cli-build.js#L48.

Having a dedicated command would certainly be nice but we'll have the version is right etc.

olleolleolle:
@marcoow The use-case for having a "non-production build invocation for YUIDoc" that would build the same docs is "I want to know I'm fixing the docs right".

(Perhaps these non-production docs builds would land output in a different directory, so that the `../docs` output location outside current working directory does not trip up new or casual users.)

marcoow:
It would actually be important that both doc build tasks result in the same output so they'd have to use the same config. At that point I wouldn't be so sure anymore what the value of the second task is anymore as the only difference would be the invocation then really (`ember build -e production` vs. `npm run yuidoc`).

marcoow:
Thanks for your effort but I'm closing this as `ember build -e production` already works well and having to maintain 2 tasks that actually do the same thing only adds complexity while not actually providing additional value.

Issue #1122, opened 2016-12-08
Deprecation messages ember-simple-auth@1.2.0-beta.1 with ember@2.9.1


frunjik:
Installing version 1.2.0-beta.1 gives me the following deprecation messages:

ember.debug.js:6628DEPRECATION: When calling `Ember.deprecate` you must provide `until` in options. [deprecation id: ember-debug.deprecate-until-missing] See http://emberjs.com/deprecations/v2.x/#toc_ember-debug-function-options for more details.
        at logDeprecationStackTrace (http://localhost:4200/assets/vendor.js:17143:19)
        at HANDLERS.(anonymous function) (http://localhost:4200/assets/vendor.js:17253:7)
        at raiseOnDeprecation (http://localhost:4200/assets/vendor.js:17173:12)
        at HANDLERS.(anonymous function) (http://localhost:4200/assets/vendor.js:17253:7)
        at invoke (http://localhost:4200/assets/vendor.js:17269:7)
        at deprecate (http://localhost:4200/assets/vendor.js:17237:32)
        at deprecate (http://localhost:4200/assets/vendor.js:17230:7)
        at Object.deprecate (http://localhost:4200/assets/vendor.js:29444:37)
        at Module.callback (http://localhost:4200/assets/vendor.js:93874:21)

ember.debug.js:6628DEPRECATION: ember-getowner-polyfill is now a true polyfill. Use Ember.getOwner directly instead of importing from ember-getowner-polyfill [deprecation id: ember-getowner-polyfill.import]
        at logDeprecationStackTrace (http://localhost:4200/assets/vendor.js:17143:19)
        at HANDLERS.(anonymous function) (http://localhost:4200/assets/vendor.js:17253:7)
        at raiseOnDeprecation (http://localhost:4200/assets/vendor.js:17173:12)
        at HANDLERS.(anonymous function) (http://localhost:4200/assets/vendor.js:17253:7)
        at invoke (http://localhost:4200/assets/vendor.js:17269:7)
        at deprecate (http://localhost:4200/assets/vendor.js:17237:32)
        at Object.deprecate (http://localhost:4200/assets/vendor.js:29444:37)
        at Module.callback (http://localhost:4200/assets/vendor.js:93874:21)
        at Module.exports (http://localhost:4200/assets/vendor.js:132:32)

ember.debug.js:6797DEBUG: Ember             : 2.9.1
ember.debug.js:6797DEBUG: Ember Data        : 2.10.0
[...]

PR #1124, opened 2016-12-09
fix deprecated import of getOwner

sly7-7:
Should fix the second deprecation in #1122

Not sure if using setOwner this way in the tests is the right way though.
At least for the SessionService, maybe using the [setupTest thing](https://github.com/emberjs/ember-mocha#setup-tests) from is the way to go, but I guess it implies a bigger refactor

I just realize that it would be breaking change...

### DIFF
--- a/package.json
+++ b/package.json
@@ -78,7 +78,7 @@
     "ember-cli-import-polyfill": "^0.2.0",
     "ember-cli-is-package-missing": "^1.0.0",
     "ember-cookies": "^0.0.10",
-    "ember-getowner-polyfill": "^1.0.1",
+    "ember-getowner-polyfill": "^1.1.0",
     "ember-network": "^0.3.0",
     "silent-error": "^1.0.0"
   },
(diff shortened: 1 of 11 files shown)


package.json
```diff
@@ -78,7 +78,6 @@
     "ember-cli-import-polyfill": "^0.2.0",
     "ember-cli-is-package-missing": "^1.0.0",
     "ember-cookies": "^0.0.10",
-    "ember-getowner-polyfill": "^1.0.1",
```
marcoow: This is still needed but in a later version

tests/unit/internal-session-test.js
```diff
@@ -7,19 +7,20 @@ import InternalSession from 'ember-simple-auth/internal-session';
 import EphemeralStore from 'ember-simple-auth/session-stores/ephemeral';
 import Authenticator from 'ember-simple-auth/authenticators/base';
 
-const { RSVP, K, run: { next } } = Ember;
+const { RSVP, K, run: { next }, setOwner } = Ember;
 
-describe('InternalSession', () => {
+describe('InternalSession', function() {
```
marcoow: using arrow functions should be fine here unless you want to use `this`

package.json
sly7-7: I removed that because of `ember-getowner-polyfill is not required for Ember 2.3.0 and later, please remove from your `package.json`` at build time, but I think if you want to keep ember-simple-auth compatible with ember 1.12+ this dependency is required, right ?

marcoow: Yes, we'll want to maintain the current compatibility unless not possible anymore.

marcoow:
🎉

PR #1131, opened 2016-12-13
replace (run.cancel, run.later) with run.debounce

sly7-7:
I wonder if this could also be applied [here](https://github.com/simplabs/ember-simple-auth/blob/97a78c08244bdf6b99e458868e5529ae52508ab5/addon/authenticators/oauth2-password-grant.js#L350)

### DIFF
--- a/addon/session-stores/cookie.js
+++ b/addon/session-stores/cookie.js
@@ -2,7 +2,7 @@ import Ember from 'ember';
 import BaseStore from './base';
 import objectsAreEqual from '../utils/objects-are-equal';
 
-const { RSVP, computed, inject: { service }, run: { next, cancel, later, scheduleOnce }, isEmpty, typeOf, testing, isPresent, K, A, getOwner } = Ember;
+const { RSVP, computed, inject: { service }, run: { next, debounce, scheduleOnce }, isEmpty, typeOf, testing, isPresent, K, A, getOwner } = Ember;
 
 const persistingProperty = function(beforeSet = K) {
   return computed({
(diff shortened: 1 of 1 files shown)


marcoow:
🎉 awesome!

teddyzeenny:
The debounce is leaking after the session object has been destroyed. So when you destroy and recreate the container (via an app reset for example) the destroyed session store will keep syncing data and there's no way to break the loop. It will also throw errors if it detects changes because it will attempt to set the `content` property on a destroyed object (the destroyed internal session).  This bug actually existed before this PR but the PR made it impossible to break the loop in a subclass as the debounced tasks are no longer tracked.

As a result, I believe we need to keep track of these timers (as we were before this PR) and `Ember.run.cancel` them in a `willDestroy` hook. However, it seems it's currently [not possible to cancel a debounce created by another debounce](https://github.com/ebryn/backburner.js/issues/183) so going back to `run.later` and calling `run.cancel` when destroying may be the most straightforward approach. Thoughts?

marcoow:
Ouch, you're right @teddyzeenny - this should be fixed. Can you create an issue so we can track the problem there and move the discussion there? If there's no way to cancel the `debounce`s now anymore I think you're right and we need to go back to `later`/`cancel`.

sly7-7:
Sorry for the bad idea😓

Le 2 janv. 2017 12:50, "Marco Otte-Witte" <notifications@github.com> a
écrit :

Ouch, you're right @teddyzeenny <https://github.com/teddyzeenny> - this
should be fixed. Can you create an issue so we can track the problem there
and move the discussion there? If there's no way to cancel the debounces
now anymore I think you're right and we need to go back to later/cancel.

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/simplabs/ember-simple-auth/pull/1131#issuecomment-269962750>,
or mute the thread

marcoow:
@sly7-7: no, thanks for your contribution! We'll get it figured out!

teddyzeenny:
@sly7-7 not a bad idea at all. Your PR merely uncovered an already existing issue.

Issue #1127, opened 2016-12-09
[Proposal] Store accessed route before redirection to authentication route while using Fastboot


juanazam:
Hi,

First of all, thanks for this great addon and the efforts done to make it work with fastboot. Most of the things I had in place, worked without any changes except the `attemptedTransition` functionality on the `authenticated` and `unauthenticated` mixins.

Storing the transition in the session works perfectly on regular ember apps, but it doesn't work while running on fastboot mode.

In order to do so I had to implement support for that explicitly. I think it would be great if the addon could handle both cases out of the box without too much work.

In order to make it work with both cases, I overrode those mixins and added a conditional:

```
export default Mixin.create(AuthenticatedRouteMixin, {
  fastboot: service(),
  routing: service("-routing"),
 
  beforeModel(transition) {
    if (!get(this, 'session.isAuthenticated')) {
      assert('The route configured as Configuration.authenticationRoute cannot implement the AuthenticatedRouteMixin mixin as that leads to an infinite transitioning loop!', this.get('routeName') !== Configuration.authenticationRoute);

      set(this, 'session.attemptedTransition', transition);

      if (get(this, "fastboot.isFastBoot")) {
        /*global FastBoot*/
        let querystring = FastBoot.require('querystring');
        let url = querystring.escape(get(this, "fastboot.request.path"));

        return this.transitionTo(Configuration.authenticationRoute, { queryParams: { redirectUrl: url } });
      } else {
        return this.transitionTo(Configuration.authenticationRoute);
      }
    } else {
      return this._super(...arguments);
    }
  }
});
```

What do you think?

marcoow:
Yeah this is a good point. I think supporting that might be a bit trickier though. When redirecting in FastBoot, you could get the target ember route from the `transition` and e.g. submit it back to the client in a response header. In order to use that transition on the client then, a new transition object would have to be created (e.g. in an initializer) but I'm not sure how that object would look like. Also I think it's probably not public API…

juanazam:
Building a transition again might be a bit too tricky, that's why opted out for using a query param in the redirection for that particular case, so when the app gets on the auth route on fastboot (and then after loading the ember app), there is a simple way to retrieve that path. I know is not exactly the same workflow as for the regular ember app case, but it serves the same purpose.

As far as I have played around with transitions, they are not very friendly, for example building a path out one is more difficult than it should (IMHO).

marcoow:
[...]

PR #1136, opened 2016-12-14
Store attemptedTransition target url as a cookie in fastboot mode

DrewCarlson:
Fixes https://github.com/simplabs/ember-simple-auth/issues/1127

### DIFF
--- a/addon/mixins/application-route-mixin.js
+++ b/addon/mixins/application-route-mixin.js
@@ -92,18 +92,26 @@ export default Mixin.create({
     event. If there is a transition that was previously intercepted by
     {{#crossLink "AuthenticatedRouteMixin/beforeModel:method"}}the
     AuthenticatedRouteMixin's `beforeModel` method{{/crossLink}} it will retry
-    it. If there is no such transition, this action transitions to the
+    it. If there is no such transition, the `ember_simple_auth-redirectTarget`
+    cookie will be checked for a url that represents an attemptedTransition
+    that was aborted in Fastboot mode, otherwise this action transitions to the
     {{#crossLink "Configuration/routeAfterAuthentication:property"}}{{/crossLink}}.
+
 
     @method sessionAuthenticated
     @public
   */
   sessionAuthenticated() {
     const attemptedTransition = this.get('session.attemptedTransition');
+    const cookies = getOwner(this).lookup('service:cookies');
+    const redirectTarget = cookies.read('ember_simple_auth-redirectTarget');
 
     if (attemptedTransition) {
       attemptedTransition.retry();
       this.set('session.attemptedTransition', null);
+    } else if (redirectTarget) {
+      this.transitionTo(redirectTarget);
+      cookies.clear('ember_simple_auth-redirectTarget');
     } else {
       this.transitionTo(this.get('routeAfterAuthentication'));
     }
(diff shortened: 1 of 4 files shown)


addon/mixins/application-route-mixin.js
```diff
@@ -58,6 +58,10 @@ export default Mixin.create({
     return fastboot ? fastboot.get('isFastBoot') : false;
   }),
 
+  _cookies: computed(function() {
+    return getOwner(this).lookup('service:cookies');
+  }),
```
marcoow: We should lookup the service inside of the `sessionAuthenticated` method so that we don't add any unnecessary properties via the mixin.

addon/mixins/authenticated-route-mixin.js
```diff
@@ -33,12 +33,25 @@ export default Mixin.create({
   */
   session: service('session'),
 
+  _cookies: computed(function() {
+    return getOwner(this).lookup('service:cookies');
+  }),
```
marcoow: We should lookup the service inside of the sessionAuthenticated method so that we don't add any unnecessary properties via the mixin.

addon/mixins/authenticated-route-mixin.js
```diff
@@ -33,12 +33,25 @@ export default Mixin.create({
...
   }),
 
+  _secureCookies: computed(function() {
+    if (this.get('_isFastBoot')) {
+      const fastboot = getOwner(this).lookup('service:fastboot');
+      return fastboot.get('request.protocol') === 'https';
+    }
+
+    return window.location.protocol === 'https:';
+  }).volatile(),
```
marcoow: We should move this into the `sessionAuthenticated` method so that we don't add any unnecessary properties via the mixin. Also this will only ever be used when running in FastBoot so `window.location.protocol` can be ignored.

addon/mixins/authenticated-route-mixin.js
```diff
@@ -78,7 +94,12 @@ export default Mixin.create({
       let authenticationRoute = this.get('authenticationRoute');
       assert('The route configured as Configuration.authenticationRoute cannot implement the AuthenticatedRouteMixin mixin as that leads to an infinite transitioning loop!', this.get('routeName') !== authenticationRoute);
 
-      if (!this.get('_isFastBoot')) {
+      if (this.get('_isFastBoot')) {
+        this.get('_cookies').write('ember_simple_auth-redirectTarget', transition.intent.url, {
+          path: `/${authenticationRoute}`,
```
marcoow: `authenticationRoute` will be the Ember route and would contain dots instead of slashes (e.g. `app.internal.login` instead of `app/internal/login`). Is there any way we can get the URL path for the Ember route? If not I think we'll have to set the path to `/`.

tests/unit/mixins/application-route-mixin-test.js
```diff
@@ -12,12 +12,19 @@ const { Route, run: { next } } = Ember;
 describe('ApplicationRouteMixin', () => {
   let session;
   let route;
+  let cookiesMock;
 
   beforeEach(() => {
     session = InternalSession.create({ store: EphemeralStore.create() });
+    cookiesMock = {
+      read() {},
+      write() {},
```
marcoow: `write` shouldn't actually be needed here.

addon/mixins/application-route-mixin.js
DrewCarlson: I had `_cookies` and `_secureCookies` as members strictly for mocking in tests.  Is there a reference in the test suite for mocking the lookup function?

marcoow: Yeah, that makes testing easier for sure. There should be examples in the code base where the container is mocked though.

marcoow:
Yeah, it's probably fine to simply set the path to `/`.

marcoow:
awesome!

PR #1139, opened 2016-12-15
Improve current-user service guides

juanazam:
This PR aims to improve `current-user` service guides a bit, including the case when using a dedicated endpoint but including the actual current using id in the response instead of a special value (e.g.: me).

The newly added section explains why using EmberData's`find` method with id `me` and returning an actual id in the response is a problem. For more information, please check: https://github.com/emberjs/data/issues/4414.

It also closes a stalled issue and PR:
https://github.com/simplabs/ember-simple-auth/issues/1064
https://github.com/simplabs/ember-simple-auth/pull/1065

### DIFF
--- a/guides/managing-current-user.md
+++ b/guides/managing-current-user.md
@@ -50,7 +50,14 @@ export default Component.extend({
 
 In this example, the service does not need to know the ID of the current user
 as it uses a dedicated endpoint instead that will always respond with the user
-belonging to the authorization token in the request:
+belonging to the authorization token in the request.
+
+Note: Using `store.queryRecord` is the correct way to query for a record when id is
+unknown. Ember data expects the returned model to have the same id, as otherwise an
+unused empty record with `id` of `me` is in the store.
+
+We can override the adapter to generate `api/users/me` when `store.queryRecord` is
+invoked with a query param where the `me` param is present.
 
 ```js
 // app/services/current-user.js
@@ -64,10 +71,24 @@ export default Ember.Service.extend({
 
   load() {
     if (this.get('session.isAuthenticated')) {
-      return this.get('store').find('user', 'me').then((user) => {
+      return this.get('store').queryRecord('user', { me: true }).then((user) => {
         this.set('user', user);
       });
     }
+  }
+});
+
+// app/adapters/user.js
+import ApplicationAdapter from './application';
+
+export default ApplicationAdapter.extend({
+  urlForQueryRecord(query) {
+    if (query.me) {
+      delete query.me;
+      return `${this._super(...arguments)}/me`;
+    }
+
+    return this._super(...arguments);
   }
 });
 ```
@@ -89,17 +110,14 @@ export default Ember.Service.extend({
   store: service(),
 
   load() {
-    return new RSVP.Promise((resolve, reject) => {
-      let userId = this.get('session.data.authenticated.user_id');
-      if (!isEmpty(userId)) {
-        this.get('store').find('user', userId).then((user) => {
-          this.set('user', user);
-          resolve();
-        }, reject);
-      } else {
-        resolve();
-      }
-    });
+    let userId = this.get('session.data.authenticated.user_id');
+    if (!isEmpty(userId)) {
+      return this.get('store').findRecord('user', userId).then((user) => {
+        this.set('user', user);
+      });
+    } else {
+      return Ember.RSVP.resolve();
+    }
   }
 });
 ```


marcoow:
@pangratz: can you review?

guides/managing-current-user.md
```diff
@@ -72,6 +72,50 @@ export default Ember.Service.extend({
...
+});
+```
+
+If the `find` method is called with id `me` and the response includes the actual
+user id, ember data will create an empty record with id `me` and all other model
+attributes empty.
+
+Using the `queryRecord` method and overriding the adapter avoid this problem.
+The adapter override will generate `api/users/me` when the `me` query param is
+present.
```
pangratz: Let's change the ending to something like this:

... when `store.queryRecord()` is invoked with a query where the `me` param is present.

guides/managing-current-user.md
```diff
@@ -89,17 +133,14 @@ export default Ember.Service.extend({
...
-          this.set('user', user);
-          resolve();
-        }, reject);
-      } else {
-        resolve();
-      }
-    });
+    let userId = this.get('session.data.authenticated.user_id');
+    if (!isEmpty(userId)) {
+      return this.get('store').find('user', userId).then((user) => {
```
pangratz: `store.findRecord` is the public way.

guides/managing-current-user.md
```diff
@@ -72,6 +72,50 @@ export default Ember.Service.extend({
...
+    }
+
+    return this._super(...arguments);
+  }
+});
+```
+
+If the `find` method is called with id `me` and the response includes the actual
+user id, ember data will create an empty record with id `me` and all other model
+attributes empty.
```
pangratz: `find` is actually private, I think we should use `store.findRecord` here as it is more idiomatic ember-data use, even though it has been mentioned before.

I am not sure we should start here why `findRecord('user', 'me')` is the wrong choice and instead say why we are using `store.queryRecord()`: because it's the way to fetch a record in ember data when the `id` is not known beforehand.

And then maybe outline with a `Note: ` why we are not using `findRecord("user", "me")`: because ember data expects the returned model to have the same id, as otherwise an unused empty record with `id` of `me` is in the store.

pangratz:
@juanazam overall this looks great! Thanks so much for tackling this!

guides/managing-current-user.md
juanazam: Thanks for the feedback, I'll reword it :)

guides/managing-current-user.md
juanazam: 👍

juanazam:
@pangratz thanks for reviewing and commenting! I have updated my changes with your suggestions, hope it looks ok now.

guides/managing-current-user.md
```diff
@@ -64,14 +64,57 @@ export default Ember.Service.extend({
 
   load() {
     if (this.get('session.isAuthenticated')) {
-      return this.get('store').find('user', 'me').then((user) => {
+      return this.get('store').findRecord('user', 'me').then((user) => {
```
marcoow: We should probably have `queryRecord` and make what is described in _"Using a dedicated endpoint and returning actual user id in response"_ below the default recommendation as `findRecord` will lead to problems and thus shouldn't be in the guides at all.

guides/managing-current-user.md
```diff
@@ -64,14 +64,57 @@ export default Ember.Service.extend({
...
+  }
+});
+```
+
+Note: Using `store.queryRecord` is the correct whey to query for a record when id is
+unknown. Ember data expects the returned model to have the same id, as otherwise an
+unused empty record with `id` of `me` is in the store.
+
+The adapter override will generate `api/users/me` when `store.queryRecord` is
+invoked with a query param where the `me` param is present.
```
marcoow: We should put the explanation before the code sample so the guide is easier to follow when reading through it top-down.

guides/managing-current-user.md
juanazam: So, if I understood correctly, we should remove the first section right?

marcoow: Yeah, replace this sample with the one using `queryRecord` you added.

juanazam:
@marcoow thanks for your suggestions, I have included those changes too.

juanazam:
🎉

guides/managing-current-user.md
```diff
@@ -89,17 +110,14 @@ export default Ember.Service.extend({
...
-          resolve();
-        }, reject);
-      } else {
-        resolve();
-      }
-    });
+    let userId = this.get('session.data.authenticated.user_id');
+    if (!isEmpty(userId)) {
+      return this.get('store').findRecord('user', userId).then((user) => {
+        this.set('user', user);
```
mcfiredrill: Shouldn't there be another `      return Ember.RSVP.resolve();` here?

Issue #1128, opened 2016-12-09
Not redirecting to `routeAfterAuthentication` after logging in


anthonycollini:
Hello,

When I log in for the first time, ember simple auth properly redirects to the route set in the`routeAfterAuthentication` setting. However, if I log out, and then back in, it redirects me to the `index` route, instead of obeying this setting.

Any thoughts as to why this is happening?

marcoow:
Can't reproduce this in the dummy app. Can you provide a bit more info:

* ESA version
* your config

texpert:
The redirection isn't working with ESA 1.2.0-beta-1 with the same configuration that is working with ESA 1.1.0.

Nothing fancy in config:

```js
/* jshint node: true */

module.exports = function(environment) {
  let ENV = {
    modulePrefix: 'tickets-front',
    environment: environment,
    rootURL: '/',
    locationType: 'auto',
    EmberENV: {
      FEATURES: {
        // Here you can enable experimental features on an ember canary build
        // e.g. 'with-controller': true
      },
      EXTEND_PROTOTYPES: {
        // Prevent Ember Data from overriding Date.parse.
        Date: false
      }
    },
    contentSecurityPolicy: {
      'connect-src': "*"
    },

    apiHost: 'http://localhost:3000',
    namespace: 'api',

    APP: {
      // Here you can pass flags/options to your application instance
      // when it is created
    }
  };

  ENV['ember-simple-auth'] = {
    routeAfterAuthentication: 'dashboard',
    routeIfAlreadyAuthenticated: 'dashboard'
  };

  if (environment === 'development') {
    ENV.APP.LOG_RESOLVER = true;
    ENV.APP.LOG_ACTIVE_GENERATION = true;
    ENV.APP.LOG_TRANSITIONS = true;
    ENV.APP.LOG_TRANSITIONS_INTERNAL = true;
    ENV.APP.LOG_VIEW_LOOKUPS = true;
  }

  if (environment === 'test') {
    // Testem prefers this...
    ENV.locationType = 'none';

    // keep test console output quieter
    ENV.APP.LOG_ACTIVE_GENERATION = false;
    ENV.APP.LOG_VIEW_LOOKUPS = false;

    ENV.APP.rootElement = '#ember-testing';
  }

  if (environment === 'production') {

  }

  return ENV;
};
```

marcoow:
[...]

PR #1140, opened 2016-12-15
devise authenticator now also processing the JSON API data.

texpert:


marcoow:
Thanks for your effort but I don't think JSON API should be supported with the devise adapter. JSON API is a resource based format and I don't think is a great fit for authentication. This could certainly live on in an addon to ember-simple-auth of course that people could use that actually require JSON API for auth requests on the backend.

In that case I'd suggest to extend the `Devise` authenticator though so you don't have to keep your version of the base authenticator up to date in the addon.

marcoow:
also see my comment on the topic here: https://github.com/simplabs/ember-simple-auth/issues/1128#issuecomment-266970372

texpert:
As you wish, of course, @marcoow , but just to mention that a common enough use case for authorizing a resource based API is using Devise authenticator + Doorkeeper authorizer.

marcoow:
Sure but I think what you should do is use resource-focussed endpoints and JSON API for managing resources and sth. else for authentication/unauthentication requests. In the end, there is no resource you're creating when authenticating - there's no model neither on the client or server side etc.

Both JSON API and Devise/Doorkeeper can coexist in an app perfectly fine - I would just not use JSON API for the auth requests.

bravo-kernel:
For future reference by people looking into this... in all honesty using JSON API for authentication is perfectly fine considering one is basically creating a new (jwt) token (resource) when authenticating. Add to that the benefits of DRY use of shared application logic for validation, errors, etc and it is actually a pretty sane choice.

However... as much as I stand for the above I gave up and made the exception as well, no JSON API for auth ;)

PR #1142, opened 2016-12-18
ESA and torii GitHub guide

srvance:
I've written a guide that takes you step by step through the process of authorizing your app to use the GitHub APIs. This is a complete first draft.

I plan to add a diagram of the OAuth2 explicit grant flow, but the guide is large enough that I wanted to give you reading time.

Note that the "Useful Links" section at the end contains a link to a repo on my account with the source that follows the guide. If you want to move that under the project or into a simplabs repo, I'm open for that.


guides/auth-torii-with-github.md
```diff
@@ -0,0 +1,581 @@
...
+application` button. Fill in the fields similar to the following.
+
+![Registering an OAuth Application with GitHub](./assets/github-registering-oauth-application.png)
+
+You will likely want to register different application keys for different deployment stages
+because of the differences in the `Authorization callback URL`.
+
+Once you register your application, you will have a chance to add an application logo.
+Most importantly, you will be given a client ID and a client secret. _The client secret is
+a password and should be protected as such!_ You will use the client ID in your web
```
marcoow: Not sure whether the term _"password"_ might cause confusion here. Maybe we can say something like _"The client secret must be kept secret and must never be included e.g. in the Ember application's source"_ or so?

guides/auth-torii-with-github.md
```diff
@@ -0,0 +1,581 @@
...
+{{else}}
+    Unauthenticated
+{{/if}}
+
+<hr />
+
+{{outlet}}
+```
+
+The `session` service is automatically injected, so there's no code for that. We'll add the `config` entry for now to
```
marcoow: Auto-injection might actually be removed with ESA 2.0. We should show how to inject the service in the `application` controller here.

guides/auth-torii-with-github.md
```diff
@@ -0,0 +1,581 @@
...
+import Ember from 'ember';
+import AuthenticatedRouteMixin from 'ember-simple-auth/mixins/authenticated-route-mixin';
+
+export default Ember.Route.extend(AuthenticatedRouteMixin, {
+});
+```
+
+If you're still running the app when you save this, you will see it redirect to the `login` route.
+
+We'll also designate the `login` route as available for unauthenticated access by applying the
```
marcoow: This should maybe read _"for unauthenticated access __only__"_ as that's really the point of the mixin - making sure authenticated users can __not__ access it.

guides/auth-torii-with-github.md
```diff
@@ -0,0 +1,581 @@
...
+    }
+  }
+```
+
+Finally, change your `login` template to send the action when the button is pressed.
+
+```handlebars
+{{!-- app/templates/login.hbs --}}
+
+<button {{action "login"}}>Log in to GitHub</button>
```
marcoow: We should use modern action in this guide, helping make people familiar with modern/idiomatic Ember concepts, so this would be:

```hbs
<button onclick={{action "login"}}>Log in to GitHub</button>
```

and the above implementation of the action would go to the controller.

guides/auth-torii-with-github.md
```diff
@@ -0,0 +1,581 @@
...
+test it as we go.
+
+First, we'll add a logout button to our `application` template right after the text "Authenticated". This seems a
+reasonable place since it serves as our application's rudimentary menu bar and should only show when the user is
+authenticated.
+
+```handlebars
+{{!-- add to app/templates/application.hbs --}}
+
+    <button {{action "logout"}}>Log Out</button>
```
marcoow: See above - we should be using modern closure actions:

```hbs
<button onclick={{action "logout"}}>Log Out</button>
```

guides/auth-torii-with-github.md
```diff
@@ -0,0 +1,581 @@
...
+    }
+  }
+```
+
+Clicking the "Log Out" button will take you back to the login state.
+
+### Obtaining an Access Token
+
+The final step in the process is to exchange your authorization code for an API access token. This can only be done
+securely through a back end service because it needs to know your application password. Right now, our app looks like
```
marcoow: I think we should use the correct term _"Client Secret"_ here instead of _"application password"_.

guides/auth-torii-with-github.md
```diff
@@ -0,0 +1,581 @@
...
+            }
+            console.log(`Successful response: ${body}`);
+            done(null, body);
+        });
+    });
+    req.on('error', callback);
+    req.write(data);
+    req.end();
+};
+```
```
marcoow: It might make sense to show how to do this with Ember CLI's mock server for the development environment as well.

guides/auth-torii-with-github.md
srvance: Should we change it now or when that change takes place? Why remove injection?

guides/auth-torii-with-github.md
srvance: Do you want the emphasis on "only" in the guide or did you include it to point the difference out to me?

guides/auth-torii-with-github.md
srvance: Done.

guides/auth-torii-with-github.md
srvance: Done.

guides/auth-torii-with-github.md
srvance: Agreed.

guides/auth-torii-with-github.md
srvance: Done. The changes are committed to the demo app, but we should re-create the history if we want to officially use it. I structured the app so each commit corresponds to the section of the guide.

guides/auth-torii-with-github.md
marcoow: I would change it now already so that people get used to explicitly inject the service instead of relying on the auto-injection. The advantage of explicit injection (`session: Ember.inject.service`) is simply that it's easier to follow and less magic.

guides/auth-torii-with-github.md
```diff
@@ -0,0 +1,605 @@
...
+and for use against GitHub using `ember-simple-auth` and `torii` including a registered
+GitHub application, an unauthenticated login route, an authenticated content route, a
+token exchange service, and use of an authorizer to inject the token into
+requests. We'll use the [ember-data-github](https://github.com/elwayman02/ember-data-github)
+addon so we don't need to create our own GitHub models, serializers, and adapters.
+
+While most of the pieces are documented in the `ember-simple-auth` and `torii`
+documentation, putting them together properly for the particular scenario has not been
+fully documented to date. Additionally, many of the sources on the general internet are
+helpful but dated or incomplete.
```
marcoow: I'm not sure it's necessary to explain the motivation for this guide here as with the release of this the problem that's being described doesn't exist anymore so the paragraph isn't actually relevant anymore ;)

guides/auth-torii-with-github.md
```diff
@@ -0,0 +1,605 @@
+[Back to Main README](../README.md)
+
+# Full GitHub Authorization with the `torii-provider`
+
+This guide walks through the process of authenticating and authorizing an Ember app with
+and for use against GitHub using `ember-simple-auth` and `torii` including a registered
+GitHub application, an unauthenticated login route, an authenticated content route, a
+token exchange service, and use of an authorizer to inject the token into
+requests. We'll use the [ember-data-github](https://github.com/elwayman02/ember-data-github)
+addon so we don't need to create our own GitHub models, serializers, and adapters.
```
marcoow: I think it would be good to briefly explain that the general concepts are the same for any other OAuth 2.0 provider (Facebook, Google etc.). Otherwise people who try to use one of those providers might feel that this wouldn't help them while it actually would. Maybe we could even rename the whole guide to _"Implementing the OAuth 2.0 Authorization Code Grant Type with Ember Simple Auth and torii"_ or so and simply explain in the guide that github is only used as an example.

guides/auth-torii-with-github.md
```diff
@@ -0,0 +1,605 @@
...
+```
+
+To keep our client secret secret, we'll create a `settings.js` file in the `/server` directory looking like
+
+```js
+module.exports = {
+  CLIENT_ID: '<YOUR CLIENT ID>',
+  CLIENT_SECRET: '<YOUR CLIENT SECRET>',
+  USER_AGENT: '<YOUR APPLICATION NAME>'
+};
```
marcoow: Couldn't these be read from the `.env` file as well? With this you end up duplicating these settings.

guides/auth-torii-with-github.md
```diff
@@ -0,0 +1,605 @@
...
+`.env` and the following to the development section of your `environment.js`. __Note that you should use `https` for the
+token exchange service in production!__
+
+```js
+// add to development section of config/environment.js
+
+ENV.torii.providers['github-oauth2'].tokenExchangeUri = process.env.DEV_TOKEN_EXCHANGE_URL;
+```
+
+Now that we've set up a token exchange service, let's use it. Replace your `torii` adapter with
```
marcoow: > Replace your `torii` adapter

This should be _"Replace your `torii` __authenticator__"_

guides/auth-torii-with-github.md
```diff
@@ -0,0 +1,605 @@
...
+          access_token: JSON.parse(response).access_token,
+          provider: data.provider
+        };
+      });
+    });
+  }
+});
+```
+
+This makes a `POST` request to your service and gives the response back to Torii. Notice that it returns the `access_token`
```
marcoow: > gives the response back to Torii

That's not really what happens. Instead it gets the authorization code from torii (via `_super`) and exchanges that for an access token via the API service. torii is actually done and never called again when the promise returned from `_super` resolves.

guides/auth-torii-with-github.md
srvance: Makes sense. I had some confusion about where it was auto-injected and where it wasn't. I've gone back to the initializers a couple of times to refresh that memory.

guides/auth-torii-with-github.md
srvance: Removed.

guides/auth-torii-with-github.md
srvance: They could, but I see the server as a separate application and wanted to keep them in that scope. The only duplicated setting is the `CLIENT_ID`. The Ember app should NOT have access to the CLIENT_SECRET as a security precaution.

guides/auth-torii-with-github.md
srvance: Done.

guides/auth-torii-with-github.md
srvance: Changed "Torii" to "Ember Simple Auth"

guides/auth-torii-with-github.md
srvance: Done.

guides/auth-torii-with-github.md
marcoow: Hm, I see what you mean but you would actually want to keep these settings in env vars also when running the server in production (think of when you're running it in a container) so the general concept of env vars makes a lot of sense for the server part in particular. Maybe it would work to use 2 separate env vars?

srvance:
That's the way I do it in my Lambda. I'll see if I can use a `.env` later today. With the `settings` approach you can still keep credentials out of source with the `.gitignore` instructions (my biggest problem in the demo app) but it's not a production quality solution. 

Stephen Vance

srvance:
The server only uses the `.env` from the project root. You can't add your own at the server root with the current functionality of `ember-cli-dotenv`. I can leave `settings` as is or move them to `.env` with additional explanation about what a production configuration should look like. Do you have a preference or another option you like better?

marcoow:
I'd prefer keeping everything in `.env` actually as that's going to result in more production-like code for the server where you'd get these values from env vars as well.

srvance:
Done!

marcoow:
awsome! 🎉

srvance:
🎉 
What are your thoughts about the demo app?

marcoow:
I'm fine with the demo app staying under your account - credit where credit is due!

PR #1148, opened 2016-12-20
Re-export authorizers and authenticators

srvance:
Create app re-exports of the authorizers and authenticators so that they can be used directly without extension. Base versions were omitted because they should not be used directly. Session stores were omitted because there should only be one, which is already implemented.

There will need to be documentation changes to communicate this effectively. I will wait until the other pending documentation PRs are merged and you indicate that this is a change you want before doing that.

### DIFF
--- a/app/authenticators/devise.js
+++ b/app/authenticators/devise.js
@@ -1,0 +1,1 @@
+export { default } from 'ember-simple-auth/authenticators/devise';
--- a/app/authenticators/oauth2-password-grant.js
+++ b/app/authenticators/oauth2-password-grant.js
@@ -1,0 +1,1 @@
+export { default } from 'ember-simple-auth/authenticators/oauth2-password-grant';
--- a/app/authenticators/test.js
+++ b/app/authenticators/test.js
@@ -1,0 +1,1 @@
+export { default } from 'ember-simple-auth/authenticators/test';
--- a/app/authenticators/torii.js
+++ b/app/authenticators/torii.js
@@ -1,0 +1,1 @@
+export { default } from 'ember-simple-auth/authenticators/torii';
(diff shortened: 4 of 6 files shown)


marcoow:
In many cases you'd only use one of these authenticators and one authorizer. It's also very unlikely that you'll be using the OAuth 2.0 authenticator __and__ the Devise authenticator. This is why they aren't merged into the app automatically as that would most often only add unused things to the container. This is basically the same approach that Ember Data takes with adapters and serializers that are never merged into the app directly.

srvance:
According to @runspired, both app and addon are merged and packaged in until we get tree shaking, so there's no additional cost and some additional convenience and code reduction.

marcoow:
Sure, this is more about the cognitive weight that the presence of unused authenticators adds vs. the actual weight in terms of file size.

marcoow:
Also this change would make later tree shaking much harder or impossible as the app would actually use all of the modules then and the tree shaker would have to include them in the app bundle then although the app might not actually use any of the functionality.

srvance:
Your call. Not sure I understand the cognitive weight since they're hidden in the addon. I also assumed tree shaking would remove all unused, not just those in addon, but I haven't tracked the details of it closely.

marcoow:
The tree-shaker could not just remove modules that are in the app folder but never get imported anywhere as everything that's in the `app` folder could be retrieved via the container as e.g. `authenticator:devise`. Not sure how that will work when added to Ember CLI but my assumption that nothing in the `app` folder will ever be shaken by the tree shaker.

The point about the cognitive weight is mostly that you'll see all of these authenticators in the app's source (especially when looking at the source in e.g. Developer Tool's _"Source"_ tab) and I think that might confuse people as they'd e.g. see a devise authenticator when they only do OAuth 2.0 with a Node backend for example.

srvance:
Thanks for your patience in talking this through.

marcoow:
Sure, thanks for your effort!

PR #1155, opened 2016-12-31
upgrade version that shows in Ember inspector info tab

jedrula:
This was likely missed when creating 1.2.0-beta.1 release.


marcoow:
ouch, thanks

PR #1160, opened 2017-01-02
Add docs and warning for `cookieExpirationTime`

### LINKED ISSUE #1134: Cookie store session always expires after cookie Expiration Time even with user activities
I'm using ember simple auth using cookie store and session always expires after cookie Expiration Time even with user activities. Is this is the default behavior ? Shouldn't we renew session expiration if user is not idle 

Ember	2.9.1
Ember Data	2.10.0
Ember Simple Auth	1.1.0

stevenwu:
Fix #1134


marcoow:
Awesome! Test would be great and then :shipit:

marcoow:
We also need to investigate those random Travis errors…

PR #1164, opened 2017-01-04
Remove setOwner() calls

Turbo87:
`setOwner()` is not polyfilled by ember-getowner-polyfill and fails on older Ember versions. Using `setOwner()` in the tests was introduced by https://github.com/simplabs/ember-simple-auth/pull/1124, which should have failed CI, but apparently went through for some unknown reason (possibly a caching issue or some transitive dependency constraint on the polyfill?).

This is needed until https://github.com/rwjblue/ember-getowner-polyfill/pull/14 lands.

### DIFF
--- a/tests/helpers/create-with-container.js
+++ b/tests/helpers/create-with-container.js
@@ -1,0 +1,14 @@
+import Ember from 'ember';
+
+const { setOwner } = Ember;
+
+export default function createWithContainer(Constructor, options, container) {
+  if (setOwner) {
+    let instance = Constructor.create(options);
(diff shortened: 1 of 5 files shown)


marcoow:
@Turbo87 don't really understand why the tests would fail here…

Turbo87:
because passing in `container` apparently doesn't work anymore in newer Ember versions 💥 

got a workaround prepared already. will push soon.

Turbo87:
@marcoow updated the PR with the new solution

PR #1165, opened 2017-01-04
add ability to send client_id via authorization header or in POST body

jasonmacgowan:
The way that were sending a client_id currently (in an authorization header with an empty password) explicitly implies that we're doing client authentication.

The OAuth2 spec outlines that we can't do client authentication from a user-agent based client.
https://tools.ietf.org/html/rfc6749#page-16

This PR adds an attribute (enableClientAuth) that will send the client_id in the POST body instead of in the authorization header if it's set to false.

In my mind, the default should be false, but for now it's set to true for backwards compatibility.

### DIFF
--- a/addon/authenticators/oauth2-password-grant.js
+++ b/addon/authenticators/oauth2-password-grant.js
@@ -336,7 +336,7 @@ export default BaseAuthenticator.extend({
 
     const clientIdHeader = this.get('_clientIdHeader');
     if (this.get('enableLegacyClientIdentification') && !isEmpty(clientIdHeader)) {
-      deprecate('Ember Simple Auth: sending client_id as authoriation header is no longer ' +
+      deprecate('Ember Simple Auth: sending client_id in the Authorization header is no longer ' +
         'supported, please set enableLegacyClientIdentification property to false',
         false,
         { id: `ember-simple-auth.oauth2-password-grant.makeRequest`, until: '2.0.0' });
--- a/addon/authenticators/oauth2-password-grant.js
+++ b/addon/authenticators/oauth2-password-grant.js
@@ -338,7 +338,7 @@ export default BaseAuthenticator.extend({
     if (this.get('enableLegacyClientIdentification') && !isEmpty(clientIdHeader)) {
       deprecate('Ember Simple Auth: sending client_id as authoriation header is no longer ' +
         'supported, please set enableLegacyClientIdentification property to false',
-        false, 
+        false,
         { id: `ember-simple-auth.oauth2-password-grant.makeRequest`, until: '2.0.0' });
       merge(options.headers, clientIdHeader);
     }
--- a/addon/authenticators/oauth2-password-grant.js
+++ b/addon/authenticators/oauth2-password-grant.js
@@ -336,10 +336,12 @@ export default BaseAuthenticator.extend({
 
     const clientIdHeader = this.get('_clientIdHeader');
     if (this.get('enableLegacyClientIdentification') && !isEmpty(clientIdHeader)) {
-      deprecate('Ember Simple Auth: sending client_id in the Authorization header is no longer ' +
+      deprecate(
+        'Ember Simple Auth: sending client_id in the Authorization header is no longer ' +
         'supported, please set enableLegacyClientIdentification property to false',
         false,
-        { id: `ember-simple-auth.oauth2-password-grant.makeRequest`, until: '2.0.0' });
+        { id: `ember-simple-auth.oauth2-password-grant.makeRequest`, until: '2.0.0' }
+      );
       merge(options.headers, clientIdHeader);
     }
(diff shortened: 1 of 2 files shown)


thoptersword:
@marcoow my interpretation of section [2.3](https://tools.ietf.org/html/rfc6749#section-2.3) and [2.3.1](https://tools.ietf.org/html/rfc6749#section-2.3.1) is that we should only send HTTP Basic authentication if we have both a client_id and client_secret AND we're sending from a "confidential" client as defined in [2.1](https://tools.ietf.org/html/rfc6749#section-2.1)

marcoow:
Yeah, that seems right but I don't see where the spec defines that the client_id should be sent in the body if the client is non-confidential.

thoptersword:
@marcoow good point.  The OAuth2 authorization server that I'm using accepts it in the POST body but I don't see any guidance in the spec for what that standard would be

marcoow:
Maybe there's an informal standard for this in terms of the majority of libraries doing it a certain way? If we find out that's the case I'm happy to adapt to that.

thoptersword:
@marcoow I checked Auth0 and Stormpath and they both send client_id in POST body for implicit (password) grant type.  node-oauth2-server accepts the same.  Not sure what the best way to get a community consensus would be

marcoow:
Hm, the spec actually doesn't seem to say anything about it. An alternative would be sending the client_id in the query string though. I'm not sure which one is the better option.

We should also change the name of the `enableClientAuth` setting as this was never meant to be client __authentication__ really and using that name for the setting might lead people to assume it is (we had endless discussions here about why ESA doesn't support actual client auth incl. a client secret). I think something like `enableLegacyClientIdentification` which defaults to `true` and also triggers a deprecation when its value is `true` would be better.

marcoow:
Sorry for the delay. Reading the spec again I found it says

> The authorization server MAY establish a client authentication method with public clients.  However, the authorization server MUST NOT rely on public client authentication for the purpose of identifying the client.

and 

> client_secret
>   REQUIRED.  The client secret.  The client MAY omit the parameter if the client secret is an empty string.

and 

> Including the client credentials in the request-body using the two  is NOT RECOMMENDED and SHOULD be limited to clients unable to directly utilize the HTTP Basic authentication scheme (or other password-based HTTP authentication schemes)

marcoow:
I think this does actually make sense. As the spec says:

> The client MAY omit the parameter if the client secret is an empty string.

In the case of public clients the client secret is not an empty string though but actually doesn't exist at all so it's probably not really correct to still send the client id and (empty) client secret in the `Authorization` header.

Also this would allow us to drop the base64 dependency with ESA 2.0.

marcoow:
@jasonmacgowan: can you allow me to push to this PR so I can rebase?

marcoow:
nevermind - I have permissions already and was just using git wrong…

addon/authenticators/oauth2-password-grant.js
```diff
@@ -318,7 +335,11 @@ export default BaseAuthenticator.extend({
     };
 
     const clientIdHeader = this.get('_clientIdHeader');
-    if (!isEmpty(clientIdHeader)) {
+    if (this.get('enableLegacyClientIdentification') && !isEmpty(clientIdHeader)) {
+      deprecate('Ember Simple Auth: sending client_id as authoriation header is no longer ' +
```
Turbo87: typo

tests/unit/authenticators/oauth2-password-grant-test.js
```diff
@@ -127,6 +127,34 @@ describe('OAuth2PasswordGrantAuthenticator', () => {
       authenticator.authenticate('username', 'password');
     });
 
+    it('sends an AJAX request to the token endpoint without client_id Basic Auth header when enableLegacyClientIdentification is false', function(done) {
+      server.post('/token', (request) => {
+        expect(request.requestHeaders.hasOwnProperty('authorization')).to.eql(false);
```
Turbo87: `.to.be.false;`

tests/unit/authenticators/oauth2-password-grant-test.js
```diff
@@ -127,6 +127,34 @@ describe('OAuth2PasswordGrantAuthenticator', () => {
...
+          'password': 'password',
+          'client_id': 'test-client'
+        });
+        done();
+      });
+
+      authenticator.set('clientId', 'test-client');
+      authenticator.set('enableLegacyClientIdentification', false);
+      authenticator.authenticate('username', 'password');
+    });
```
Turbo87: I think we could combine the two tests since they are essentially doing the same thing but asserting on different parts of the request

marcoow: I'd prefer leaving assertions for separate aspects in separate tests.

marcoow:
this has been done in #1679

PR #1172, opened 2017-01-11
Removes unused ember-getowner-polyfill dependency

xcambar:
ember-getowner-polyfill is no longer used in ESA and triggers a deprecation warning in latest releases of Ember.

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,7 +22,6 @@
   },
   "dependencies": {
     "ember-cli-babel": "^5.1.7",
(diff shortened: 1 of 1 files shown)


marcoow:
ember-getowner-polyfill is actually still used - you just don't need to `import getOwner` anymore as it is now a proper polyfill that adds the method to the `Ember` global. The deprecation was resolved in #1124

xcambar:
Sorry I missed it. Thanks

marcoow:
Thanks for the effort!

PR #1176, opened 2017-01-22
Improved current-user and it's loading

mfazekas:
- `api/users/me` version of `currentUser` returns a promise even if session is invalid, so we can always call `get('currentUser').load().catch( () => this.get('session').invalidate() )`
- In `currentUser` loading we need to invalidate session if `currentUser` fails. The scenarios is that we had valid session so ESA did `restore` therefore `beforeModel` is called (and no `sessionAuthenticated`), but in this case it's possible that backend no longer considers the session valid, and it manifests as an error for `currentUser`, we need to invalidate session in this case too.

### DIFF
--- a/guides/managing-current-user.md
+++ b/guides/managing-current-user.md
@@ -74,6 +74,8 @@ export default Ember.Service.extend({
       return this.get('store').queryRecord('user', { me: true }).then((user) => {
         this.set('user', user);
       });
+    } else {
+      return Ember.RSVP.resolve();
     }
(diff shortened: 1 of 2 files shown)


marcoow:
This seems good. Can you update the [dummy app](https://github.com/simplabs/ember-simple-auth/blob/master/tests/dummy/app/routes/application.js) as well?

mfazekas:
@marcoow updated dummy app

PR #1182, opened 2017-01-26
Update "ember-cli" to v2.11.0

Turbo87:


.travis.yml
```diff
@@ -18,6 +18,7 @@ env:
   - EMBER_TRY_SCENARIO=ember-release
   - EMBER_TRY_SCENARIO=ember-beta
   - EMBER_TRY_SCENARIO=ember-canary
+  - EMBER_TRY_SCENARIO=ember-default
```
marcoow: I think we should actually migrate to using ember-try's `versionCompatibility` feature eventually.

Turbo87: I'm not sure if that's necessary for `ember-simple-auth`. Do we use any private APIs in this project that we need to make sure are available in all minor releases? In the other projects it makes sense since we use Handlebars AST transforms which are not public API yet.

marcoow: The main benefit I'm seeing is that we wouldn't nave to maintain this list ourselves anymore.

PR #1186, opened 2017-01-27
Use "eslint-config-simplabs"

Turbo87:
@marcoow please check if those changes are okay, otherwise we'll have to adjust the linting rules.


addon/configuration.js
```diff
@@ -3,9 +3,9 @@ import Ember from 'ember';
 const { getWithDefault, typeOf, deprecate } = Ember;
 
 const DEFAULTS = {
-  baseURL:                     '',
-  authenticationRoute:         'login',
-  routeAfterAuthentication:    'index',
+  baseURL: '',
+  authenticationRoute: 'login',
+  routeAfterAuthentication: 'index',
```
marcoow: I actually think aligned hashes improve readability a lot. Not aligning assignments is fine I guess but I'd keep it for hashes.

ember-cli-build.js
```diff
@@ -1,46 +1,47 @@
...
-  var app = new EmberAddon(defaults, {
+  let app = new EmberAddon(defaults, {
     storeConfigInMeta: true,
     jscsOptions: {
       enabled: true,
-      testGenerator: function(relativePath, errors) {
+      testGenerator(relativePath, errors) {
         if (errors) {
-          errors = "\\n" + this.escapeErrorString(errors);
+          // eslint-disable-next-line prefer-template
```
marcoow: Why not use a template string instead?

tests/unit/services/session-test.js
```diff
@@ -41,7 +41,7 @@ describe('SessionService', () => {
 
   it('forwards the "invalidationSucceeded" event from the session', (done) => {
     let triggered = false;
-    sessionService.one('invalidationSucceeded', () => triggered = true);
+    sessionService.one('invalidationSucceeded', () => { triggered = true; });
```
marcoow: Why are we enforcing braces here?

ember-cli-build.js
Turbo87: good question... actually all of that stuff can be removed anyways since we don't need JSCS anymore...

tests/unit/services/session-test.js
Turbo87: because you shouldn't `return` assignments, and the arrow function is implicitly doing that.

marcoow: I see - that makes sense

Turbo87:
@marcoow @pangratz updated the PR with a few adjustments

Turbo87:
@marcoow I'll release `eslint-config-simplabs`, update this PR and then merge it

PR #1190, opened 2017-01-29
Update dependencies

ValentinTrinque:


package.json
```diff
@@ -25,12 +25,11 @@
     "ember-cli-import-polyfill": "^0.2.0",
     "ember-cli-is-package-missing": "^1.0.0",
     "ember-cookies": "^0.0.11",
-    "ember-getowner-polyfill": "^1.1.0",
```
marcoow: This is actually needed

ValentinTrinque: Actually, since Ember 2.3, it isn't. But does this package ensure retro-compatibility ? I runned the tests (`ember test` am I right ?)  without experiencing any issues.

I can't spot where the `errored` state comes from on TravisCI. But I see that one of your scenario is testing against ember 1.12 isn't it ? This could be the origin.

marcoow: ESA supports Ember version starting from 1.12 so we need the polyfill for those versions.

ValentinTrinque: Ok I see. I will re-establish this package ASAP. Thanks for your quick answers.

Turbo87:
@ValentinTrinque could you rebase this on top of `master` to resolve the conflict?

Turbo87:
thanks @ValentinTrinque!

ValentinTrinque:
That's a pleasure, and I also greatly appreciate your reactivity @Turbo87 !

PR #1200, opened 2017-02-06
simplify import code

marcoow:
As we now use ember-cli-import-polyfill as a dependency we don't need the code in `index.js` anymore that deals with the dependency-of-a-dependency situation anymore.

### DIFF
--- a/index.js
+++ b/index.js
@@ -7,11 +7,6 @@ module.exports = {
   included(app) {
     this._super.included.apply(this, arguments);
 
-    // see: https://github.com/ember-cli/ember-cli/issues/3718
-    if (typeof app.import !== 'function' && app.app) {
-      app = app.app;
-    }
-
     app.import('vendor/ember-simple-auth/register-version.js');
   }
 };


Turbo87:
this thing is always confusing me due to the missing documentation... 😭 

https://github.com/ember-cli/ember-cli-import-polyfill/issues/1#issuecomment-237672090 seems to state that this polyfill is for apps, not addons, so I'm really not sure if we should use it here at all. The more common solution to this seems to be e.g. https://github.com/cibernox/ember-power-select/blob/a8abd8d78e23b56ef263f59ad0099a228a06b1f4/index.js#L4-L14

marcoow:
> Addon authors should be telling users to upgrade to ember-cli 2.7 or install the polyfill.

Hm, not sure I get that. Isn't including ember-cli-import-polyfill as an ember-simple-auth dependency _"telling users to install the polyfill"_? ;)

Turbo87:
> Isn't including ember-cli-import-polyfill as an ember-simple-auth dependency "telling users to install the polyfill"?

nope, unfortunately that's not how it works AFAIK 😞

PR #1196, opened 2017-02-05
Added possibility to define cookie path parameter

IgorKvasn:
In cookie store I am missing ability to define Cookie PATH parameter - this is especially useful when Ember app is being served from a different context path then '/'

PR #1201, opened 2017-02-06
Added possibility to define cookie path parameter #2

IgorKvasn:
replacement for a bad PR #1196

### DIFF
--- a/addon/session-stores/cookie.js
+++ b/addon/session-stores/cookie.js
@@ -100,6 +100,17 @@ export default BaseStore.extend({
   }),
 
   /**
+    The path to use for the cookie, e.g., "/", "/something".
+
+    @property cookiePath
+    @type String
+    @default '/'
+    @public
+  */
+  _cookiePath: '/',
+  cookiePath: persistingProperty(),
+
+  /**
     The expiration time for the cookie in seconds. A value of `null` will make
     the cookie a session cookie that expires and gets deleted when the browser
     is closed.
@@ -218,7 +229,7 @@ export default BaseStore.extend({
     let cookieOptions = {
       domain: this.get('cookieDomain'),
       expires: isEmpty(expiration) ? null : new Date(expiration),
-      path: '/',
+      path: this.get('cookiePath'),
       secure: this.get('_secureCookies')
     };
     if (this._oldCookieName) {
--- a/addon/session-stores/adaptive.js
+++ b/addon/session-stores/adaptive.js
@@ -80,6 +80,17 @@ export default Base.extend({
   cookieName: proxyToInternalStore(),
 
   /**
+    The path to use for the cookie, e.g., "/", "/something".
+
+    @property cookiePath
+    @type String
+    @default '/'
+    @public
+  */
+  _cookiePath: '/',
+  cookiePath: proxyToInternalStore(),
+
+  /**
     The expiration time for the cookie in seconds if `localStorage` is not
     available. A value of `null` will make the cookie a session cookie that
     expires and gets deleted when the browser is closed.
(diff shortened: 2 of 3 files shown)


addon/session-stores/cookie.js
```diff
@@ -148,6 +161,8 @@ export default BaseStore.extend({
   init() {
     this._super(...arguments);
 
+    this.set('_cookiePath', Configuration.baseURL);
```
marcoow: This should only be set when it is `null` (as it might have been set when the cookie store instance was created with `Cookie.create({ cookiePath: '/whatever' })` and shouldn't be overridden in this case).

tests/unit/session-stores/adaptive-test.js
```diff
@@ -6,13 +6,18 @@ import Adaptive from 'ember-simple-auth/session-stores/adaptive';
...
+import Configuration from 'ember-simple-auth/configuration';
 
 const { assign: emberAssign, merge, run } = Ember;
 const assign = emberAssign || merge;
 
 describe('AdaptiveStore', () => {
   let store;
 
+  beforeEach(function() {
+    Configuration.load({ baseURL: '/' });
```
marcoow: Why is this required?

tests/unit/session-stores/cookie-test.js
```diff
@@ -17,6 +18,7 @@ describe('CookieStore', () => {
 
   beforeEach(function() {
     store = createCookieStore(FakeCookieService.create());
+    Configuration.load({ baseURL: '/' });
```
marcoow: Why is this required?

addon/session-stores/cookie.js
IgorKvasn: tests in `cookie-store-behaviour.js` expect that cookie path is set to '/' e.g.
https://github.com/simplabs/ember-simple-auth/blob/master/tests/unit/session-stores/shared/cookie-store-behavior.js#L110

I guess there are 2 ways to sort this out - either change it in all asserts from `path === '/'` to `path === Configuration.baseURL` or just set a default baseURL for all tests (which I did)

tests/unit/session-stores/adaptive-test.js
IgorKvasn: tests in `cookie-store-behaviour.js` expect that cookie path is set to '/' e.g.
https://github.com/simplabs/ember-simple-auth/blob/master/tests/unit/session-stores/shared/cookie-store-behavior.js#L110

I guess there are 2 ways to sort this out - either change it in all asserts from `path === '/'` to `path === Configuration.baseURL` or just set a default baseURL for all tests (which I did)

marcoow: I see. I'd just set the `cookiePath` property on the created store instance instead ([here](https://github.com/simplabs/ember-simple-auth/blob/master/tests/unit/session-stores/shared/cookie-store-behavior.js#L98)) in these cases so the test is independent of the `Configuration`.

IgorKvasn: yes, that's probably a better option then to globally force this setting to all tests

IgorKvasn:
hmm there are some regression tests failing (at least that's how I understood Travis) - I guess to make it backwards compatible, cookie path should be  `null` by default (or to be precise - do not add any cookie path at all)

IgorKvasn:
nope, fastboot tests are failing

marcoow:
> I guess to make it backwards compatible, cookie path should be null by default (or to be precise - do not add any cookie path at all)

You are right - defaulting to `/` would actually be a breaking change:
https://tools.ietf.org/html/rfc6265#section-4.1.2.4

marcoow:
@IgorKvasn: can you rebase on `master`?

addon/session-stores/cookie.js
```diff
@@ -218,7 +229,7 @@ export default BaseStore.extend({
     let cookieOptions = {
       domain: this.get('cookieDomain'),
       expires: isEmpty(expiration) ? null : new Date(expiration),
-      path: '/',
+      path: this.get('cookiePath'),
```
marcoow: Hm, this is actually a breaking change. We'll need to default the property to `'/'` instead.

marcoow: I fixed it

PR #1224, opened 2017-02-20
Provide opt outs for async infinite sync loop

gfmartinez:
For context, I work with a dev team on http://dailyuv.com
Ember version: 2.8.3 - Fastboot enabled
Ember-simple-auth: 1.1.0 (we did update to 1.2.0 for testing)

We are using fastboot and have been running into many issues with its implementation and support in add-on libraries.

One issue we noticed in our application is that the ember-simple-auth cookie.js file (at https://github.com/simplabs/ember-simple-auth/blob/master/addon/session-stores/cookie.js#L257) creates an infinite loop. For our application it is unnecessary.

So, this Pull Request is meant to provide options for opting out of that infinite loop or to change its interval. Same with the expiration renewal interval. In our testing this does not affect our ability to maintain the current user or with any other functionality.

### DIFF
--- a/addon/session-stores/cookie.js
+++ b/addon/session-stores/cookie.js
@@ -73,6 +73,29 @@ export default BaseStore.extend({
   _renewExpirationTimeout: null,
 
   /**
+    The inverval for syncing data in milliseconds. A value of `null` will make
+    the data sync only once and not in a loop with the syncDataInverval
+
+    @property syncDataInterval
+    @type number
+    @default 500
+    @public
+  */
+  _syncDataInterval: 500,
+  syncDataInterval: persistingProperty(),
+
+  /**
+    The inverval for renewing cookie expirationt time in seconds.
+
+    @property renewExpirationTime
+    @type number
+    @default 60
+    @public
+  */
+  _renewExpirationTime: 60,
+  renewExpirationTime: persistingProperty(),
+
+  /**
     The domain to use for the cookie, e.g., "example.com", ".example.com"
     (which includes all subdomains) or "subdomain.example.com". If not
     explicitly set, the cookie domain defaults to the domain the session was
@@ -252,9 +275,9 @@ export default BaseStore.extend({
         this._lastData = data;
         this.trigger('sessionDataUpdated', data);
       }
-      if (!testing) {
+      if (!testing && this.syncDataInterval) {
         cancel(this._syncDataTimeout);
-        this._syncDataTimeout = later(this, this._syncData, 500);
+        this._syncDataTimeout = later(this, this._syncData, this.syncDataInterval);
       }
     });
   },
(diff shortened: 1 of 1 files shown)


marcoow:
Tab/window syncing of cookies should actually never happen in FastBoot - [this line](https://github.com/simplabs/ember-simple-auth/blob/master/addon/session-stores/cookie.js#L162) should actually prevent that already. The same goes for expiration renewal.

Could you trace how the `_syncData`/`_renewExpiration` methods ever get invoked on your server at all?

gfmartinez:
@marcoow Thanks for the reply! We have an application session-store that actually overrides simple-auth's `restore, persist, clear, _renew` methods in the case of fastboot. For each method we have something like:

`if(isFastBoot) {
      return RSVP.resolve();
    } else {
      return this._super();
    }`

So that we don't utilize those methods during fastboot loading.

Once our app loads and initializes, the init method in simple-auth is called again and the `_syncData` method is called from that same line that you provided above as part of the `init` call.

The `_syncData` cancel/later loop then continues and the `restore` method is called every 500ms infinitely.

marcoow:
> @marcoow Thanks for the reply! We have an application session-store that actually overrides simple-auth's restore, persist, clear, _renew methods in the case of fastboot.

I'd be interested in the motivation for that. Actually the cookie session store should work with FastBoot out of the box without a need to override anything.

> For each method we have something like:
> if(isFastBoot) { return RSVP.resolve(); } else { return this._super(); }
> So that we don't utilize those methods during fastboot loading.

I don't think this is what you want to be honest as it effectively disables 2-way session syncing.

> Once our app loads and initializes, the init method in simple-auth is called again and the _syncData method is called from that same line that you provided above as part of the init call.

That seems very strange. Can you check whether the `fastboot` service's `isFastBoot` property is actually `true` when running in FastBoot?

gfmartinez:
@marcoow We disabled session in fastboot intentionally because we don't want fastboot to have access to authenticated data. The reason is that we need to cache pages and we can't if they contain user dependent data.

I think the fastboot portion of this request is misleading. Apologies. Even if we cut out all of our overrides and don't use fastboot, just the stock configuration, we still see this infinite recalling of the `_syncData` (and thus the `restore` method).

According to the code at (https://github.com/gfmartinez/ember-simple-auth/blob/713d860aae443a2416f98784cd223edfe8e431bd/addon/session-stores/cookie.js#L272) which is called by the `init` at (https://github.com/gfmartinez/ember-simple-auth/blob/713d860aae443a2416f98784cd223edfe8e431bd/addon/session-stores/cookie.js#L187) there seems to be no way to cancel the loop once started. So that is why we have put forth this Pull Request.

We don't need the loop and it seems to affect performance.

marcoow:
To be clear: do you see the infinite loop in the FastBoot server or in the browser? Of course this is also an intentional infinite loop actually.

If you don't want session syncing across windows/tabs or expiration renewal you can of course use a custom store and override the `_syncData` and `_renewExpiration` methods in that.

gfmartinez:
@marcoow Thank you. We can certainly override the `_syncData` method.
We see the infinite loop in the browser.

Thank you for clearing up that the infinite loop is intentional.
It did strike me as strange that in a framework built on promises and event triggers that the syncing of data would rely on an infinite loop, which is why we approached this as a 'problem'.

When we override the `_syncData` method we see no change in the authentication protocol and no infinite loop. So that is what we will do.

Thanks again.

marcoow:
> It did strike me as strange that in a framework built on promises and event triggers that the syncing of data would rely on an infinite loop, which is why we approached this as a 'problem'.

We need to check for changes periodically as there are no events triggered when cookies change. There's really no other way to do this unfortunately.

Issue #1138, opened 2016-12-15
Session service not instantiated in Fastboot


sevab:
I have just upgraded to master and started integrating `ember-simple-auth` with fastboot, switching to Cookie session store from Adaptive as [instructed](https://github.com/simplabs/ember-simple-auth#fastboot). 

However the session's state doesn't seem to be instantiated on the server. While `ember_simple_auth-session` cookie is present with all the necessary data, the `session.isAuthenticated` still returns `false` and `session.data` is empty in fastboot mode:

```javascript
// /app/routes/application.js
import Ember from 'ember';
import ApplicationRouteMixin from 'ember-simple-auth/mixins/application-route-mixin';

export default Ember.Route.extend(ApplicationRouteMixin, {
  currentUser: Ember.inject.service(),
  fastboot: Ember.inject.service(),

  beforeModel() {
    // Cookie is present:
    Ember.Logger.log('cookies', this.get('fastboot.request.cookies'));
    // => { 'ember_simple_auth-session': '{"authenticated":{"authenticator":"authenticator:devise","token":"XhIJ...","email":"...@...com"}}' }

    // But session is not restored from cookie:
    Ember.Logger.log('session.data', this.get('session.data'));
    // => { authenticated: {} } {}
    Ember.Logger.log('session.isAuthenticated', this.get('session.isAuthenticated'));
    // => false

    return this._loadCurrentUser();
  },
  _loadCurrentUser() { return this.get('currentUser').load(); },
  sessionAuthenticated() {
    this._loadCurrentUser().then(() => {
      if (this.get('session.data.redirectUrl')) {
        var redirectUrl = this.get('session.data.redirectUrl');
        this.get('session').set('data.redirectUrl', null);
        window.location.replace(redirectUrl);
      }
    }).catch(() => this.get('session').invalidate() );
  },
});
```

On the frontend the same code works as expected returning true for `session.isAuthenticated` and restoring `session.data` from cookie.

Not sure if I'm doing something wrong, or if this is a bug.

sevab:
The issue seems to be gone now, possibly after reinstalling all npm packages.

Can be closed.

willviles:
I'm having this same problem with ESA 1.2.0. 

The cookie is being read correctly from the headers, but in my application.js `beforeModel`, `this.get('session.data') => { authenticated: {} }`. 

[...]

PR #1239, opened 2017-03-01
FastBoot dependency whitelisting

willviles:
#### Adds node-fetch to fastbootDependencies
~~This PR fixes Ember Simple Auth from breaking in Fastboot when using the OAuth2PasswordGrant authenticator, as referenced in [#1138](https://github.com/simplabs/ember-simple-auth/issues/1138#issuecomment-283318213). The fix is to simply add node-fetch to the addon's whitelisted `fastbootDependencies`.~~

```json
"ember-addon": {
  "configPath": "tests/dummy/config",
  "fastbootDependencies": {
    "node-fetch"
  }
}
```

See below for updated PR info.

### DIFF
--- a/README.md
+++ b/README.md
@@ -605,6 +605,21 @@ import CookieStore from 'ember-simple-auth/session-stores/cookie';
 export default CookieStore.extend();
 ```
 
+If you are using the
+[`OAuth2PasswordGrantAuthenticator`](http://ember-simple-auth.com/api/classes/OAuth2PasswordGrantAuthenticator.html),
+or
+[`DeviseAuthenticator`](http://ember-simple-auth.com/api/classes/DeviseAuthenticator.html),
+you must add `node-fetch` to your list of FastBoot whitelisted dependencies
+in `package.json`:
+
+```json
+{
+  "fastbootDependencies": [
+    "node-fetch"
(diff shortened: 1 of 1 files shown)


marcoow:
I think this won't work because of ember-fastboot/ember-cli-fastboot#188

willviles:
You're right. I've changed the PR to simply update the README to notify of the need to add `node-fetch` to the app's whitelisted `fastbootDependencies`.

README.md
```diff
@@ -605,6 +605,16 @@ import CookieStore from 'ember-simple-auth/session-stores/cookie';
 export default CookieStore.extend();
 ```
 
+If you are using the [`OAuth2PasswordGrantAuthenticator`](http://ember-simple-auth.com/api/classes/OAuth2PasswordGrantAuthenticator.html), you must add `node-fetch` to your list of FastBoot whitelisted dependencies in `package.json`:
```
marcoow: This actually applies to the Devise authenticator as well.

marcoow: Can you format the text so that it adheres to the 80 chars line limit for consistency with the rest of the README?

willviles: Updated! 👍

PR #1242, opened 2017-03-02
GK: Ignore ESLint plugin updates

Turbo87:
see https://github.com/simplabs/ember-simple-auth/pull/1223#issuecomment-283595471

### DIFF
--- a/package.json
+++ b/package.json
@@ -89,5 +89,11 @@
   },
   "ember-addon": {
     "configPath": "tests/dummy/config"
+  },
+  "greenkeeper": {
+    "ignore": [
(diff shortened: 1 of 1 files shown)


marcoow:
I guess we should probably ignore ember-cli updates as well (and all dependencies that are managed by ember-cli)?

Turbo87:
not necessarily. IMHO it's good that greenkeeper opens PRs for those since you can then just work on top of the PR and add the other necessary changes from the blueprints. Also there should usually be no reason to *not* update dependencies "managed by Ember CLI", it's only problematic if you don't update those deps but do update CLI itself.

PR #1247, opened 2017-03-08
don't return transitionTo from beforeModel

### LINKED ISSUE #1238: AuthenticatedRouteMixin
Hello,

On line 96 of the authenticated-route-mixin.js file, why is the transition to the authenticationRoute returned? 

`...`
`return this.transitionTo(authenticationRoute);`
`...`

This made me to waste quite some time today. After I overrode the beforeModel in my app and removed the return statement, the application worked as expected. 

Is there a reason for the return statement?

marcoow:
This removes unnecessary `return`s from the `AuthenticatedRouteMixin` and `UnauthenticatedRouteMixin` mixins - there's basically no difference between `return this.transitionTo(…` and `this.transitionTo(…` in `beforeModel` anyway and the code currently is a bit confusing actually.

closes #1238

### DIFF
--- a/addon/mixins/authenticated-route-mixin.js
+++ b/addon/mixins/authenticated-route-mixin.js
@@ -93,7 +93,7 @@ export default Mixin.create({
         this.set('session.attemptedTransition', transition);
       }
 
-      return this.transitionTo(authenticationRoute);
+      this.transitionTo(authenticationRoute);
     } else {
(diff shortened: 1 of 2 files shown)


marcoow:
@stevenwu: can you approve via the Review UI if this is fine so we get the green check mark?

PR #1252, opened 2017-03-08
Finish OAuth2 Implicit Grant Authenticator

dknutsen:
This PR wraps up a couple tiny things to finish off the awesome work that @256dpi did in #1057.

I rebased against master, fixed the linting issues, and addressed the 3 comments from the last review in #1057. I did not squash the commits from the other PR with my new one so it was a little clearer what was my edits and what was the original, however I can do that if need be.


addon/authenticators/oauth2-implicit-grant.js
```diff
@@ -0,0 +1,103 @@
...
+
+ @class OAuth2ImplicitGrantAuthenticator
+ @module ember-simple-auth/authenticators/oauth2-implicit-grant
+ @extends BaseAuthenticator
+ @public
+ */
+export default BaseAuthenticator.extend({
+  /**
+   Restores the session from a session data object; __will return a resolving
+   promise when there is a non-empty `access_token` in the session data__ and
```
marcoow: This should mention that we check for presence of `state` and `token_type` as well (if those are even needed - see below).

addon/authenticators/oauth2-implicit-grant.js
```diff
@@ -0,0 +1,103 @@
...
+
+   @method restore
+   @param {Object} data The data to restore the session from
+   @return {Ember.RSVP.Promise} A promise that when it resolves results in the session becoming or remaining authenticated
+   @public
+   */
+  restore(data) {
+    return new RSVP.Promise((resolve, reject) => {
+      if (!this._validateData(data)) {
+        return reject('the token could not be restored, invalid token: missing access_token, state, or token_type');
```
marcoow: Should we say something like `'Could not restore session from incomplete session data - "access_token", "state" or "token_type" missing.'`?

addon/authenticators/oauth2-implicit-grant.js
```diff
@@ -0,0 +1,103 @@
...
+   @public
+   */
+  authenticate(hash) {
+    return new RSVP.Promise((resolve, reject) => {
+      let response = this._parseResponse(hash);
+
+      if (response.error) {
+        reject(response.error);
+      } else if (!this._validateData(response)) {
+        reject('invalid token: missing access_token, state, or token_type');
```
marcoow: I guess this say something like `'Invalid auth params - "access_token", "state" or "token_type" missing.'`

marcoow: I'm also not actually sure whether/how token types other then `Bearer` would work in the browser?

marcoow: `state` is also only required when it was present in the client authorization request. If we need to check for it at all that check should probably be something that users can opt-in to depending on whether they set it for the auth request.

addon/authenticators/oauth2-implicit-grant.js
```diff
@@ -0,0 +1,103 @@
...
+    const regex = /([^#?&=]+)=([^&]*)/g;
+    let match;
+
+    // decode all parameter pairs
+    while ((match = regex.exec(query)) !== null) {
+      params[decodeURIComponent(match[1])] = decodeURIComponent(match[2]);
+    }
+
+    return params;
+  }
```
marcoow: I think we should move this into the route mixin and pass an actual hash (e.g. POJO) to the authenticator instead so we don't leak the query string into the authenticator.

addon/mixins/oauth2-implicit-grant-callback-route-mixin.js
```diff
@@ -0,0 +1,57 @@
+import Ember from 'ember';
+
+const { inject: { service }, Mixin } = Ember;
+
+export default Mixin.create({
```
marcoow: This needs documentation - have a look at e.g. the `AuthenticatedRouteMixin` for reference.

addon/mixins/oauth2-implicit-grant-callback-route-mixin.js
```diff
@@ -0,0 +1,57 @@
...
+   @method activate
+   @public
+   */
+  activate() {
+    let authenticator = this.get('authenticator');
+
+    let hash = '';
+
+    // test if fastboot is available without requiring the fastboot service to be present
+    if (typeof FastBoot === 'undefined') {
```
marcoow: There is a FastBoot service actually with an `isFastBoot` property that we should use here instead. Also we need to return early from this method when we're running in FastBoot actually.

addon/authenticators/oauth2-implicit-grant.js
dknutsen: 👍

addon/authenticators/oauth2-implicit-grant.js
dknutsen: 👍

addon/authenticators/oauth2-implicit-grant.js
dknutsen: My only quibble with 'Invalid auth params' is that it isn't clear exactly where those params are missing from (that is, the received token)

dknutsen: Are you saying that we should maybe only allow the "Bearer" token_type vs just checking that the "token_type" param is there?

dknutsen: Yeah it doesn't seem compliant with the spec to always require `state`, do you have any ideas on how the 'opt-in' mechanism could work?

addon/authenticators/oauth2-implicit-grant.js
dknutsen: 👍

addon/mixins/oauth2-implicit-grant-callback-route-mixin.js
dknutsen: 👍

addon/mixins/oauth2-implicit-grant-callback-route-mixin.js
dknutsen: I'm not as familiar with FastBoot but that seems straightforward enough. I'll see what I can do.

marcoow: see e.g. [this](https://github.com/simplabs/ember-simple-auth/blob/master/addon/mixins/application-route-mixin.js#L55) for reference

addon/authenticators/oauth2-implicit-grant.js
marcoow: > My only quibble with 'Invalid auth params' is that it isn't clear exactly where those params are missing from (that is, the received token)

Hm, I think the point is you're not actually receiving only the token with the redirect but a set of auth params that __include__ the token.

> Are you saying that we should maybe only allow the "Bearer" token_type vs just checking that the "token_type" param is there?

Yeah, I'm not actually sure how MAC tokens would work and what else would have to be added to ESA to make them work (although I'm pretty sure we'd have to add more things). In the OAuth 2.0 Password Grant authenticator [we only check for presence of the `access_token` though](https://github.com/simplabs/ember-simple-auth/blob/master/addon/authenticators/oauth2-password-grant.js#L386) so it's probably fine to do that here as well which simplifies the whole thing anyway.

dknutsen: Oh you're totally right I think I'm mixing that up with some JWT stuff I was working on a couple days ago. My bad! 😬 

I'll go ahead and leave the `token_type` check how it is for now then, that sounds reasonable to me.

marcoow: What I meant was actually that we should maybe not care about the `token_type` at all and only check for presence of the `access_token`. JWTs would generally be used as Bearer tokens as well anyway e.g.

dknutsen: Ah I see. So you don't think we should check for `token_type` even though it's [a required field](https://tools.ietf.org/html/rfc6749#section-4.2.2) in the spec? I'll defer to you on that, I'm certainly not an authority on any of this stuff.

marcoow: Yeah, I'd only assert on the presence of the `access_token` for now.

dknutsen:
@marcoow had a little time today to try and address your comments, let me know what you think! And thanks for your patience, I'm a noob with some of this.

tests/unit/mixins/oauth2-implicit-grant-callback-route-mixin-test.js
```diff
@@ -0,0 +1,48 @@
...
+
+      route.activate();
+      setTimeout(() => {
+        expect(route.error).to.eq('access_denied');
+        expect(session.authenticate).to.have.been.calledWith('authenticator:oauth2');
+        done();
+      }, 10);
+    });
+  });
+});
```
marcoow: This should actually have a test that asserts that the authenticator is used to authenticate the session and the hash is correctly passed (I guess we'll need to use sinon to mock `window.location.hash`).

addon/mixins/oauth2-implicit-grant-callback-route-mixin.js
```diff
@@ -0,0 +1,91 @@
...
+  activate() {
+    if (!testing && this.get('_isFastBoot')) {
+      return;
+    }
+
+    let authenticator = this.get('authenticator');
+
+    let hash = this._parseResponse(window.location.hash);
+
+    return this.get('session').authenticate(authenticator, hash).catch((err) => {
```
marcoow: I don't think this `return` is actually needed.

dknutsen: Yeah I was wondering about that, I let it be last time but I'll take it out.

dknutsen:
@marcoow I started implementing this in the dummy app with Facebook (since there's already an app ID and everything) but Facebook adds in the extra step of an authorization code, so I don't think it's a great example authentication provider. I think it would make more sense to use something else like Google. However I figured you may want to set up the app so you have control over the settings and client ID and such. What do you think?

dknutsen:
@marcoow I went ahead and added an implicit grant example using Google auth to the dummy app. I created a project and client id of my own. Let me know if you want me to change the auth provider or the project/client id. 

I think that addresses all of the items you mentioned above.

addon/mixins/oauth2-implicit-grant-callback-route-mixin.js
```diff
@@ -0,0 +1,102 @@
...
+    Passes the hash received with the redirection from the authentication
+    server to the
+    {{#crossLink "OAuth2ImplicitGrantAuthenticator"}}{{/crossLink}} and
+    authenticates the session with the authenticator.
+
+    @method activate
+    @public
+  */
+  activate() {
+    if (!testing && this.get('_isFastBoot')) {
```
marcoow: Why is this not executed when `testing`?

addon/mixins/oauth2-implicit-grant-callback-route-mixin.js
```diff
@@ -0,0 +1,102 @@
...
+
+  _isFastBoot: computed(function() {
+    const fastboot = getOwner(this).lookup('service:fastboot');
+
+    return fastboot ? fastboot.get('isFastBoot') : false;
+  }),
+
+  _windowLocationHash() {
+    // we wrap this so we can stub it with sinon
+    return window.location.hash;
```
marcoow: Can't we stub the `hash` property of `window.location` with sinon instead?

tests/unit/authenticators/oauth2-implicit-grant-test.js
```diff
@@ -0,0 +1,80 @@
...
+  let authenticator;
+  let server;
+
+  let data = {
+    'access_token': 'secret-token'
+  };
+
+  beforeEach(function() {
+    authenticator = OAuth2ImplicitGrant.create();
+    server = new Pretender();
```
marcoow: This does not seem to be used at all?

tests/dummy/config/environment.js
```diff
@@ -43,6 +43,8 @@ module.exports = function(environment) {
 
     apiHost: 'http://localhost:4200',
 
+    googleClientID: '271351926386-jdh86lc996s5r6kia2dk2r7edrsmke85.apps.googleusercontent.com',
```
marcoow: Here's the Client ID of the ESA example app: 694766332436-1g5bakjoo5flkfpv3t2mfsch9ghg7ggd.apps.googleusercontent.com

addon/mixins/oauth2-implicit-grant-callback-route-mixin.js
dknutsen: I think it was because I couldn't get the getOwner to return anything in testing mode, and _isFastBoot depends on getOwner. In retrospect there's probably a better way to handle that but I wasn't sure what it was.

marcoow: Yeah, we actually want to run this in tests. I guess you'd either have to stub the owner or simply stub `_isFastBoot` which is probably the better option.

addon/mixins/oauth2-implicit-grant-callback-route-mixin.js
dknutsen: That's what I was trying to do originally but it doesn't sound like that's possible (see below). It sounds like some of the window properties are set in such a way that they cannot be overridden. This seemed to be the most general solution, but I'm certainly open to other ideas if you have anything that you like better. It's definitely not the prettiest.  

References:
http://stackoverflow.com/questions/36678302/stubbing-window-location-href-with-sinon
http://stackoverflow.com/a/11960524/977044
https://groups.google.com/forum/?fromgroups=#!topic/sinonjs/MMYrwKIZNUU%5B1-25%5D

marcoow: Yeah, I was afraid that's the case already. In this case this should be fine I guess.

tests/dummy/config/environment.js
dknutsen: Looks like the ESA example app needs the following set in the **Authorized redirect URIs** section:`http://localhost:4200/callback`

addon/mixins/oauth2-implicit-grant-callback-route-mixin.js
dknutsen: Alright I'll try the latter, thanks!

tests/dummy/config/environment.js
marcoow: Can you point me to the page to set it at? I can't find it ;(

dknutsen:
Ok I think that takes care of that round. Other than the changed needed in the `Authorized redirect URIs` of the google app settings (should be `http://localhost:4200/callback`) I think the dummy app should be working as expected too.

tests/dummy/config/environment.js
dknutsen: Yeah it's a very confusing console IMHO. The URL for the credentials section is: `https://console.developers.google.com/apis/credentials?project=<project-name>` and then in the `OAuth 2.0 client IDs` table, find the right row, then click the Edit/pencil icon, scroll down to `Authorized redirect URIs` and add it there. Hope that works!

marcoow:
Awesome 🎉 

I updated the allowed URLs and now it works. I'll hold off merging this for now as we're planning to release 1.2.1 first and then 1.3 with this include after that. This should be ready to be merged once 1.2.1 is out though!

256dpi:
That's big! Thank you guys!

jamesdixon:
@marcoow any update on 1.3? Really looking forward to this being added 😄

marcoow:
@jamesdixon see https://github.com/simplabs/ember-simple-auth/pull/1057#issuecomment-298527005

addon/authenticators/oauth2-implicit-grant.js
marcoow: We're only checking for `access_token` now so this is fine.

addon/authenticators/oauth2-implicit-grant.js
```diff
@@ -0,0 +1,85 @@
...
+   Authenticates the session using the specified location `hash`
+   (see https://tools.ietf.org/html/rfc6749#section-4.2.2).
+
+   __If the access token is valid and thus authentication succeeds, a promise that
+   resolves with the access token is returned__, otherwise a promise that rejects
+   with the error code as returned by the server is returned
+   (see https://tools.ietf.org/html/rfc6749#section-4.2.2.1).
+
+   @method authenticate
+   @param {Object} hash The optional location hash
```
marcoow: I don't think this is actually optional?

addon/authenticators/oauth2-implicit-grant.js
```diff
@@ -0,0 +1,85 @@
...
+   @return {Ember.RSVP.Promise} A promise that when it resolves results in the session becoming authenticated
+   @public
+   */
+  authenticate(response) {
```
marcoow: Might be a good idea to rename this to `hash` so it's in line with the above docs.

dknutsen:
@marcoow made those change but the build is failing. Doesn't look like anything to do with this code though, and the tests passed locally before i committed. Any ideas? Should I resync with master?

marcoow:
@dknutsen: this is failing on master as well right now - probably some updated dependency. I'm having a look.

marcoow:
@dknutsen: should be fixed once #1308 is merged and this is rebased.

marcoow:
@dknutsen: #1308 has been merged

256dpi:
Yeah 🚢 it!

dknutsen:
@marcoow rebased and all tests pass now 👍

marcoow:
@dknutsen: I fixed the rebase - will merge once build is green.

jamesdixon:
@marcoow will you be releasing a documentation update soon for 1.3? I just tried to go to http://ember-simple-auth.com/api/classes/OAuth2ImplicitGrantAuthenticator.html, but it wasn't found.

Great work everyone!

dknutsen:
@jamesdixon there is an example of usage in the dummy app that you can take a look at in the meantime. Demonstrates a basic implicit grant authenticator and use of the callback route mixin.

marcoow:
Sorry, always forget to push the latest docs after releases - strong sign we should automate that I guess. Will push the update later today!

marcoow:
pushed the docs for 1.3.0: http://ember-simple-auth.com/api/

jamesdixon:
@marcoow thanks!

@dknutsen I'll check it out -- thank you!

jamesdixon:
Quick question for either of you: my intention was to use this to be able to authenticate via Google/Dropbox, store the access token, and then use it later to sync backup data to those services. 

I'm already using ESA to authenticate my app using JWT; can I use this to connect to Google/Dropbox without messing with my existing session?

PR #1257, opened 2017-03-13
Restore cookieExpirationTime from cookie value

stevenwu:
When `AdaptiveStore` is using a `CookieStore`, the cookie values (name, domain, path) are only accessible through `AdaptiveStore._cookies`. I don't think they are restored, so `proxyToInternalStore` cannot access them. Do they need to be set in `init` as well?

/cc @marcoow

### DIFF
--- a/addon/session-stores/adaptive.js
+++ b/addon/session-stores/adaptive.js
@@ -133,6 +133,11 @@ export default Base.extend({
       const options = this.getProperties('cookieDomain', 'cookieName', 'cookieExpirationTime', 'cookiePath');
       options._fastboot = this.get('_fastboot');
       options._cookies = this.get('_cookies');
+
+      let expirationCookieName = `${this.get('cookieName')}-expiration_time`;
+      let expirationTime = parseInt(options._cookies.read(expirationCookieName), 10);
+      this.set('cookieExpirationTime', expirationTime);
+
       store = this._createStore(Cookie, options);
     }
     this.set('_store', store);
@@ -140,12 +145,6 @@ export default Base.extend({
 
   _createStore(storeType, options) {
     const store = storeType.create(options);
-
-    if (!this.get('_isLocalStorageAvailable')) {
-      let expirationCookieName = `${this.get('cookieName')}-expiration_time`;
-      let expirationTime = parseInt(options._cookies.read(expirationCookieName), 10);
-      this.set('cookieExpirationTime', expirationTime);
-    }
 
     store.on('sessionDataUpdated', (data) => {
       this.trigger('sessionDataUpdated', data);
--- a/addon/session-stores/adaptive.js
+++ b/addon/session-stores/adaptive.js
@@ -134,11 +134,8 @@ export default Base.extend({
       options._fastboot = this.get('_fastboot');
       options._cookies = this.get('_cookies');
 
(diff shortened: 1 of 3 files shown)


addon/session-stores/adaptive.js
```diff
@@ -133,6 +133,11 @@ export default Base.extend({
       const options = this.getProperties('cookieDomain', 'cookieName', 'cookieExpirationTime', 'cookiePath');
       options._fastboot = this.get('_fastboot');
       options._cookies = this.get('_cookies');
+
+      let expirationCookieName = `${this.get('cookieName')}-expiration_time`;
+      let expirationTime = parseInt(options._cookies.read(expirationCookieName), 10);
+      this.set('cookieExpirationTime', expirationTime);
```
marcoow: I don't think we need this here actually. When the cookie service is instantiated it will restore its own `cookieExpirationTime` from the cookie anyways.

addon/session-stores/cookie.js
```diff
@@ -159,6 +159,11 @@ export default BaseStore.extend({
   init() {
     this._super(...arguments);
 
+    if (!this.get('cookieExpirationTime')) {
```
marcoow: I think we actually always want to override `cookieExpirationTime`. The idea here is that if you define your application session store as

```js
export default Cookie.extend({
  cookieExpirationTime: 60 * 60
});
```

but then have a remember-me action in the UI that updates the expiration time like

```js
rememberMe() {
  this.set('session.store.cookieExpirationTime', 60 * 60 * 24);
}
```

you'll want that value of `60 * 60 * 24` to be applied after the next reload although the initial value of `60 * 60` would be set in this case as well.

marcoow:
Hm, there's only one thing that I just though of - if we always override the expiration time with the value restored from the cookie you could actually never change the expiration time later on.

addon/session-stores/adaptive.js
stevenwu: I wasn't sure whether to add this, or change the test at https://github.com/simplabs/ember-simple-auth/pull/1257/files#diff-37b831f2713acdc8f0acf654aa735d70R310.

If the store is an `Adaptive` store, should we do something like
```
expect(store.get('_cookies.cookieExpirationTime')).to.equal(expirationTime);
```
?

marcoow: No, because the adaptive store's `cookieExpirationTime` property is already an alias for the underlying cookie store's `cookieExpirationTime` property anyway.

stevenwu: The test fails without this code. Maybe there's a bug elsewhere?

addon/session-stores/adaptive.js
```diff
@@ -133,7 +133,9 @@ export default Base.extend({
       const options = this.getProperties('cookieDomain', 'cookieName', 'cookieExpirationTime', 'cookiePath');
       options._fastboot = this.get('_fastboot');
       options._cookies = this.get('_cookies');
+
       store = this._createStore(Cookie, options);
+      this.set('cookieExpirationTime', store.get('cookieExpirationTime'));
```
stevenwu: Made a small change. If `cookieExpirationTime` is not set here, then it stays `null` because the setter in `proxyToInternalStore` is never called.

marcoow: Makes sense 👍

Issue #1090, opened 2016-11-03
Automatically logged out after refresh


anthonycollini:
I am having this issue where the system is logging me out. I have found a pattern that causes it. A logged out user makes an unauthorized call the to the API, then logs in. The next refresh will log you out. I can't see anything in my code, or anything that jumps out that would cause this. My local storage looks fine before and after logging in.

- Before logging in:
```
{"authenticated":{}}
```

- Make an authenticated call.

- After logging in:
```
{"authenticated":{"authenticator":"authenticator:oauth2-gateway","access_token":"XXXXX","token_type":"bearer","expires_in":600,"refresh_token":"XXXXXX","created_at":1477929759,"expires_at":1477930359010}}
```

- Refresh, and it logs you out of the ember app

owldesign:
@anthonycollini does local storage still have `authenticated` object set or is it cleared out after refresh? Is `oauth2-gateway` a custom authenticator, I don't see it in the list of built in once?

I recently had a similar issue with my custom authenticator, in my case I missed setting up the `restore(data)` method. 

I think when page refreshes `restore` gets run with the data from your `authenticated` object, it needs to return a resolved promise with passed data `resolve(data)`. See this example [authenticators/oauth2-password-grant.js](https://github.com/simplabs/ember-simple-auth/blob/master/addon/authenticators/oauth2-password-grant.js#L162)

Again, that is what I ran into and it sounds similar, good luck!

jfschaff:
Hi guys! And thanks! I had a similar issue with a custom Torii provider, in which I had not implemented the `fetch` function, which is called by ESA `restore`. Implementing `fetch` fixed it!

marcoow:
@anthonycollini: any more info on this? Particularly, can you share the code of your custom authenticator?

marcoow:
Closing for lack of feedback

jamiechong:
@marcoow I'm having a similar issue. Using a custom authenticator. When I login, I can see the localStorage is set correctly, but when I refresh the page, the localStorage is wiped clean. I never see the `restore` method of my authenticator called (likely because on page load the localStorage data is missing).

marcoow:
> I never see the restore method of my authenticator called (likely because on page load the localStorage data is missing).

Yes, that sounds like a good explanation. Please post the source of your Authenticator and the session store definition. Also, which ESA Version are you using?

jamiechong:
[...]

Issue #990, opened 2016-06-01
Refresh page causes logout


jmimi:
I use my custom authenticator whch is the same as device authenticator almost.
When I refresh the page, local storage cleared and user logged out.

```
Ember version : 2.5.1
Ember Simple Auth version : 1.1.0
```

Why?
Thanks in advance

jmimi:
`InternalSession.restore()` should be called once in initializing. But I didn't see anything.
I created a instance-initializer and calling that `restore` function, then everything is ok

marcoow:
@jmimi: you should never call `InternalSession.restore` yourself - that's public API that might change any time.

jmimi:
@marcoow I don't call `InternalSession.restore()` directly.
I have done this in my `session` service which inherited from ESA session service :
`this.get('session').restore();`

marcoow:
Yep, that's what you should not do ;)

jmimi:
@marcoow So what should I do to solve my problem?

marcoow:
Can you post the code of your custom authenticator?

jmimi:
It's so much similar to devise authenticator.

```
import Ember from 'ember';
import Base from 'ember-simple-auth/authenticators/base';
import config from 'my-app/config/environment';

const {RSVP: {Promise}, isEmpty, run, get, $} = Ember;

export default Base.extend({
    serverTokenEndpoint: Ember.computed(function () {
        return `${config.Backend.baseUrl}/${config.Backend.defaultNamespace}/accounts/login`;
    }),
    serverTokenEndpointLogout: Ember.computed(function () {
        return `${config.Backend.baseUrl}/${config.Backend.defaultNamespace}/accounts/logout`;
    }),

    restore(data) {
        const tokenAttribute = get(data, 'token');
        const identificationAttribute = get(data, 'username');

        if (!isEmpty(tokenAttribute) && !isEmpty(identificationAttribute)) {
            return Promise.resolve(data);
        } else {
            return Promise.reject();
        }
    },
    authenticate(identification, password) {
        return new Promise((resolve, reject) => {
            const data = {};
            data['password'] = password;
            data['username'] = identification;

            return this.makeRequest(data).then(
                (response) => run(null, resolve, response),
                (xhr) => run(null, reject, xhr.responseJSON || xhr.responseText)
            );
        });
    },
    invalidate() {
        return Promise.resolve();
    },

    makeRequest(data, options){
[...]

PR #1267, opened 2017-03-23
Fix issue where store is emptied on page refresh

jamiechong:
Related #990 #1090.

Early in the initialization of this object, `setUnknownProperty` gets called with `__OWNER__` and `__NAME_KEY__` (system properties?). This happens before `this.content` is present. The side effect is that `null` is persisted, which trashes the data within our store. This fix checks to make sure `this.content` exists before persisting. 

Related discussion #990, #1090

### DIFF
--- a/addon/internal-session.js
+++ b/addon/internal-session.js
@@ -158,7 +158,10 @@ export default ObjectProxy.extend(Evented, {
   setUnknownProperty(key, value) {
     assert('"authenticated" is a reserved key used by Ember Simple Auth!', key !== 'authenticated');
     let result = this._super(key, value);
-    this._updateStore();
+    if (this.content) {
+      // only update the store if there's content
+      this._updateStore();
+    }
     return result;
   },
 
--- a/addon/internal-session.js
+++ b/addon/internal-session.js
@@ -158,8 +158,7 @@ export default ObjectProxy.extend(Evented, {
   setUnknownProperty(key, value) {
     assert('"authenticated" is a reserved key used by Ember Simple Auth!', key !== 'authenticated');
     let result = this._super(key, value);
-    if (this.content) {
-      // only update the store if there's content
+    if (!(/^_/).test(key)) {
       this._updateStore();
     }
     return result;
--- a/tests/unit/internal-session-test.js
+++ b/tests/unit/internal-session-test.js
@@ -511,15 +511,31 @@ describe('InternalSession', () => {
 
   describe("when the session's content changes", function() {
     describe('when a single property is set', function() {
-      beforeEach(function() {
-        session.set('some', 'property');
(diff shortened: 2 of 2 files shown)


marcoow:
How can the case that this fixes ever happen? The [`content` property is set in the `init` method](https://github.com/simplabs/ember-simple-auth/blob/master/addon/internal-session.js#L27) and thus should never be empty. Can you add some more info on the exact scenario and add a test case?

jamiechong:
@marcoow my first guess is that the `__OWNER__` and `__NAME_KEY__` properties are set when you call `this._super(...arguments);`

My setup is pretty basic. Latest ember-cli and ember.

jamiechong:
@marcoow I created a bare bones app that demonstrates the login issue. https://github.com/jamiechong/simple-auth-bug

To reproduce: 
- `ember serve`
- go to localhost:4200
- login with anything (see the authenticator - it returns true always)
- refresh

You can see this [single commit](https://github.com/jamiechong/simple-auth-bug/commit/72a6248e3588fe4522458a3e88c00d592fbb7ca5) for the changes I made to the stock ember app.

marcoow:
This is a bug with Ember 2.13 actually. It can be fixed by downgrading to 2.12:

```
diff --git i/package.json w/package.json
index d71ab1b..2c7acc2 100644
--- i/package.json
+++ w/package.json
@@ -17,7 +17,7 @@
   "devDependencies": {
     "broccoli-asset-rev": "^2.4.5",
     "ember-ajax": "^2.4.1",
-    "ember-cli": "2.13.0-beta.1",
+    "ember-cli": "2.12",
     "ember-cli-app-version": "^2.0.0",
     "ember-cli-babel": "^6.0.0-beta.3",
     "ember-cli-dependency-checker": "^1.3.0",
@@ -29,12 +29,12 @@
     "ember-cli-shims": "^1.0.2",
     "ember-cli-sri": "^2.1.0",
     "ember-cli-uglify": "^1.2.0",
-    "ember-data": "^2.13.0-beta.1",
+    "ember-data": "^2.12",
     "ember-export-application-global": "^1.0.5",
     "ember-load-initializers": "^0.6.0",
     "ember-resolver": "^3.0.0",
     "ember-simple-auth": "^1.2.0",
-    "ember-source": "~2.13.0-beta.1",
+    "ember-source": "~2.12",
     "ember-welcome-page": "^2.0.2",
     "loader.js": "^4.2.3"
   },
```

That doesn't mean we don't need to fix it of course ;) First of all we'd need a failing test case that reproduces the scenario though…

marcoow:
#990 isn't actually related as this only occurs with Ember >= 2.13.

marcoow:
@jamiechong: I changed it so that it doesn't persist private keys in the store which is what we want I think.

PR #1273, opened 2017-03-27
update dependencies

marcoow:


package.json
```diff
@@ -34,60 +34,55 @@
...
     "body-parser": "^1.17.1",
     "broccoli-asset-rev": "^2.4.5",
-    "broccoli-merge-trees": "^2.0.0",
-    "broccoli-yuidoc": "~2.1.0",
+    "broccoli-yuidoc": "^2.1.0",
     "chai": "^3.5.0",
-    "cors": "~2.8.1",
+    "cors": "^2.8.1",
     "ember-ajax": "^2.4.1",
-    "ember-browserify": "1.1.13",
```
Turbo87: not needed anymore?

package.json
```diff
@@ -34,60 +34,55 @@
...
-    "ember-cli": "2.11.0",
-    "ember-cli-addon-tests": "^0.6.0",
-    "ember-cli-app-version": "^2.0.0",
+    "ember-cli": "2.12.0",
```
Turbo87: most recent is `2.12.1`

package.json
```diff
@@ -34,60 +34,55 @@
   "devDependencies": {
     "body-parser": "^1.17.1",
     "broccoli-asset-rev": "^2.4.5",
-    "broccoli-merge-trees": "^2.0.0",
```
Turbo87: not needed?

package.json
```diff
@@ -34,60 +34,55 @@
...
-    "ember-export-application-global": "^1.0.5",
-    "ember-load-initializers": "^0.6.3",
-    "ember-resolver": "^2.0.3",
-    "ember-sinon": "~0.6.0",
-    "ember-source": "^2.11.0",
+    "ember-export-application-global": "^2.0.0",
+    "ember-load-initializers": "^1.0.0",
+    "ember-resolver": "^3.0.0",
+    "ember-source": "~2.12.0",
+    "ember-sinon": "^0.6.0",
```
Turbo87: `sinon` should be above `source` to match alphabetic sort order

package.json
marcoow: I think this was a dependency that Ember CLI used to add? I didn't find any usage in ESA for this.

package.json
marcoow: Doesn't look like it - couldn't find any usage.

package.json
Turbo87: nope, it's usually used to adjust the build pipeline, but I'm not aware that Ember CLI ever added that as a project dependency

marcoow: Hm, maybe old ESA dependency, pretty sure it's not needed though. Will check again.

marcoow: No, don't need it - should be safe to remove.

PR #1275, opened 2017-03-27
Add ability to override or hydrate transition to authentication route

### LINKED ISSUE #1268: Passing arbitrary context through to the `authenticationRoute` when transition is rejected by the AuthenticatedRouteMixin
Great tool! Thanks so much. I rely on this code heavily at work. 

If a user hits a page where they must be authenticated, we currently kick them out and over to the `authenticationRoute` as provided by the `AuthenticatedRouteMixin`.  I have been tasked with adding context to this rejection. Basically, showing the user 'WHY' we kicked them out and over to the login page. In my implementation, I would do this by passing a queryParam through to the `authenticationRoute` transition, making it easy to define on each authenticatedRoute a reason why someone would not be allowed to go there unauthenticated. 

I would like to change [this line in the `authenticatedRouteMixin`](https://github.com/simplabs/ember-simple-auth/blob/master/addon/mixins/authenticated-route-mixin.js#L96) to pass arbitrary context to the `authenticationRoute`

from: 
```javasacript
this.transitionTo(authenticationRoute);
```
to:
```javasacript
this.transitionTo(authenticationRoute, get(this, 'authenticationRouteContext'));
```

Is anyone else interested in this? or should I abandon my quest to push this upstream from my implementation into `ember-simple-auth` core.

Duder-onomy:
Adds method to `authenticated-route-mixin` called `triggerAuthentication` which will now handle the actual transition to the configured authentication route

Adds notes to the README.md describing this change and its purpose
Adds test to the `authenticated-route-mixin` tests to ensure that this method is in fact called
fixes #1268

### DIFF
--- a/package.json
+++ b/package.json
@@ -65,7 +65,7 @@
     "ember-export-application-global": "^1.0.5",
     "ember-load-initializers": "^0.6.3",
     "ember-resolver": "^4.1.0",
-    "ember-sinon": "~0.6.0",
+    "ember-sinon": "~0.7.0",
     "ember-source": "^2.11.0",
     "eslint-config-simplabs": "^0.4.0",
     "eslint-plugin-ember": "^3.0.1",


README.md
```diff
@@ -276,6 +276,9 @@ Router.map(function() {
 The route to transition to if the session is not authenticated can also be
 [configured](#configuration) to be another one than `login`.
 
+This mixin defines a method called `triggerAuthentication` that handles the routing to the configured authentication route.
+This method can be overridden if there is some custom behavior you desire.
+
```
marcoow: I think this is too much detail for the README actually. Having this method in the API docs should be fine.

addon/mixins/authenticated-route-mixin.js
```diff
@@ -93,9 +90,26 @@ export default Mixin.create({
...
+  },
+
+  /**
+    Handles the transition to the authentication route when this mixin's beforeModel hook
+    determines that you are not authenticated.
+
+    This can be overridden to add custom data to the transition, or to override the transition entirely.
+
+    @method triggerAuthentication
+    @param {}
```
marcoow: This does not actually take any params.

addon/mixins/authenticated-route-mixin.js
```diff
@@ -93,9 +90,26 @@ export default Mixin.create({
...
+    @public
```
marcoow: I think we're marking methods like this as `@protected` actually.

tests/unit/mixins/authenticated-route-mixin-test.js
```diff
@@ -90,6 +91,14 @@ describe('AuthenticatedRouteMixin', () => {
         expect(route.transitionTo).to.have.been.calledWith(authenticationRoute);
       });
 
+      it('internally calls `triggerAuthentication` to accomplish the transition to the authentication route', function() {
+        let authenticationRoute = 'path/to/route';
+        route.set('authenticationRoute', authenticationRoute);
+
+        route.beforeModel(transition);
+        expect(route.triggerAuthentication).to.have.been.called;
+      });
```
marcoow: I don't think this needs to be tested. This is an internal implementation detail of the mixin that the test should not actually have to worry about.

marcoow:
argh, I messed up rebasing - rebased with #1278

PR #1278, opened 2017-03-29
Add ability to override or hydrate transition to authentication route

### LINKED ISSUE #1268: Passing arbitrary context through to the `authenticationRoute` when transition is rejected by the AuthenticatedRouteMixin

marcoow:
Adds method to `authenticated-route-mixin` called `triggerAuthentication` which will now handle the actual transition to the configured authentication route

Adds notes to the README.md describing this change and its purpose
Adds test to the `authenticated-route-mixin` tests to ensure that this method is in fact called
fixes #1268

### DIFF
--- a/README.md
+++ b/README.md
@@ -276,6 +276,9 @@ Router.map(function() {
 The route to transition to if the session is not authenticated can also be
 [configured](#configuration) to be another one than `login`.
 
+This mixin defines a method called `triggerAuthentication` that handles the routing to the configured authentication route.
+This method can be overridden if there is some custom behavior you desire.
+
 To prevent a route from being accessed when the session is authenticated (which
 makes sense for login and registration routes for example), mix the
 [`UnauthenticatedRouteMixin`](http://ember-simple-auth.com/api/classes/UnauthenticatedRouteMixin.html)
--- a/README.md
+++ b/README.md
@@ -276,9 +276,6 @@ Router.map(function() {
 The route to transition to if the session is not authenticated can also be
 [configured](#configuration) to be another one than `login`.
 
-This mixin defines a method called `triggerAuthentication` that handles the routing to the configured authentication route.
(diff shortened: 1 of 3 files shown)


Duder-onomy:
Rock! And during ember conf no less!

dkalinin:
```this.transitionTo``` doesn't interrupt ```beforeModel``` hook.
Therefore, if you have your route with overridden ```beforeModel```:
```javascript
export default Ember.Route.extend(AuthenticatedRouteMixin, {
  beforeModel() {
    this._super(...arguments);
    console.log('Do something that requires authentication'); // this will be executed
  }
}
```
```console.log``` will be executed even if ```AuthenticatedRouteMixin``` calls ```transitionTo```.
It means that only adding the mixin is not enough, I need to check ```session.isAuthenticated``` in overridden ```beforeModel``` hook.

Issue #1251, opened 2017-03-08
Switch to yarn


marcoow:
We should switch to `yarn` for a more stable build, a smaller `node_modules` and faster setup times.

PR #1283, opened 2017-04-04
Use yarn

stevenwu:
#1251


stevenwu:
Bower still contains 4 dependencies: `base64`, `bootstrap`, `pretender`, `sinonjs`. `pretender` and `sinonjs` can be removed. Do we want to do anything about `base64` and `bootstrap`?

marcoow:
I think for now it's fine to still have bower dependencies.

marcoow:
@stevenwu: I rebased on `master`

@Turbo87: can you have a look just to be sure?

.travis.yml
```diff
@@ -6,18 +6,19 @@ node_js:
...
-  - npm install -g bower phantomjs-prebuilt
+  - curl -o- -L https://yarnpkg.com/install.sh | bash
+  - export PATH=$HOME/.yarn/bin:$PATH
+  - yarn global add bower phantomjs-prebuilt
   - bower --version
   - phantomjs --version
 
 install:
-  - npm install
+  - yarn install
```
Turbo87: we prefer `yarn install --no-lockfile` (see https://github.com/ember-cli/ember-cli/blob/master/blueprints/addon/files/.travis.yml#L35) for addons, to make sure you get notified when a transitive dependency update breaks your addon. otherwise you have to wait for your users to notify you.

package.json
```diff
@@ -96,5 +96,6 @@
       "eslint-plugin-ember",
       "eslint-plugin-mocha"
     ]
-  }
+  },
+  "main": "index.js"
```
Turbo87: I don't think that's necessary for Ember CLI addons 🤔

stevenwu:
Thanks @Turbo87

PR #1288, opened 2017-04-13
Switch from ember-network to ember-fetch

ef4:
ember-fetch is being more actively maintained and is more up-to-date with the fetch spec.

See discussion in https://github.com/tomdale/ember-network/issues/17 and https://github.com/stefanpenner/ember-fetch/issues/21

The one change I made in the tests is because `headers.getAll` was [removed from the current WHATWG standard](https://www.fxsitecompat.com/en-CA/docs/2016/headers-getall-has-been-removed-in-favour-of-get-now-returning-all-values/), so ember-fetch doesn't implement it.

### DIFF
--- a/addon/authenticators/devise.js
+++ b/addon/authenticators/devise.js
@@ -1,6 +1,6 @@
 import Ember from 'ember';
 import BaseAuthenticator from './base';
-import fetch from 'ember-network/fetch';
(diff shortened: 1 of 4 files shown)


marcoow:
Thanks, we were just talking about this recently and now it's solved already 🎉

PR #1310, opened 2017-05-02
Torii restore generic data

elidupuis:
I think I've found a small bug in `authenticator:torii`. It seems that on `restore`, session data (possibly resolved by extra functionality in `authenticate`) is being clobbered by the data returned from `this.get('torii').fetch()`.

I've updated a test with some generic data to demonstrate a failing result and including a fix that simply merges the `sessionData` passed into `restore` with the data returned from the `.fetch()` call.

Let me know if I'm missing something here, or if you'd like anything changed/updated.

### DIFF
--- a/tests/unit/authenticators/torii-test.js
+++ b/tests/unit/authenticators/torii-test.js
@@ -40,9 +40,9 @@ describe('ToriiAuthenticator', () => {
           sinon.stub(torii, 'fetch').returns(RSVP.resolve({ some: 'other data' }));
         });
 
-        it('returns a promise that resolves with the session data', function() {
-          return authenticator.restore({ some: 'data', provider: 'provider' }).then((data) => {
-            expect(data).to.eql({ some: 'other data', provider: 'provider' });
+        it('returns a promise that resolves with the session data merged with the data fetched from torri', function() {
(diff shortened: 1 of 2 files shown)


elidupuis:
Looks like there's some tests failing on `master`. I can rebase after #1308 is merged if you're interested in this patch.

tests/unit/authenticators/torii-test.js
```diff
@@ -44,8 +44,8 @@ describe('ToriiAuthenticator', () => {
         });
 
         it('returns a promise that resolves with the session data', function() {
-          return authenticator.restore({ some: 'data', provider: 'provider' }).then((data) => {
-            expect(data).to.eql({ some: 'other data', provider: 'provider' });
+          return authenticator.restore({ some: 'data', provider: 'provider', other: 'unrelated data' }).then((data) => {
+            expect(data).to.eql({ some: 'other data', provider: 'provider', other: 'unrelated data' });
```
marcoow: How would `other` end up in the authenticated session data in the first place?

elidupuis: In this case, `other` is part of the data resolved by a custom `authenticator.authenticate()` method.

I'll put an example together shortly.

elidupuis:
Great! Sorry I never got around to providing a better example!

PR #1318, opened 2017-05-05
Removed unused variables

Alonski:
isEmpty and RSVP are defined but never used.

### DIFF
--- a/guides/managing-current-user.md
+++ b/guides/managing-current-user.md
@@ -63,7 +63,7 @@ invoked with a query param where the `me` param is present.
 // app/services/current-user.js
 import Ember from 'ember';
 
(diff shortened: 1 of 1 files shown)


guides/managing-current-user.md
```diff
@@ -63,7 +63,7 @@ invoked with a query param where the `me` param is present.
 // app/services/current-user.js
 import Ember from 'ember';
 
-const { inject: { service }, isEmpty, RSVP } = Ember;
```
marcoow: `RSVP` is used below as `Ember.RSVP` actually - I think it would be better to keep it here and change the usage below.

Alonski: All right I updated the PR

PR #1325, opened 2017-05-10
Remove bower

stevenwu:
How do we know if dependencies are imported correctly?


marcoow:
@Turbo87 is the expert for these things - can you have a look?

bower.json
```diff
@@ -1,9 +1,6 @@
 {
   "name": "ember-simple-auth",
   "dependencies": {
-    "base64": "~1.0.0",
-    "bootstrap": "~3.3.7",
-    "pretender": "~1.1.0",
-    "sinonjs": "~1.17.1"
+    "base64": "~1.0.0"
```
Turbo87: why keep the bower dependency when you have it as a npm dep too now?

index.js
```diff
@@ -14,11 +17,21 @@ module.exports = {
...
+      "Ember.libraries.register('Ember Simple Auth', '" + version + "');";
+    var registerVersionTree = writeFile(
+      'ember-simple-auth/register-version.js',
+      content
+    );
+    var base64Tree = new Funnel(path.dirname(require.resolve('base-64')), {
+      files: ['base64.js']
+    });
+
+    return MergeTrees([vendorTree, registerVersionTree, base64Tree]);
```
Turbo87: since you don't have a `vendor` folder you could skip the `vendorTree` here, but then again once you add a folder you might get surprised in the future when it would not work if you removed it here 😉

stevenwu: @Turbo87 I wasn't sure if we needed `vendorTree`. `return MergeTrees([registerVersionTree, base64Tree])` is ok?

Turbo87: I think that should be enough, but as I said above it also doesn't hurt to include the `vendorTree` there.

index.js
```diff
@@ -14,11 +17,21 @@ module.exports = {
...
+  treeForVendor() {
+    var content =
+      "Ember.libraries.register('Ember Simple Auth', '" + version + "');";
+    var registerVersionTree = writeFile(
+      'ember-simple-auth/register-version.js',
+      content
+    );
+    var base64Tree = new Funnel(path.dirname(require.resolve('base-64')), {
+      files: ['base64.js']
+    });
```
marcoow: I think this will not automatically polyfill the `atob` and `btoa` methods. We'd probably have to do that outselves in a vendor file.

marcoow: fixed that

Issue #1338, opened 2017-05-17
Fix Tests


marcoow:
The tests are currently failing which is most likely caused by a new ember-cli-fastboot release. This is bad as we can't merge anything currently.

marcoow:
Actually the test run hangs when the `ember new` command that we use in the FastBoot tests asks whether to override `.eslintrc.js` when installing `ember-cli-eslint`:

```
Overwrite .eslintrc.js? (Yndeh)
```

This might be fixed by upgrading to the latest Ember CLI that includes ember-cli/ember-cli#6974.

marcoow:
fixed via #1340

PR #1340, opened 2017-05-17
Fix tests

marcoow:
…by upgrading to the latest Ember CLI - see https://github.com/simplabs/ember-simple-auth/issues/1338#issuecomment-302219000


marcoow:
@Turbo87: the interesting thing is only the second FastBoot test fails (reproducibly) which might indicate that [stopping the server in the first test](https://github.com/simplabs/ember-simple-auth/blob/master/fastboot-tests/graceful-regression-test.js#L32) hasn't fully completed when the second test starts. Running the tests in the opposite order makes the other one fail (which is then running second).

Turbo87:
@rwjblue @stefanpenner @hjdivad this sounds like a `process.exit()` refactoring issue to me. can you confirm?

@marcoow in the meantime it might be worth trying to update to the latest v2.12.x release of CLI which should have that double ESLint file issue fixed too but hopefully none of the process exit issues yet.

marcoow:
@Turbo87: downgrading to 2.12.1 does indeed fix it 👍

PR #1343, opened 2017-05-18
handle JSON.parse exception in makeRequest

dulmandakh:
makeRequest is failing when response is not JSON string. For example if a server returns 401 with empty body or some non-JSON string.

### DIFF
--- a/addon/authenticators/oauth2-password-grant.js
+++ b/addon/authenticators/oauth2-password-grant.js
@@ -325,7 +325,13 @@ export default BaseAuthenticator.extend({
     return new RSVP.Promise((resolve, reject) => {
       fetch(url, options).then((response) => {
         response.text().then((text) => {
-          let json = text ? JSON.parse(text) : {};
+          var json;
+          try {
+            json = JSON.parse(text);
+          } catch (e) {
+            json = {};
+          }
+
           if (!response.ok) {
             response.responseJSON = json;
             reject(response);


marcoow:
I think error responses are defined in [RFC6749](https://tools.ietf.org/html/rfc6749) - I'm not quite sure what they have to look like though atm.

dulmandakh:
This patch resolved an issue I had

marcoow:
Yes, but the question is whether the response you're seeing is even allowed according to RFC6749.

michsindelar:
Yes, you are right, but all responses, which are not compatible with RFC6749 should be rejected!!! And I think this is quite useful, because sometimes you can not affect backend server and if it responds something like 500 Internal Server Error "Something went wrong", you get unfulfilled promise and login page without information.

marcoow:
Maybe [this](https://github.com/simplabs/ember-simple-auth/pull/1343/files#diff-931dc51edd167948e773858f72a47233R336) could be removed and the response parsing could be moved into the else [here](https://github.com/simplabs/ember-simple-auth/pull/1343/files#diff-931dc51edd167948e773858f72a47233R339) then?

michsindelar:
I think whole fetch should looks like

```javascript
fetch(url, options)
.then((response) => {
        if (response.status == 200) {
            resolve(JSON.parse(text));
        } else {
            reject(response);
        }
    });
})
.catch(reject)
```

marcoow:
yeah that seems about right

michsindelar:
Do you wanna PR or you can manage it by yourself?

marcoow:
PR would be great

addon/authenticators/oauth2-password-grant.js
```diff
@@ -325,7 +325,13 @@ export default BaseAuthenticator.extend({
     return new RSVP.Promise((resolve, reject) => {
       fetch(url, options).then((response) => {
         response.text().then((text) => {
-          let json = text ? JSON.parse(text) : {};
+          var json;
+          try {
+            json = JSON.parse(text);
+          } catch (e) {
```
RobinDaugherty: I've been battling this issue as well, so this is a good change, but this too broad. Catching all exceptions means that _any_ failure will be swallowed. Only a `SyntaxError` should be caught.

marcoow: good point 👍

marcoow:
replaced by #1442

PR #1353, opened 2017-05-25
Improve implicit grant authenticator to accept hash as string

alex-kovshovik:
OAuth2ImplicitGrantAuthenticator now accepts the location hash
either as an object or as a string.

When the string is passed in:

- The leading "#" character is removed if present.
- The string is parsed into an object.
- The rest of the flow works normally.

## Explanation

**TL;DR:** `authenticate` call will now accept either an object with access_token or a string with access_token encoded as GET parameters, which will definitely make my life easier - as a user if this library.

I'm building an Ember application that does not have its own backend, it only uses the external API and authenticates to that API via OAuth2, using "implicit grant" strategy.

The typical OAuth2 workflow uses token authentication, which goes like this (roughly):

1. Client app redirects user to the auth provider.
1. Auth provider redirects the user back with `authentication_code`.
1. Client app passes the received `authentication_code` to its backend.
1. Backend makes an API call to auth provider to exchange the `authentication_code` for `authentication_token`.
1. Backend stores the auth token and uses it for subsequent external API calls.

Since my app does not have a backend, there is no need to exchange `authentication_code` for the `authentication_token`, instead I can get the `authentication_token` from auth provider directly, so I use "implicit grant" strategy. However, in it the `authentication_token` is returned as a "location hash" parameter, which you can get by calling `window.location.hash`. This returns a string, which you're responsible for parsing. **Now the implicit grant authenticator will do it for you.**

### DIFF
--- a/addon/authenticators/oauth2-implicit-grant.js
+++ b/addon/authenticators/oauth2-implicit-grant.js
@@ -47,11 +47,13 @@ export default BaseAuthenticator.extend({
    (see https://tools.ietf.org/html/rfc6749#section-4.2.2.1).
 
    @method authenticate
-   @param {Object} hash The location hash
+   @param {Object} hash The location hash: object or a string
    @return {Ember.RSVP.Promise} A promise that when it resolves results in the session becoming authenticated
    @public
    */
   authenticate(hash) {
+    hash = typeof hash === 'string' ? this._parseLocationHash(hash) : hash;
+
     return new RSVP.Promise((resolve, reject) => {
       if (hash.error) {
         reject(hash.error);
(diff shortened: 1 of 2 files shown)


marcoow:
The `window.location.hash` is currently converted to an Object in the [`OAuth2ImplicitGrantCallbackMixin`](https://github.com/simplabs/ember-simple-auth/blob/master/addon/mixins/oauth2-implicit-grant-callback-route-mixin.js). Why does that not work for you?

alex-kovshovik:
It didn't out of the box, but now that I know of this mixin, I will check why it doesn't work. Thank you so much!

I will close this PR for now.

Issue #1352, opened 2017-05-25
Allow rejectWithResponse for successful but malformed responses


The-Don-Himself:
Before you submit an issue, please review [the guidelines](https://github.com/simplabs/ember-simple-auth/blob/master/CONTRIBUTING.md) for this repository.
-Done.

1. Please only create issues for bugs, feature requests etc. - mere questions can be asked (and usually will be answered) on stackoverflow.
- This issue is a feature request

2. Update to the latest release of Ember Simple Auth if possible (see the releases page).
- As of the time of writing, it applies to the latest versions 1.2 and below.

3. Update to the latest releases of Ember.js, jQuery, Ember Data if possible
- Done.

4. Include as much information as possible - this includes full stack traces etc.
- The issue is to suggest that the rejectWithResponse be activated even when there was a successful response but missing token. Currently, it's only set for 4xx and 5xx responses but I have a use case where a 2xx response is returned intentionally without a token. I have implemented service workers for my ambitious [web app](https://campus-discounts.com/), leveraging https://github.com/DockYard/ember-service-worker. I have now created a plugin extending it (beta) to enqueue requests that fail because of a flaky network and/or server errors https://github.com/campus-discounts/ember-service-worker-enqueue. It currently only applies to mutation requests POST/PUT/DELETE etc. The service worker intercepts all such requests and for those that fail because of no network or server errors. it responds with a 202 HTTP_ACCEPTED response then queues them in localforage to be processed later via background sync and/or periodically.

The issue is that Ember Simple Auth treats a 202 response as successful but then fails to authenticate because of a lack of token which is ok, but I would like to get the response status code so as to tell the user that the login attempt has been queued and will automatically be processed later, but this is not exposed. My current ugly hack is this:-
````
            this.get('session').authenticate('authenticator:oauth2', identification, password).catch((reason) => {
                if(reason == 'access_token is missing in server response') {
                    this.get('systemMessages').show('Your login attempt has been queued & will be processed when the network is stable');
....
````
But I can't rely on that for the long term, I propose to allow us to expose the response in the event of authentication failure not just if on error. If this is OK, I can submit a PR.

5. Set up a demo that demonstrates the issue on JSFiddle or JSBin. If your project is open source, provide a link to the online repo.
- N/A

marcoow:
> but I would like to get the response status code so as to tell the user that the login attempt has been queued and will automatically be processed later,

[...]

PR #1356, opened 2017-05-31
reject _refreshAccessToken with response

vvsalmin:
_refreshAccessToken rejection should return something. Then you can implement your own actions for situations when refresh is rejected due lack of internet connectivity or when refresh is rejected by server answering error (ie. token expired).

I really cannot see any reason why response shouldn't be returned by reject :)


marcoow:
see https://github.com/simplabs/ember-simple-auth/issues/1352#issuecomment-305106004 - ESA should reject with `Error`s instead of responses or strings.

PR #1389, opened 2017-07-03
[WIP] Dummy app on `now`

stevenwu:
###TODO

- [x] when we release a new tag, deploy server via travis (check npm release step)
- [x] make it so that the deployed server has a static domain name (so we don't have to modify the client's config when we deploy that; `*.now.sh` alias is fine)
- [x] when we release a new tag, also deploy client via travis; also with static `*.now.sh` alias
- [ ] add link to demo app on ember-simple-auth.com (after we verify this is working)


stevenwu:
@marcoow where is the source code for ember-simple-auth.com?

marcoow:
@stevenwu: in the [`gh-pages` branch](https://github.com/simplabs/ember-simple-auth/tree/gh-pages)

.travis.yml
```diff
@@ -39,8 +39,17 @@ deploy:
     tags: true
     repo: simplabs/ember-simple-auth
 
+after_deploy:
+  - npm i -g now-pipeline
+  - now-pipeline-prune && now-pipeline --alias ember-simple-auth.now.sh
+  - cd server && now-pipeline-prune && now-pipeline --alias ember-simple-auth-server.now.sh
```
marcoow: I think we need

```yaml
on:
  tags: true
  repo: simplabs/ember-simple-auth
```

here so we only ever deploy the latest tag (which is when we also release a new version to npm).

package.json
```diff
@@ -17,6 +17,8 @@
     "diagrams": "mermaid --sequenceConfig guides/assets/esa-initial-flow.config.json --width 2400 -o guides/assets guides/assets/esa-initial-flow.txt",
     "fastboot": "echo \"Please also run npm start\"; ember fastboot --serve-assets",
     "lint": "eslint app addon blueprints config server test-support tests *.js",
+    "now-build": "ember build --environment=production",
```
marcoow: I wonder whether we should actually not build production so people can inspect the app's source easily. Or we could deploy sourcemaps as well.

package.json
```diff
@@ -33,6 +35,8 @@
     "ember-cookies": "^0.0.13",
     "ember-fetch": "^2.1.0 || ^3.0.0",
     "ember-getowner-polyfill": "^1.1.0",
+    "fastboot-app-server": "^1.0.0-rc.5",
+    "node-fetch": "^1.7.1",
```
marcoow: Isn't this a subdependency of `ember-fetch` already?

.travis.yml
stevenwu: `after_deploy` should only run after deployments. We can keep an eye on Now and follow up if necessary.

stevenwu:
This should be ready. I guess there's no way to verify results until next release?

marcoow:
> I guess there's no way to verify results until next release?

I'm afraid not 😬 

> Prettier added some extra commas. Do we want to keep those?

let's only include related changes in this.

package.json
marcoow: we do ship sourcemaps now

PR #1400, opened 2017-07-19
fix storage event handling

marcoow:
This changes handling of the _"storage"_ event so that it uses `addEventListener`/`removeEventListener` instead of the jQuery API and also adds some proper tests.

This depends on #1403

### DIFF
--- a/addon/session-stores/local-storage.js
+++ b/addon/session-stores/local-storage.js
@@ -3,7 +3,7 @@ import Ember from 'ember';
 import BaseStore from './base';
 import objectsAreEqual from '../utils/objects-are-equal';
 
-const { RSVP, $: jQuery, computed, getOwner } = Ember;
+const { RSVP, computed, getOwner, run: { bind } } = Ember;
 
 /**
   Session store that persists data in the browser's `localStorage`.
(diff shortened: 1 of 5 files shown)


.travis.yml
```diff
@@ -3,7 +3,15 @@ language: node_js
 node_js:
   - "6"
 
-sudo: false
+sudo: required
+dist: trusty
```
Turbo87: this will slow down the CI builds. why are you changing this?

marcoow:
> Chrome change seems 1) unrelated

It's necessary for this as spying on `window.addEventListener`/`window.removeEventListener` doesn't work in PhantomJS.

Turbo87:
> It's necessary for this as spying on window.addEventListener/window.removeEventListener doesn't work in PhantomJS.

then it should IMHO still be in a separate PR that needs to be merged before this one

.travis.yml
marcoow: This was copied from ember-cookies (https://github.com/simplabs/ember-cookies/blob/master/.travis.yml#L7). Chrome installation fails without it.

Turbo87: nope, see https://github.com/ember-cli/ember-cli/pull/7161

marcoow:
@Turbo87 #1403

PR #1403, opened 2017-07-19
run tests in headless chrome

marcoow:


.travis.yml
```diff
@@ -3,7 +3,10 @@ language: node_js
 node_js:
   - "6"
 
-sudo: false
```
Turbo87: hmm, I think we should probably keep this line to make sure we're using containers 🤔

marcoow: good point - fixed

PR #1410, opened 2017-07-27
add simplabs header signature

marcoow:
bildschirmfoto 2017-07-27 um 20 44 18

### DIFF
--- a/index.html
+++ b/index.html
@@ -116,6 +116,18 @@
         text-align: center;
       }
 
+      .header {
+        text-align: right;
+        padding: 10px 0;
+      }
+
+      .header a {
+        color: #656565;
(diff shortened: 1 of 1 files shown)


marcoow:
I configured Travis CI so that it ignores branches where no `.travis.yml` is present so the failing build can be ignore here and in the future it should not try building `gh-pages` at all anymore.

EmanuelaTrabaldoLena:
Maybe like this? I think the orange bar fixed on top looks better

image

marcoow:
The idea was to use the same header we use on our other pages, e.g. http://codebot.chat or https://ember-workshop.simplabs.com.

PR #1411, opened 2017-07-27
Use "New Module Imports"

Turbo87:
see https://github.com/ember-cli/ember-rfc176-data


Turbo87:
@marcoow please do not release this yet. I've introduced a bug because the new imports are actually not available in the app folder of addons 😞

PR #1417, opened 2017-08-02
chore: doc routeAfterTransition on README

geekygrappler:


marcoow:
I think this is too much detail for the README. Also if we document this in the README we'd have to document all the other properties as well while it's all in the API docs anyway.

Issue #1429, opened 2017-08-31
Setting a baseURL at environment.js does not work, because it get overwritten at ember-simple-auth-initializer


wuarmin:
Hello,
I love this addon, but I have some trouble with setting up the baseURL as described in the docs:
[http://ember-simple-auth.com/api/classes/Configuration.html](http://ember-simple-auth.com/api/classes/Configuration.html). The setting has not effect, because of the initializer-code at app/initializers/ember-simple-auth.js.

marcoow:
`baseURL` is read-only as it merely reflects the setting in your `config/environment.js`. What is it you're trying to achieve?

wuarmin:
thank you for the fast response. I want to set the baseURL at environment.js, but that was not possible, because it got overwritten. I posted a PR, which fixes my issue.

marcoow:
Hm, `baseURL` is not supposed to be configured at all as I said above. It merely reflects the [Ember CLI setting from `config/environment.js`](https://ember-cli.com/user-guide/#history-api-and-root-url). Again, what is it you're trying to achieve here?

wuarmin:
My rootURL is '/', but this is not the baseURL ember-simple-auth should use. So I want to set it as described at the API Docs. I know I could overwrite sessionInvalidated, but the only reason I have to implement it, is that: 
`ENV['ember-simple-auth'] = {
  baseURL: ''
}` in environment.js has no effect.

marcoow:
The API Docs says it's a read-only property: 
<img width="653" alt="bildschirmfoto 2017-08-31 um 21 16 11" src="https://user-images.githubusercontent.com/1510/29941108-b14f25d0-8e91-11e7-826d-98ed3acb22cc.png">

The way to go is to override `sessionInvalidated` as you say.

wuarmin:
Ok, what do you think about changing the example code at the api docs. I think some readers could be confused, because at the example baseURL get set.

Thanks for the clarification.

marcoow:
Ouch - yeah, just seeing that example code at the top now. Basically the example should just be removed as all of the properties are read-only by now - PR removing that would be great ❤️

wuarmin:
Ok, I'll provide some, but let me know, why they are all readonly? I think its a pretty feature to set authenticationRoute, or routeAfterAuthentication. So everybody has to live with the defaults.

marcoow:
[...]

PR #1430, opened 2017-08-31
CHG: now config.baseURL won't get overwritten, if set in environment.js

wuarmin:
fixes the issue #1429. 

WDYT?


marcoow:
Thanks for the effort but this is meant to be read-only.

PR #1433, opened 2017-09-01
Remove confusing configuration-example from Configuration API

wuarmin:
PR to remove confusing config-example mentioned in #1429

### DIFF
--- a/addon/configuration.js
+++ b/addon/configuration.js
@@ -12,9 +12,6 @@ const DEFAULTS = {
 /**
   Ember Simple Auth's configuration object.
 
-  These Ember Simple Auth properties should no longer be overridden in the application's environment object, that's why they are marked readonly.
-  To change any of these values, set them in the respective routes.
-
   @class Configuration
   @extends Object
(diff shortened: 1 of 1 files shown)


addon/configuration.js
```diff
@@ -12,15 +12,8 @@ const DEFAULTS = {
...
-  object, e.g.:
-
-  ```js
-  // config/environment.js
-  ENV['ember-simple-auth'] = {
-    baseURL: 'path/to/base/url'
-  };
-  ```
+  These Ember Simple Auth properties should no longer be overridden in the application's environment object, that's why they are marked readonly.
+  To change any of these values, set them in the respective routes.
```
marcoow: I think we can leave this paragraph out - this info is already in the deprecation notes below.

wuarmin:
Ok, I removed the paragraph.

marcoow:
http://ember-simple-auth.com/api/ documents the current release version and this change has not been released yet. We'll update the docs as soon as a new release is published.

PR #1439, opened 2017-09-12
Fixed error response in password authenticator.

michsindelar:
Related to PR  #1343

### DIFF
--- a/addon/authenticators/oauth2-password-grant.js
+++ b/addon/authenticators/oauth2-password-grant.js
@@ -331,7 +331,7 @@ export default BaseAuthenticator.extend({
             } catch (e) {
               json = {};
             }
-            if (response.status === 200) {
+            if (response.ok) {
               resolve(json);
             } else {
               response.responseJSON = json;
--- a/addon/authenticators/oauth2-password-grant.js
+++ b/addon/authenticators/oauth2-password-grant.js
@@ -322,17 +322,24 @@ export default BaseAuthenticator.extend({
       merge(options.headers, clientIdHeader);
     }
     return new RSVP.Promise((resolve, reject) => {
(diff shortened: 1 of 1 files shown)


addon/authenticators/oauth2-password-grant.js
```diff
@@ -322,17 +322,24 @@ export default BaseAuthenticator.extend({
...
+      fetch(url, options)
+        .then((response) => {
+          response.text().then((text) => {
+            let json;
+            try {
+              json = JSON.parse(text);
+            } catch (e) {
+              json = {};
+            }
+            if (response.status === 200) {
```
marcoow: Can we continue using `response.ok` here?

michsindelar: If I'm right, response.ok does mean response.status >= 200 && response.status < 300 according to fetch, but RFC6749 allows only 200 for successful response.

marcoow: Yeah, relying on 200 would be a breaking change though and we'd need to include a deprecation if we see any other status codes. I'm all for doing that but should be a different PR.

michsindelar: It depends on you, I think this solution is fixing current bug. All responses, which are not in accordance with RFC6749 will be rejected.
But if you want, ill give there back response.ok;)

marcoow: Yes, please use `response.ok` for now - happy to merge another PR that adds a deprecation for status codes other than 200!

michsindelar:
Sorry, Im not familiar with Ember tests.

marcoow:
replaced by #1442

PR #1442, opened 2017-09-18
Fix JSON parsing exception in password grant authentication response

RobinDaugherty:
When the server responds with something unexpected, the authenticator was failing with an exception and actually rejecting with `null` if `rejectWithResponse` was disabled (the terrible default).

With this change, if the body cannot be parsed as JSON, it's stored as `responseText` in the response object. With `rejectWithResponse` disabled, the rejection contains the text of the response. With `rejectWithResponse` enabled, the response object in the rejection contains `responseText` and `responseJSON` is left undefined.

Failure to parse the response as JSON is also considered a failure, and the promise will be rejected in that case as well, containing the body of the server response. I don't love this behavior, but since `rejectWithResponse` is false by default, this seems to be the only way to be consistent. If `rejectWithResponse` is true, the response of course contains more detail.

This solves #1443 and #1439.

### DIFF
--- a/addon/authenticators/oauth2-password-grant.js
+++ b/addon/authenticators/oauth2-password-grant.js
@@ -248,7 +248,7 @@ export default BaseAuthenticator.extend({
           resolve(response);
         });
       }, (response) => {
-        run(null, reject, useResponse ? response : response.responseJSON);
+        run(null, reject, useResponse ? response : (response.responseJSON || response.responseText));
       });
     });
   },
@@ -325,12 +325,17 @@ export default BaseAuthenticator.extend({
     return new RSVP.Promise((resolve, reject) => {
       fetch(url, options).then((response) => {
         response.text().then((text) => {
-          let json = text ? JSON.parse(text) : {};
+          try {
+            response.responseJSON = text ? JSON.parse(text) : {};
+          } catch (SyntaxError) {
+            response.responseText = text;
+            reject(response);
+          }
+
           if (!response.ok) {
-            response.responseJSON = json;
             reject(response);
           } else {
-            resolve(json);
+            resolve(response.responseJSON);
           }
         });
       }).catch(reject);
--- a/tests/unit/authenticators/oauth2-password-grant-test.js
+++ b/tests/unit/authenticators/oauth2-password-grant-test.js
@@ -200,21 +200,53 @@ describe('OAuth2PasswordGrantAuthenticator', () => {
         server.post('/token', () => [400, { 'Content-Type': 'application/json', 'X-Custom-Context': 'foobar' }, '{ "error": "invalid_grant" }']);
       });
 
-      it('rejects with the correct error', function(done) {
+      it('rejects with the parsed JSON response', function(done) {
         authenticator.authenticate('username', 'password').catch((error) => {
           expect(error).to.eql({ error: 'invalid_grant' });
           done();
         });
       });
 
-      describe('when reject with response is enabled', function() {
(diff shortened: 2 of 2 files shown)


tests/unit/authenticators/oauth2-password-grant-test.js
```diff
@@ -228,6 +238,38 @@ describe('OAuth2PasswordGrantAuthenticator', () => {
         });
       });
     });
+
+    describe('when the authentication request fails without a valid response', function() {
+      beforeEach(function() {
+        server.post('/token', () => [500, { 'Content-Type': 'text/plain', 'X-Custom-Context': 'foobar' }, 'The server has failed completely.']);
+      });
+
+      it('rejects with the response body', function(done) {
```
marcoow: No need to use `done` here - you can `return authenticator.authenticate(…` instead.

tests/unit/authenticators/oauth2-password-grant-test.js
```diff
@@ -228,6 +238,38 @@ describe('OAuth2PasswordGrantAuthenticator', () => {
...
+          done();
+        });
+      });
+
+      describe('when rejectWithResponse is enabled', function() {
+        beforeEach(function() {
+          authenticator.set('rejectWithResponse', true);
+        });
+
+        it('rejects with xhr object containing responseText', function() {
```
marcoow: I'd say _"response object"_ here instead. With `fetch` this will not be an xhr anymore.

addon/authenticators/oauth2-password-grant.js
```diff
@@ -324,12 +324,17 @@ export default BaseAuthenticator.extend({
     return new RSVP.Promise((resolve, reject) => {
       fetch(url, options).then((response) => {
         response.text().then((text) => {
-          let json = text ? JSON.parse(text) : {};
+          try {
+            response.responseJSON = text ? JSON.parse(text) : {};
```
marcoow: I'm not sure we should always be modifying `response` like this. Actually modifying it at all was a bad thing to add but we now need to keep it for backwards compatibility. We don't always need to modify though but only when we [reject with the response](https://github.com/simplabs/ember-simple-auth/pull/1442/files#diff-931dc51edd167948e773858f72a47233R335).

tests/unit/authenticators/oauth2-password-grant-test.js
RobinDaugherty: I copied this from another example in this file, but I'm pretty sure this is necessary. The assertions are in the `catch` on the promise, so if the promise doesn't reject, the test will pass without any assertions being made.

tests/unit/authenticators/oauth2-password-grant-test.js
RobinDaugherty: Quite right.

tests/unit/authenticators/oauth2-password-grant-test.js
marcoow: ah that's a very good point :+1:

addon/authenticators/oauth2-password-grant.js
marcoow: @RobinDaugherty: what do you think about that? ^

RobinDaugherty: Sorry I haven't replied on this. I started making some changes and I ran into some new questions. Please consider this incomplete for now, and I'll respond on this when I can return to it.

PR #1467, opened 2017-10-30
fix login with facebook

clairton:


package.json
```diff
@@ -43,6 +43,7 @@
     "broccoli-yuidoc": "^3.0.0",
     "chai": "^4.1.0",
     "cors": "^2.8.4",
+    "ember-ajax": "^3.0.0",
```
marcoow: This should not be needed

clairton: simple auth error

Without it throw this error.

marcoow: Ah right, didn't realize we're using that in the demo 😞

PR #1470, opened 2017-11-09
Exclude /fastboot-tests from npm

simonihmig:
Don't know about `/tests`, they seem to be included intentionally?

Also was unsure about `docs`, `guides` and `server`, which are all published, does that make sense?

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -5,6 +5,9 @@
 /node-tests
 /tmp
 /fastboot-dist
+/fastboot-tests
+/docs
+/guides
 **/.gitkeep
 .bowerrc
 .editorconfig


marcoow:
Nice, thanks. I think all of the other folders you mention should probably be excluded as well actually.

simonihmig:
> I think all of the other folders you mention should probably be excluded as well actually.

Thought the same. The commented out lines came from https://github.com/simplabs/ember-simple-auth/pull/1389, so was not sure if there is same fancy stuff going on I don't understand!? 😉

I can update the PR if you don't see any reason to keep those...

marcoow:
Oh, thanks for the hint. We do indeed need the tests as we’re deploying the dummy app to now.sh as the demo for the website.

simonihmig:
Ok, so I leave it like this. Not sure what Travis does not like though, certainly not related. Seems master is also failing in node4!?

Btw, as I have you here... didn't have a chance to say thanks for EmberFest on the last evening, so... thanks! 😀  Was really a great event!

marcoow:
> Ok, so I leave it like this.

Excluding `/docs` and `/guides` should be fine - we only need `/server` and `/tests` so we can deploy the dummy app and the mock server to now.sh.

> Not sure what Travis does not like though, certainly not related. Seems master is also failing in node4!?

It looks like it's failing with Node 6 actually - I'll have a look.

> Btw, as I have you here... didn't have a chance to say thanks for EmberFest on the last evening, so... thanks! 😀 Was really a great event!

Thank you for coming! Hope to see you there next year (and in Portland in March?)

simonihmig:
> Excluding /docs and /guides

done!

> Hope to see you there next year (and in Portland in March?)

Sure, hope to make it! 👍

marcoow:
Tests should be fixed - can you rebase on `master`?

simonihmig:
Done!

marcoow:
thanks!

PR #1473, opened 2017-11-13
fix tests

### LINKED ISSUE #1472: Fix Build
The build is failing since this commit https://travis-ci.org/simplabs/ember-simple-auth/builds/298626798 which is likely unrelated to the failures though 😞

marcoow:
Since we're ignoring the `yarn.lock` file on Travis we get the new and fixed ember-try version there that changes our test runs so that they actually run against early versions of ember and ember-data which did lead to errors (as the scenarios generated automatically by ember-try don't include correct versions for ember-cli-shims and ember-data which broke our tests).

The only fix for now sadly is to go back to explicit ember-try scenarios…

closes #1472

### DIFF
--- a/config/ember-try.js
+++ b/config/ember-try.js
@@ -1,5 +1,189 @@
 /* eslint-env node */
 module.exports = {
   command: 'yarn test',
-  useVersionCompatibility: true
+  scenarios: [
+    {
+      name: 'default',
+      dependencies: { }
+    },
+    {
+      name: 'ember-1.12',
+      bower: {
+        dependencies: {
+          ember: '~1.12.0',
+          'ember-cli-shims': '0.0.6',
+          'ember-data': '~1.13.0',
+        },
+      },
+      npm: {
+        devDependencies: {
+          'ember-cli-shims': null,
+          'ember-data': '~1.13.0',
+          'ember-source': null,
+        },
+      },
+    },
+    {
+      name: 'ember-1.13',
+      bower: {
+        dependencies: {
+          ember: '~1.13.0',
+          'ember-cli-shims': '0.0.6',
+          'ember-data': '~1.13.0',
+        },
+      },
+      npm: {
+        devDependencies: {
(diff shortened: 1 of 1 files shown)


config/ember-try.js
```diff
@@ -1,5 +1,189 @@
...
+      name: 'default',
+      dependencies: { }
+    },
+    {
+      name: 'ember-1.12',
+      bower: {
+        dependencies: {
+          ember: '~1.12.0',
+          'ember-cli-shims': '0.0.6',
+          'ember-data': '~1.13.0',
```
Turbo87: shouldn't this use Ember Data 1.12 too?

package.json
```diff
@@ -39,6 +39,7 @@
   "devDependencies": {
     "body-parser": "^1.17.2",
     "bootstrap": "^3.3.7",
+    "bower": "^1.8.2",
```
Turbo87: this shouldn't be needed anymore, https://github.com/ember-cli/ember-try/issues/161 got fixed tonight.

config/ember-try.js
marcoow: Ember Data 1.12 doesn't seem to exist…

package.json
marcoow: 👍

config/ember-try.js
Turbo87: ah, indeed. according to https://github.com/emberjs/data/tags?after=v1.13.3 there was only v1.0.0 betas before the v1.13.0 release 🤔

marcoow: The Ember Data version doesn't really matter much for the ESA tests though. In fact it's only needed so that the current user can be fetched during an acceptance test where the fetching itself doesn't really relate to ESA at all.

marcoow: > there was only v1.0.0 betas before the v1.13.0 release 🤔

Those were the wild and crazy days, in particular with Ember Data after all 😜

PR #1475, opened 2017-11-14
During restore, preserve data from previous authenticate

BryanCrotaz:


marcoow:
What's the motivation for this?

BryanCrotaz:
Take the situation where the developer wants to store the userId in the store. According to your docs and your examples, they override authenticate and add user_Id to the hash.  All lovely and simple.

But when a token restore happens, the session store is stomped on by the result of the restore call, removing the user_id. So refreshing the page before this works, refreshing after doesn't.

The developer could also override restore and do

// pseudo code
data = get session.data.authenticated
super.then result => ...
  result.user_id = data.user_id
  return result

A lot of work for the developer just to keep the data in place.

Refresh token doesn't change the logged in principal, just the token, so any data grabbed during authenticate is still valid. If the dev wants they can provide a new value in restore, but this will be a rare case.

BryanCrotaz:
|Plus these instructions don't work without this fix for the above reason

https://github.com/simplabs/ember-simple-auth/blob/master/guides/managing-current-user.md

marcoow:
> According to your docs and your examples, they override authenticate and add user_Id to the hash. 

That's not really the idea. The idea is that the server responds with an additional `user_id` value. If you would override the `authenticate` method though you could override `restore` as well though.

The idea of the `sessionDataUpdated` event (which is used when the token is refreshed) is really that the authenticator can update anything in the session data, not just the token. In fact, in reality what might happen frequently I think is a token refresh resulting not only in a new access token but also in a new refresh token.

In the end, if you're using a custom authenticator you're responsible for making this work in all relevant scenarios. That said, it would be good to make the built-in authenticators more easily customizable by extracting methods that can be overridden and that allow fine-grained customizations without a lot of repetition. I don't think we need fundamental changes in the way the internal session works for that though.

BryanCrotaz:
yes you're responsible as the developer, but we can make it easy or we can make it hard
Having your custom data deleted when a token restore happens is unintuitive.

My PR is not a fundamental change - it's just a merge on restore instead of an overwrite.

BryanCrotaz:
`sessionDataUpdated` event is useless here - the user_id (in my example) is already gone when this fires

marcoow:
When the authenticator refreshes the token, it fires the `sessionDataUpdated` even to signal to the internal session that the session data has updated which then stores the new data. As I said above, anything in the session data might have updated. Merging the updated data instead of replacing the current data is a fundamental change that can break people's apps which is something that ESA as a project must make sure does not happen.

As discussed above, the alternative to make your situation easier is to make it easier to add the code that injects the `user_id` into the data that the authenticator resolves with to all the right places (authentication, restoration, token update etc.) which I agree is a valuable thing to do.

BryanCrotaz:
I agree that you must not break anyone's apps.

But how do you propose that this would break an app?

- The only difference is on restore.
- The only change is to preserve old data - new data is overwritten as before.
- Old data was not preserved prior to this fix so no one is relying on it.

BryanCrotaz:
The only place where custom data would be used would be by custom authenticators. They would have the same problem I just hit - that the data vanishes a few minutes after writing it to the session store.

So anyone doing this in the wild would have had to build the ugly workaround I pseudo-wrote above. 

This fix makes that case easier to implement, and doesn't break any other use case.

The use case for custom data on the session is to better describe the logged in principal - it might store user id, roles etc. Those only change on a rising edge of authenticate, and cannot change on restore. And if they occasionally do, then the developer can still override `restore`. But in the vast majority of cases, this fix makes the developer's life easier and minimises the API surface that they need to touch.

marcoow:
> But how do you propose that this would break an app?

The potentially breaking change is that data that would have been removed previously now remains present in the same scenario.

> The only place where custom data would be used would be by custom authenticators. They would have the same problem I just hit - that the data vanishes a few minutes after writing it to the session store.

This is a misunderstanding of an authenticator's API though. The contract between internal session and authenticator is that the internal session will always **replace** all of it's data (specifically the data kept in `data.authenticated`) with whatever it gets from the authenticator's `authenticate` or `restore` methods or the `sessionDataUpdated` event. So that's why I'm saying that if you're overriding one of the built-in authenticators and extend it so that it augments the returned data somehow you need to make sure it augments the data in both of the 2 methods and when triggering the event.

> This fix makes that case easier to implement, and doesn't break any other use case.

It might break people's app for the reason pointed out above. ESA has pretty wide adoption and seemingly minor changes have broken people's apps in the past.

marcoow:
Another reason not to merge session data but to always replace it is that when merging the authenticator could never **remove** anything in the session btw.

BryanCrotaz:
The only way this could break an app would be if an app put data into authenticated and expected it to vanish a few minutes later.

Note that I haven't changed the behaviour of `authenticate` and `invalidate`, only of `restore`.

If you believe that this changes the public api then just increment the major version.

But this fix makes it much easier to extend authenticators, halving the work that needs to be done to inject custom data.

marcoow:
> The only way this could break an app would be if an app put data into authenticated and expected it to vanish a few minutes later.

This is exactly the reason why we cannot merge this change.

We have discussed this extensively now and I think I have explained alternative solutions for making the thing you're trying to solve easier to solve. As I said a few times, I'd be happy to merge something that makes it easier to customize authenticators so that the data returned from the `authenticate` and `restore` methods as well as the `sessionDataUpdated` event is augmented with custom properties.

PR #1477, opened 2017-11-14
don't destructure Ember.testing

### LINKED ISSUE #1476: Destructuring of `Ember.testing` will not work.
Just happened to notice this:

https://github.com/simplabs/ember-simple-auth/blob/c295fee9b0bab6fdf7dcfca0cb706b68415caa95/addon/mixins/application-route-mixin.js#L10

Which is used here:

https://github.com/simplabs/ember-simple-auth/blob/c295fee9b0bab6fdf7dcfca0cb706b68415caa95/addon/mixins/application-route-mixin.js#L141-L149

`Ember.testing` is a getter/setter in Ember, and destructuring like this will only read its value at the time the module is evaluated. In this case, `Ember.testing` will most likely be `false` when the module is evaluated, then change to `true` _during_ the test. This behavior somewhat recently changed in https://github.com/emberjs/ember-test-helpers/pull/227 (which is included in `ember-cli-qunit@4.1.0` / `ember-qunit@3.0.0`).

I believe the fix here would be to remove the destructuring, and use `Ember.testing` directly inline.

marcoow:
close #1476

### DIFF
--- a/addon/authenticators/oauth2-password-grant.js
+++ b/addon/authenticators/oauth2-password-grant.js
@@ -13,7 +13,6 @@ import Ember from 'ember';
 import BaseAuthenticator from './base';
 import fetch from 'fetch';
 
-const { testing } = Ember;
 const assign = emberAssign || merge;
 const keys = Object.keys || emberKeys; // Ember.keys deprecated in 1.13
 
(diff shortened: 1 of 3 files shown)


rwjblue:
Thank you for tackling this @marcoow!

PR #1479, opened 2017-11-23
fix "modified twice" deprecation warning

BryanCrotaz:
…ice" deprecation when triggered by xhr outside runloop


BryanCrotaz:
Found a simpler way in #1481

PR #1480, opened 2017-11-23
fix bug where token could expire before being refreshed

BryanCrotaz:


addon/authenticators/oauth2-password-grant.js
```diff
@@ -344,7 +344,7 @@ export default BaseAuthenticator.extend({
         expiresAt = new Date(now + expiresIn * 1000).getTime();
       }
       const offset = this.get('tokenRefreshOffset');
-      if (!isEmpty(refreshToken) && !isEmpty(expiresAt) && expiresAt > now - offset) {
+      if (!isEmpty(refreshToken) && !isEmpty(expiresAt) && expiresAt > now + offset) {
```
marcoow: The idea is to refresh the token shortly (offset seconds or milliseconds - I don't recall the unit) before it would expire so I don't think this is what we want.

BryanCrotaz: That's the intent, however that's not what the code did. It refreshed *after* the expiry not before.

BryanCrotaz: (expiresAt - offset) > now is a more understandable way to write it

marcoow: Actually this is a bit different - the actual scheduling of the refresh is done here: https://github.com/BryanCrotaz/ember-simple-auth/blob/ad7cf824fa075497cccae645c2f86e6b02cb758a/addon/authenticators/oauth2-password-grant.js#L351. That runs `_refreshAccessToken` in `expiresAt - now - offset` which I think translates to _"the time it takes until `offset` seconds before the expiration"_. This check is only to verify that that calculation doesn't result in a negative number so probably replacing this with `expiresAt - now - offset > 0` would be best.

BryanCrotaz: If you get a negative number that means it’s already expired...

marcoow: Right, in that case it does not make sense to schedule a refresh. Also calling `run.later` with a negative delay is likely going to result in an error.

PR #1481, opened 2017-11-23
fix "modified twice" deprecation warning

BryanCrotaz:
`endPropertyChanges` was called twice

### DIFF
--- a/addon/internal-session.js
+++ b/addon/internal-session.js
@@ -111,8 +111,8 @@ export default ObjectProxy.extend(Evented, {
     set(this.content, 'authenticated', authenticatedContent);
     this._bindToAuthenticatorEvents();
 
-    return this._updateStore().then(() => {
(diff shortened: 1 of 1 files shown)


addon/internal-session.js
```diff
@@ -111,8 +111,8 @@ export default ObjectProxy.extend(Evented, {
     set(this.content, 'authenticated', authenticatedContent);
     this._bindToAuthenticatorEvents();
 
-    return this._updateStore().then(() => {
-      this.endPropertyChanges();
```
marcoow: I think the second call is in the rejection though. If `_updateStore` returns a resolving promise I think `endPropertyChanges` wouldn't be called at all anymore?

marcoow: @BryanCrotaz: any update on this?

marcoow:
closing due to inactivity

PR #1491, opened 2017-12-08
OAuth2 Password Grant: refresh access token from secondly interval

### LINKED ISSUE #831: Refresh tokens don't refresh when device is asleep longer than expiration time.
I'm using refresh tokens with a 15 minute expiration with everything working well, however I'm having an issue where if the users device goes to sleep for longer than the refresh token is active, when their device wakes back up, the next ajax request will log them out.

The problem is that while the device is asleep, the refresh timer doesn't kick off. Then, once the user wakes up their device and the application attempts to hit an endpoint, the access token is expired and the session is invalidated and user logs out.

Normally if a user reloads the application after the refresh token is expired, the session is restored and during that process a new access token is obtained. However, that session restore never happens as the timer expires but the device is asleep.

It would be nice if before making an ajax request, we can check to see if the token is expired and request a new one. This is the approach I take with Angular using an http interceptor to check for a 401, but Ember Data/ESA seems to have a different flow.

My approach to fix this is outlined below. The code isn't the best, but it gets it done.

1) Modify the `ajax` method in my adapter. I can access the session service and check if the token has expired. If it is, and user is authenticated, restore the session.

2) After session restore promise resolves, perform the original request.

Any thoughts or suggested ways around this? Happy to help put a pull request together as well if this option should be added to the adapter mixin.

My code is below, but any other workarounds are appreciated. Below is what I added to my application adapter.

``` js
  ajax() {
    const session = this.get('session.session');

    // We'll need to manually do the _super apply, so
    // it's bound correctly from within our promise return
    var args = arguments,
      self = this,
      _super = this._super;

    if (new Date().getTime() > session.get('content.authenticated.expires_at')) {
      return session.restore()
        .then(function() {
          return _super.apply(self, args);
        });
    } else {
    }
  }
```

rmachielse:
Fixes #831 

The main problem is that setTimeout is paused when a tab is asleep. This pr switches to using `setInterval` instead of `setTimeout` to reverify every "alive" second if the token needs to be refreshed.

### DIFF
--- a/addon/authenticators/oauth2-password-grant.js
+++ b/addon/authenticators/oauth2-password-grant.js
@@ -113,7 +113,7 @@ export default BaseAuthenticator.extend({
     return (Math.floor(Math.random() * (max - min)) + min) * 1000;
   }).volatile(),
 
-  _refreshTokenTimeout: null,
+  _refreshTokenInterval: null,
 
   _clientIdHeader: computed('clientId', function() {
     const clientId = this.get('clientId');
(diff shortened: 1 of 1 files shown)


rmachielse:
@marcoow do you have time to have a look at this soon? That would be greatly appreciated. 🙂 
Let me know if you need any further explanation or discussion on this!

marcoow:
Not yet - sorry. I'll try to get to it this week 😞

Issue #1496, opened 2017-12-18
Internet Explorer Syntax error


cdatehortuab:
Hello, this addon has a little issue when running a project including it in Internet Explorer.

The problem is browser says in this line: 
https://github.com/simplabs/ember-simple-auth/blob/50adfeb515f5983287695d0e3eeba814c308d79f/addon/session-stores/cookie.js#L298
there is a Syntax error.

So, the project never runs.

Thank you.

marcoow:
This should actually be transpiled by Babel. Which version of IE do you see the error in? What does your `config/targets.js` look like?

mdeanjones:
For a vanilla `targets.js` of

```
module.exports = {
  browsers: [
    'ie 9',
    'last 1 Chrome versions',
    'last 1 Firefox versions',
    'last 1 Safari versions'
  ]
};
```
the transpiled line becomes
```
_warn: function _warn() {
      (true && Ember.warn(...arguments));
}
```
making it the sole remaining splat in the output. Not sure if this one is on Babel.

marcoow:
I think this relates to `@ember/debug` being a special module that has all of its exports disabled in production builds. In `development` env,

```js
_warn() {
  warn(...arguments);
}
```

becomes

```js
_warn: function _warn() {
  (true && Ember.warn(...arguments));
}
```

while in production it becomes

```js
_warn: function _warn() {
  (false && Ember.warn(...arguments));
}
```

which effectively disables the call to `Ember.warn`. Pure Babel compiles the method to

```js
_warn: function _warn() {
  _debug.warn.apply(undefined, arguments);
}
```

which seems correct. To me this looks like a bug in the AST transform doing this special handling of `@ember/debug`. @Turbo87 might have an idea whether this has maybe been fixed already or where exactly this happens.

Turbo87:
see https://github.com/chadhietala/babel-plugin-debug-macros/issues/47

cah-danmonroe:
Yay, I just spent the last 4 hours tracking this down to pin point it to ember-simple-auth from 1.4.0 to 1.4.1.  I came here to report it but I guess I was a little late.

IE 10 and 11 for us.

rwjblue:
😩 sorry for not getting that upstream issue fixed already, haven't had the time to dig into it yet 😭

andyhot:
Also getting this with phantomjs in our automated tests

andyhot:
[...]

PR #1502, opened 2017-12-24
session-stores/cookie: Inline `_warn()` method

Turbo87:
`warn(...arguments)` unfortunately can't be used like this at the moment since `...arguments` won't be transpiled properly.

see https://github.com/chadhietala/babel-plugin-debug-macros/issues/47

Obviously it would be best if this issue was fixed in `babel-plugin-debug-macros` directly, but until then we should avoid using rest/spread with the `warn()` macro.

Resolves https://github.com/simplabs/ember-simple-auth/issues/1496

### DIFF
--- a/tests/unit/session-stores/shared/cookie-store-behavior.js
+++ b/tests/unit/session-stores/shared/cookie-store-behavior.js
@@ -1,5 +1,5 @@
-import Ember from 'ember';
 import { next, run } from '@ember/runloop';
+import { registerWarnHandler } from '@ember/debug';
 import {
   describe,
   beforeEach,
@@ -9,6 +9,17 @@ import {
 import { expect } from 'chai';
 import sinon from 'sinon';
 import FakeCookieService from '../../../helpers/fake-cookie-service';
+
+let warnings;
+registerWarnHandler((message, options, next) => {
+  // in case a deprecation is issued before a test is started
+  if (!warnings) {
+    warnings = [];
+  }
+
+  warnings.push(message);
+  next(message, options);
+});
 
 export default function(options) {
   let store;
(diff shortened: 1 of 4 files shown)


tests/unit/session-stores/shared/cookie-store-behavior.js
```diff
@@ -39,11 +39,12 @@ export default function(options) {
 
   describe('#persist', function() {
     beforeEach(function() {
-      sinon.spy(CookieSessionStore.prototype, '_warn');
+      sinon.spy(Ember, 'warn');
```
rwjblue: Hmm. This is what `Ember.Debug.registerWarnHandler` is for..

rwjblue: Check out:

https://github.com/emberjs/ember-test-helpers/blob/4fbc32f53c221cc4ac709ae9b99095197d6e3282/tests/test-helper.js#L37-L50

And then:

https://github.com/emberjs/ember-test-helpers/blob/4fbc32f53c221cc4ac709ae9b99095197d6e3282/tests/test-helper.js#L77-L103

(`registerDeprecationHandler` is essentially the same as `registerWarnHandler`)

Turbo87: does that work all the way down to Ember 1.12? (see https://github.com/simplabs/ember-simple-auth/blob/master/package.json#L99)

rwjblue: Yes, with the polyfill in your devDeps

rwjblue: https://github.com/rwjblue/ember-debug-handlers-polyfill

Turbo87: @rwjblue looks like that polyfill only supports `registerDeprecationHandler`, but not `registerWarnHandler` 🤔

michalsnik:
Awesome, thanks @Turbo87 🚀

PR #1509, opened 2018-01-05
upgrade to 2.18

snewcomer:
Minor upgrade to package.json

However, one issue I am encountering using the test helpers is this [line](https://github.com/simplabs/ember-simple-auth/blob/master/test-support/helpers/ember-simple-auth.js#L19).  Hopefully I can come up with a fix.  Essentially, when used in a consuming app on 2.18, its errors with `wait` is not defined.


package.json
```diff
@@ -37,7 +37,6 @@
     "ember-cli-is-package-missing": "^1.0.0",
     "ember-cookies": "^0.1.0",
     "ember-fetch": "^2.1.0 || ^3.0.0",
-    "ember-getowner-polyfill": "^1.1.0 || ^2.0.0",
```
marcoow: This is actually needed to be able to support old Ember versions (ESA still supports Ember down to 1.12). I'm not sure why the tests don't fail without this…

snewcomer:
@marcoow Any thoughts on this [line](https://github.com/simplabs/ember-simple-auth/blob/master/test-support/helpers/ember-simple-auth.js#L19) - I'm getting `wait` is not a function on ember 2.18 in the consuming app...  Lmk if I get this wrong, but I think perhaps we need to import a specific method such as [settled](https://github.com/emberjs/ember-test-helpers/blob/9eb3e32f8340d01704e950f317b282f07073f1e9/addon-test-support/%40ember/test-helpers/settled.js).  However, this only comes with `ember-mocha` or `ember-qunit` I believe.

marcoow:
Hm, yeah - I guess `wait` would have to be imported. I'm not sure how to do that in a way that works for all Ember versions from 1.12 to 2.18…

snewcomer:
Where is this `wait` method currently coming from?  

Another option would be to have two implementations....one `await authenticateSession...` in which the `wait` is removed (currently using a patched version like this in my project) and the other using the current implementation.

PR #1511, opened 2018-01-07
Clear cookie if set to an empty authenticated object

dbbk:
**Problem**
Many CDNs allow you to adjust caching based on the existence of a cookie. However, with the Cookie ESA store, when a user logs out their cookie will be set to `{authenticated: {}}`, leaving them with the existence of a cookie that essentially resolves as nothing.

**Solution**
This PR deletes the cookie if it is attempted to be set to `{authenticated: {}}`. This means that if any other metadata than an empty `authenticated` object exists, the cookie will still get set as normal. Therefore this should cover most out-of-the-box use cases.

This is my first PR so if I've done anything wrong please let me know! Thanks

### DIFF
--- a/addon/session-stores/cookie.js
+++ b/addon/session-stores/cookie.js
@@ -241,6 +241,9 @@ export default BaseStore.extend({
       });
       delete this._oldCookieName;
     }
+    if (value === JSON.stringify({ authenticated: {} })) {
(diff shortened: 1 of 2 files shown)


marcoow:
closing due to inactivity

PR #1519, opened 2018-01-18
add support for new ember-qunit 4.2.0 test helper

### LINKED ISSUE #1512: Using test helpers with new ember-qunit test hooks
I'm trying to use `authenticateSession` for an acceptance test that is using the new `setupApplicationTest` method and I get an error related to the `wait` helper not being defined. 

Is there support for the new test methods and if so, is there an example of usage?

```
import { module, test } from 'qunit';
import { setupApplicationTest } from 'ember-qunit';
import { visit, } from '@ember/test-helpers';
import { authenticateSession,  } from 'app/tests/helpers/ember-simple-auth';

module('Acceptance | auth/login', function(hooks) {
  setupApplicationTest(hooks);

  hooks.beforeEach(function() {
    this.user = server.create('user');
    server.create('company');
  });

  test('visiting / when already authenticated', async function(assert) {
    assert.expect(1);

    authenticateSession(this.owner, {user_id: 1});
    await visit('/');

    assert.equal(currentURL(), '/main');
  });
});
```

0xadada:
If a consuming application is using `ember-cli-qunit` [4.2.0 or greater](https://github.com/ember-cli/ember-cli-qunit/blob/master/CHANGELOG.md#v420-2017-12-17), they may want to migrate to the
[more modern testing syntax](https://dockyard.com/blog/2018/01/11/modern-ember-testing). In that case, they will need to be imported from the `ember-simple-auth` addon namespace rather than from the consuming application namespace.

```js
// tests/acceptance/…
import { currentSession, authenticateSession, invalidateSession} from 'ember-simple-auth/test-support';
```

**Note** The helpers imported from the addon namespace (i.e. `import { functionName } from 'ember-simple-auth/test-support';`) differ slightly from those in the consumer application namespace in that they should not take the `application` argument.

The new-style helpers have the following function signatures:
* `currentSession()` returns the current session.
* `authenticateSession(sessionData)` authenticates the session; the optional `sessionData` argument can be used to mock an authenticator response (e.g. a token or user).
* `invalidateSession()` invalidates the session.

Thanks to  @cibernox for the advice

Fixes #1512

### DIFF
--- a/README.md
+++ b/README.md
@@ -646,21 +646,68 @@ in `package.json`:
 ## Testing
 
 Ember Simple Auth comes with a __set of test helpers that can be used in
-acceptance tests__:
+acceptance tests__.
 
-* `currentSession(app)`: returns the current session.
-* `authenticateSession(app, sessionData)`: authenticates the session; the
-  optional `sessionData` argument can be used to mock an authenticator
-  response - e.g. a token.
-* `invalidateSession(app)`: invalidates the session.
+### ember-cli-qunit 4.2.0 or greater
 
-The test helpers can be imported from the `helpers/ember-simple-auth`
-module in the application's namespace:
+If your app is using `ember-cli-qunit` [4.2.0 or
+greater](https://github.com/ember-cli/ember-cli-qunit/blob/master/CHANGELOG.md#v420-2017-12-17),
+you may want to migrate to the [more modern testing
+syntax](https://dockyard.com/blog/2018/01/11/modern-ember-testing). In that
+case, helpers can be imported from the `ember-simple-auth` addon namespace.
+
+```js
+// tests/acceptance/…
+import { currentSession, authenticateSession, invalidateSession} from 'ember-simple-auth/test-support';
+```
+
+The new-style helpers have the following function signatures:
+* `currentSession()` returns the current session.
+* `authenticateSession(sessionData)` authenticates the session asynchronously;
+  the optional `sessionData` argument can be used to mock an authenticator
+  response (e.g. a token or user).
+* `invalidateSession()` invalidates the session asynchronously.
+
+New tests using the async `authenticateSession` helper will look like this:
+
+```js
+import { module, test } from 'qunit';
+import { setupApplicationTest } from 'ember-qunit';
+import { currentURL, visit } from '@ember/test-helpers';
+import { authenticateSession } from 'ember-simple-auth/test-support';
+
+module('Acceptance | super secret url', function(hooks) {
+  setupApplicationTest(hooks);
+
+  test('authenticated users can visit /super-secret-url', async function(assert) {
+    await authenticateSession({
+      userId: 1,
+      otherData: 'some-data'
+    });
+    await visit('/super-secret-url');
+    assert.equal(currentURL(), '/super-secret-url', 'user is on super-secret-url');
+  });
+});
+```
+
+### ember-cli-qunit 4.2.0 or earlier
+
+For existing apps, the test helpers are merged into your applications namespace,
+and can be imported from the `helpers/ember-simple-auth` module:
 
 ```js
 // tests/acceptance/…
 import { currentSession, authenticateSession, invalidateSession } from '<app-name>/tests/helpers/ember-simple-auth';
 ```
+
+**Note** These helpers require the `app` argument.
+
+These helpers have the following function signatures:
+* `currentSession(app)`: returns the current session.
+* `authenticateSession(app, sessionData)`: authenticates the session; the
+  optional `sessionData` argument can be used to mock an authenticator
+  response - e.g. a token.
+* `invalidateSession(app)`: invalidates the session.
 
 ## Other guides
 
--- a/README.md
+++ b/README.md
@@ -648,7 +648,7 @@ in `package.json`:
 Ember Simple Auth comes with a __set of test helpers that can be used in
 acceptance tests__.
 
-### ember-cli-qunit 4.2.0 or greater
+### ember-cli-qunit 4.2.0 and greater
 
 If your app is using `ember-cli-qunit` [4.2.0 or
 greater](https://github.com/ember-cli/ember-cli-qunit/blob/master/CHANGELOG.md#v420-2017-12-17),
@@ -690,7 +690,7 @@ module('Acceptance | super secret url', function(hooks) {
 });
 ```
 
-### ember-cli-qunit 4.2.0 or earlier
+### ember-cli-qunit 4.1.0 and earlier
 
 For existing apps, the test helpers are merged into your applications namespace,
 and can be imported from the `helpers/ember-simple-auth` module:
--- a/addon-test-support/index.js
+++ b/addon-test-support/index.js
@@ -1,0 +1,57 @@
+import { get } from '@ember/object';
+import { getContext, settled } from '@ember/test-helpers';
+import Test from 'ember-simple-auth/authenticators/test';
+
+const SESSION_SERVICE_KEY = 'service:session';
+const TEST_CONTAINER_KEY = 'authenticator:test';
+
+function ensureAuthenticator(owner) {
+  const authenticator = owner.lookup(TEST_CONTAINER_KEY);
+  if (!authenticator) {
+    owner.register(TEST_CONTAINER_KEY, Test);
+  }
+}
+
+/**
+ * Authenticates the session.
+ *
+ * @param {Object} sessionData Optional argument used to mock an authenticator
+ * response (e.g. a token or user).
+ * @return {Promise}
+ * @public
+ */
+export async function authenticateSession(sessionData) {
+  const { owner } = getContext();
+  const session = owner.lookup(SESSION_SERVICE_KEY);
+  owner.setupRouter(); // router must initialize fully before authentication
+  ensureAuthenticator(owner);
+  await session.authenticate(TEST_CONTAINER_KEY, sessionData);
+  return settled();
+}
+
+/**
+ * Returns the current session.
+ *
+ * @return {Object} a session service.
+ * @public
+ */
+export function currentSession() {
+  const { owner } = getContext();
+  return owner.lookup(SESSION_SERVICE_KEY);
+}
+
+/**
+ * Invalidates the session.
+ *
+ * @return {Promise}
+ * @public
+ */
+export async function invalidateSession() {
+  const { owner } = getContext();
+  const session = owner.lookup(SESSION_SERVICE_KEY);
+  const isAuthenticated = get(session, 'isAuthenticated');
+  if (isAuthenticated) {
+    await session.invalidate();
+  }
+  return settled();
+}
--- a/package.json
+++ b/package.json
@@ -54,7 +54,7 @@
     "ember-cli-chai": "^0.4.2",
     "ember-cli-content-security-policy": "~1.0.0",
     "ember-cli-dependency-checker": "^2.0.1",
-    "ember-cli-eslint": "^4.1.0",
+    "ember-cli-eslint": "^4.2.3",
     "ember-cli-fastboot": "^1.0.2",
     "ember-cli-htmlbars": "^2.0.2",
     "ember-cli-htmlbars-inline-precompile": "^1.0.0",
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,7 +1,11 @@
 module.exports = {
   root: true,
+  parserOptions: {
+    ecmaVersion: 2017,
+    sourceType: 'module'
+  },
   extends: [
     'simplabs',
     'simplabs/plugins/ember',
-  ],
+  ]
 };
--- a/yarn.lock
+++ b/yarn.lock
@@ -94,20 +94,20 @@
   resolved "https://registry.yarnpkg.com/@glimmer/di/-/di-0.2.0.tgz#73bfd4a6ee4148a80bf092e8a5d29bcac9d4ce7e"
 
 "@glimmer/resolver@^0.4.1":
-  version "0.4.1"
-  resolved "https://registry.yarnpkg.com/@glimmer/resolver/-/resolver-0.4.1.tgz#cd9644572c556e7e799de1cf8eff2b999cf5b878"
+  version "0.4.2"
+  resolved "https://registry.yarnpkg.com/@glimmer/resolver/-/resolver-0.4.2.tgz#60c9b492e90bc3956ac82b3134649bd337e1651c"
   dependencies:
     "@glimmer/di" "^0.2.0"
 
 abab@^1.0.3:
-  version "1.0.3"
-  resolved "https://registry.yarnpkg.com/abab/-/abab-1.0.3.tgz#b81de5f7274ec4e756d797cd834f303642724e5d"
+  version "1.0.4"
+  resolved "https://registry.yarnpkg.com/abab/-/abab-1.0.4.tgz#5faad9c2c07f60dd76770f71cf025b62a63cfd4e"
 
 abbrev@1:
-  version "1.1.0"
-  resolved "https://registry.yarnpkg.com/abbrev/-/abbrev-1.1.0.tgz#d0554c2256636e2f56e7c2e5ad183f859428d81f"
+  version "1.1.1"
+  resolved "https://registry.yarnpkg.com/abbrev/-/abbrev-1.1.1.tgz#f8f2c887ad10bf67f634f005b6987fed3179aac8"
 
-accepts@1.3.3, accepts@~1.3.3:
+accepts@1.3.3:
   version "1.3.3"
   resolved "https://registry.yarnpkg.com/accepts/-/accepts-1.3.3.tgz#c3ca7434938648c3e0d9c1e328dd68b622c284ca"
   dependencies:
@@ -137,37 +137,37 @@ acorn@^3.0.4:
   version "3.3.0"
   resolved "https://registry.yarnpkg.com/acorn/-/acorn-3.3.0.tgz#45e37fb39e8da3f25baee3ff5369e2bb5f22017a"
 
-acorn@^4.0.1, acorn@^4.0.3, acorn@^4.0.4:
+acorn@^4.0.4:
   version "4.0.13"
   resolved "https://registry.yarnpkg.com/acorn/-/acorn-4.0.13.tgz#105495ae5361d697bd195c825192e1ad7f253787"
 
-acorn@^5.0.1:
-  version "5.1.1"
-  resolved "https://registry.yarnpkg.com/acorn/-/acorn-5.1.1.tgz#53fe161111f912ab999ee887a90a0bc52822fd75"
+acorn@^5.2.1:
+  version "5.3.0"
+  resolved "https://registry.yarnpkg.com/acorn/-/acorn-5.3.0.tgz#7446d39459c54fb49a80e6ee6478149b940ec822"
 
 after@0.8.1:
   version "0.8.1"
   resolved "https://registry.yarnpkg.com/after/-/after-0.8.1.tgz#ab5d4fb883f596816d3515f8f791c0af486dd627"
 
-ajv-keywords@^1.0.0:
-  version "1.5.1"
-  resolved "https://registry.yarnpkg.com/ajv-keywords/-/ajv-keywords-1.5.1.tgz#314dd0a4b3368fad3dfcdc54ede6171b886daf3c"
+ajv-keywords@^2.1.0:
+  version "2.1.1"
+  resolved "https://registry.yarnpkg.com/ajv-keywords/-/ajv-keywords-2.1.1.tgz#617997fc5f60576894c435f940d819e135b80762"
 
-ajv@^4.7.0, ajv@^4.9.1:
+ajv@^4.9.1:
   version "4.11.8"
   resolved "https://registry.yarnpkg.com/ajv/-/ajv-4.11.8.tgz#82ffb02b29e662ae53bdc20af15947706739c536"
   dependencies:
     co "^4.6.0"
     json-stable-stringify "^1.0.1"
 
-ajv@^5.2.0:
-  version "5.2.2"
-  resolved "https://registry.yarnpkg.com/ajv/-/ajv-5.2.2.tgz#47c68d69e86f5d953103b0074a9430dc63da5e39"
+ajv@^5.1.0, ajv@^5.2.3, ajv@^5.3.0:
+  version "5.5.2"
+  resolved "https://registry.yarnpkg.com/ajv/-/ajv-5.5.2.tgz#73b5eeca3fab653e3d3f9422b341ad42205dc965"
   dependencies:
     co "^4.6.0"
     fast-deep-equal "^1.0.0"
+    fast-json-stable-stringify "^2.0.0"
     json-schema-traverse "^0.3.0"
-    json-stable-stringify "^1.0.1"
 
 align-text@^0.1.1, align-text@^0.1.3:
   version "0.1.4"
@@ -176,18 +176,6 @@ align-text@^0.1.1, align-text@^0.1.3:
     kind-of "^3.0.2"
     longest "^1.0.1"
     repeat-string "^1.5.2"
-
-alter@~0.2.0:
-  version "0.2.0"
-  resolved "https://registry.yarnpkg.com/alter/-/alter-0.2.0.tgz#c7588808617572034aae62480af26b1d4d1cb3cd"
-  dependencies:
-    stable "~0.1.3"
-
-amd-name-resolver@0.0.6:
-  version "0.0.6"
-  resolved "https://registry.yarnpkg.com/amd-name-resolver/-/amd-name-resolver-0.0.6.tgz#d3e4ba2dfcaab1d820c1be9de947c67828cfe595"
-  dependencies:
-    ensure-posix-path "^1.0.1"
 
 amd-name-resolver@0.0.7:
   version "0.0.7"
@@ -205,9 +193,9 @@ amdefine@>=0.0.4:
   version "1.0.1"
   resolved "https://registry.yarnpkg.com/amdefine/-/amdefine-1.0.1.tgz#4a5282ac164729e93619bcfd3ad151f817ce91f5"
 
-ansi-escapes@^2.0.0:
-  version "2.0.0"
-  resolved "https://registry.yarnpkg.com/ansi-escapes/-/ansi-escapes-2.0.0.tgz#5bae52be424878dd9783e8910e3fc2922e83c81b"
+ansi-escapes@^1.1.0:
+  version "1.4.0"
+  resolved "https://registry.yarnpkg.com/ansi-escapes/-/ansi-escapes-1.4.0.tgz#d3a8a83b319aa67793662b13e761c7911422306e"
 
 ansi-escapes@^3.0.0:
   version "3.0.0"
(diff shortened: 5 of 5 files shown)


cibernox:
This looks good but it will not be backwards compatible with the old testing approach because `import { getContext}`  will fail as it's not a known module unless you're on ember-qunit 4.2.


My suggestion is to put the new helpers in `/addon-test-support`, extract the shared funcinality to a private function and use that function from both the old helper and the new helpers.

This is exactly what I did here: https://github.com/jamesarosen/ember-i18n/pull/470

cibernox:
I'll try to put together an example later today, but it's going to be a long meetings day.

cibernox:
The first step is to update to the latest moche because we will need it to actually test the new helpers. Done in #1521

0xadada:
@marcoow @cibernox i updated the README, and refactored the helper a bit. Moved it into the new-style `addon-test-support` directory to it will be exported into the addon namespace rather than the consuming application namespace. 

I import the The `wait()` function from `ember-test-helpers` rather than rely on the global `app.testHelpers` property.

here https://github.com/0xadada/ember-simple-auth/commit/575d5589cd82c6318eced9fea0aa0f3716340e4e

addon-test-support/-private/index.js
```diff
@@ -0,0 +1,32 @@
+import wait from 'ember-test-helpers/wait';
```
cibernox: `import { settled } from '@ember/test-helpers';` instead of wait

addon-test-support/index.js
```diff
@@ -0,0 +1,37 @@
...
+
+/**
+ * Authenticates the session.
+ *
+ * @param {Object} sessionData Optional argument used to mock an authenticator
+ * response (e.g. a token or user).
+ * @return {Promise}
+ * @public
+ */
+export function authenticateSession(sessionData) {
```
cibernox: This looks good but when I suggested to extract the logic to some functions that receive an `owner` is because I thought we could also reuse those functions in the existing helpers, but it seems we can't because for backwards compatibility reasons the old helpers have to use `app.testHelpers.wait`.

In that case, I think that extracting the logic into separate functions, while correct, it's not necessary.

Other than that, this looks perfect.

0xadada: @cibernox @marcoow great feedback. implemented feedback, squashed commits 👍  in https://github.com/simplabs/ember-simple-auth/pull/1519/commits/ce6893a799d62d7e56f6682330bd6567c67c5293

jordpo:
@0xadada so testing these out locally, I'm running into a `hasRoute` of undefined error when the `authenticationSucceeded` event is triggered and hits the `ApplicationRouteMixin`

https://github.com/simplabs/ember-simple-auth/blob/master/addon/internal-session.js#L117

To reproduce, test visiting an authenticated route with an ApplicationRoute that has the mixin. At first I thought it was a timing issue and the test was finishing before the event could be fired but verified that that is not the case. 

I'm only experiencing this using the new test helpers and acceptance test setup. I have another test using `moduleForAcceptance` and it works fine without the `hasRoute` error.

0xadada:
@jordpo Yup, i was able to reproduce this. I'm not sure exactly why this is happening, i have a feeling its either something with `setupApplicationTest()` or a race condition. It seems the internal router property is not yet setup properly before the AuthenticatedRouteMixin makes use of it.

jordpo:
It seems to run fine if you call `authenticateSession` after you use the `visit` helper. It has to with setting up the application context fully before it gets invoked. I tried playing around with using the test helper `setupApplicationContext` but still getting the same error.

0xadada:
@jordpo i'm going to try adding a `visit()` above this line https://github.com/mirai-audio/mir/blob/16/library/tests/acceptance/media/new-test.js#L22 to test the workaround out, tonight.

0xadada:
@marcoow re: the `authenticateSession` test helper, @jordpo and I are trying to get a new qunit API compatible helper working with the addon to support Ember 3. 

Right now it's failing if an acceptance test calls `authenticateSession` as the first step of a test that excercises a route that extends `AuthenticatedRouteMixin`. The AuthenticatedRouteMixin calls the routers `transitionTo` method, but the router has not yet finished initializing: its internal `this._routerMicrolib` property is `undefined`, so the internal call `_doTransition` hits this line https://github.com/emberjs/ember.js/blob/v2.14.1/packages/ember-routing/lib/system/router.js#L773 where `this._routerMicrolib.hasRoute` is trying to access a property on `undefined`.

How does the existing helper assure that  the router is settled?

0xadada:
This is a successful test case

```js
import { module, test } from 'qunit';
import { setupApplicationTest } from 'ember-qunit';
import { currentURL, visit } from '@ember/test-helpers';
import { authenticateSession } from 'ember-simple-auth/test-support';

module('Acceptance | styleguide', function(hooks) {
  setupApplicationTest(hooks);

  test('authenticated users can visit /styleguide', async function(assert) {
    await visit('/');
    await authenticateSession({
      userId: 1,
      otherData: 'some-data'
    });

    await visit('/styleguide');

    assert.equal(currentURL(), '/styleguide');
  });
});
```

success case 
acceptance test calls  `visit('/')`.

A promise started by `ember-simple-auth/instance-initializers/setup-session-restoration` instance initializer calls the `beforeModel` hook to restore the session via `session.restore()`. `internal-session.js:restore()` returns a promise by invoking `this._callStoreAsync('restore')` (a), which invokes `this.restore` (b) which immedeately resolves with a promise call stack returning to (a). 

(b) in `restore` picks up with the promise resolves to the `then()` clause. There is no authentication in this scenario, so `restore()` invokes `this._clearWithContent`, updating `content` on `internal-session` to an empty object. it calls `this._clear()` which sets `isAuthenticated` to false on the internal-session,  and returns a promise by invoking `this.updateStore()`: this calls `_updateStore()` which sets an empty authenticator, and invokes `this._callStoreAsync('persist', data)` where data is an empty object. this (inner) `_this.callStoreAsync()` returns a new promise.

The `/` route beforeModel() hook detects the user is not authenticated, and redirects the user to `welcome` route. This calls `router.js.transitionTo()` By this point, `router.js:this._routerMicroLib` is an object with a `hasRoute`

The call stack is returned to (a) `this._callStoreAsync()` then finally rejects the promise.

acceptance test then calls `authenticateSession` (a) defined in  `ember-simple-auth/test-support/index`. 

`ember-simple-auth/test-support/index.js:authenticateSession`(b) calls session-service `authenticate()` with the test key and the mocked session data to return. `session.js:authenticate()` returns a call to its own internal session-service  `authenticate(...arguments)` defined in internal-session.js. `internal-session:authenticate()` returns a promise, and the promise immediately returns the stack back to `index.js:authenticateSession` (b) which returns a promise on `@ember/test-helpers:settled()`, returning the stack to the acceptance test (a).

The promise at `internal-session.js:authenticate() return authenticator.authenticate` resolves, and goes into the `then()` clause, invoking `internal-session.js:this._setup` and immediatly returning it (a promise).

`internal-session.js:_setup()` (c) invokes `this._updateStore()` and immedately reutrns (a promise). 

`internal-session.js`_updateStore()` sets the mocked session `data` to the authenticator object, and invokes `this._callStoreAsync(data)`. `_callStoreAsync` returns a resolved promise, returning call stack to (c).

`internal-session.js:_setup()` (c) then moves into the `then()` clause of the returning promise, triggers `authenticationSucceeded`.

`session.js:_forwardedSessionEvents()` event is triggered by the `authenticationSucceeded` event, and calls `this.trigger(event, ...arguments)` which invokes the _subscribeToSessionEvents() method on the `ApplicationRouteMixin`, calling the `sessionAuthenticated` event (d).

`application-route-mixin.js:sessionAuthenticated()` then invokes `this.transitionTo` passing the default `routeAfterAuthentication`, which has a value of `index` in this scenario.

After a chain of invokations, `ember-routing/system/router.js:_doTransition()` is invoked and runs a sanity check:
`(true && !(targetRouteName && this._routerMicrolib.hasRoute(targetRouteName)) && (0, _emberDebug.assert)('The route ' + targetRouteName + ' was not found', targetRouteName && this._routerMicrolib.hasRoute(targetRouteName)));`

and succeeds because `this._routerMicrolib` is now defined, so the call to `this._routerMicrolib.hasRoute` succeeds. The test continues successfully.

0xadada:
@jordpo can you try my branch again? I just pushed https://github.com/simplabs/ember-simple-auth/pull/1519/commits/b520d6bbd17ec2b99d792a75ae8a8b119af4d08a and i think it should fix it. Let me know.

0xadada:
@marcoow could you plz re-review? the builds are passing again 🎉

jordpo:
The latest works with my tests. 🎉

visoft:
I was struggling with this issue, thanks, @0xadada. My tests are now running as expected. 👍

0xadada:
This PR fixes https://github.com/simplabs/ember-simple-auth/issues/1531

cibernox:
From a first pass this looks like as expected: Great.

I'll review it for a last time tomorrow morning.

addon-test-support/index.js
```diff
@@ -0,0 +1,58 @@
...
+  const { owner } = getContext();
+  const session = owner.lookup(SESSION_SERVICE_KEY);
+  owner.setupRouter(); // router must initialize fully before authentication
+  ensureAuthenticator();
```
cibernox: If this function is private, you could just pass the owner in.

0xadada: I'm not sure I understand what you mean

0xadada: Oh wait, gotcha. So the getContext isn't looked up twice. I'll do that

0xadada:
@cibernox i addressed your comment in https://github.com/simplabs/ember-simple-auth/pull/1519/commits/377cbb1a33be82b3306171d497f12a19af1d9e9b

addon-test-support/index.js
```diff
@@ -0,0 +1,57 @@
...
+ * @return {Promise}
+ * @public
+ */
+export function authenticateSession(sessionData) {
+  const { owner } = getContext();
+  const session = owner.lookup(SESSION_SERVICE_KEY);
+  owner.setupRouter(); // router must initialize fully before authentication
+  ensureAuthenticator(owner);
+  session.authenticate(TEST_CONTAINER_KEY, sessionData);
+  return settled();
```
cibernox: Is authenticating the session an asynchronous task? 
If it's not, you don't need to return `settled()`, and if it is I would make `authenticateSession` an async function. I think it's sync.

addon-test-support/index.js
```diff
@@ -0,0 +1,57 @@
...
+ * @public
+ */
+export function invalidateSession() {
+  const { owner } = getContext();
+  const session = owner.lookup(SESSION_SERVICE_KEY);
+  const isAuthenticated = get(session, 'isAuthenticated');
+  if (isAuthenticated) {
+    session.invalidate();
+  }
+  return settled();
```
cibernox: Is invalidating the session an asynchronous task? 
If it's not, you don't need to return `settled()`, and if it is I would make `invalidateSession` an async function. I think it's sync.

addon-test-support/index.js
```diff
@@ -0,0 +1,57 @@
...
+ *
+ * @param {Object} sessionData Optional argument used to mock an authenticator
+ * response (e.g. a token or user).
+ * @return {Promise}
+ * @public
+ */
+export function authenticateSession(sessionData) {
+  const { owner } = getContext();
+  const session = owner.lookup(SESSION_SERVICE_KEY);
+  owner.setupRouter(); // router must initialize fully before authentication
```
cibernox: I never heard of this method before. Apparently it's private: https://github.com/emberjs/ember.js/blob/74cafbc3593319e803a2b999b99229eb0265235d/packages/ember-application/lib/system/application-instance.js#L165

@rwjblue Do we have a better way of initializing the router? Could that method be public? This is the only dubious thing I can see in the PR.

addon-test-support/index.js
0xadada: `session.authenticate` returns a promise, and an internal call there events are fired that work async (here https://github.com/simplabs/ember-simple-auth/blob/master/addon/internal-session.js#L117) then another call to `_callStoreAsync`, so looks pretty async.

addon-test-support/index.js
0xadada: I'd be interested to make both `authenticateSession` and `invalidateSession` async functions, but would like to get @marcoow opinion on the matter.

addon-test-support/index.js
cibernox: Indeed. In that case I think the best approach is to make the function async:
```js
export async function authenticateSession(sessionData) {
  const { owner } = getContext();
  const session = owner.lookup(SESSION_SERVICE_KEY);
  owner.setupRouter();
  ensureAuthenticator(owner);
  await session.authenticate(TEST_CONTAINER_KEY, sessionData);
  return settled();
}
```

addon-test-support/index.js
0xadada: @cibernox `setupRouter()` was added to address comment at https://github.com/simplabs/ember-simple-auth/pull/1519#issuecomment-359837396, i was seeing cases where a route had to be navigated to in order to initialize the router prior to calling `authenticateSession`. This fixed that problem, but would be interested to know from @rwjblue if there is a better way to do it.

cibernox: Yeah, I get why the routes has to be initialized, I'm only wondering if some wise elder can clarify if it's the best approach. 

Also, will this even works in non-application tests?? 

Usually you don't need to authenticate for renderingTests, but I guess it's not crazy to want to use this method to just setup some initial state on the session service. It would be good to at least test that it doesn't explode.

0xadada:
@cibernox i added an example to the "Testing" section of the readme in https://github.com/simplabs/ember-simple-auth/pull/1519/commits/2dabd87a394954cf17c968d8bd437b8e71a29220, which should help alot, it explicitly mentions the async nature of the helpers, have a look https://github.com/0xadada/ember-simple-auth/tree/new-ember-qunit-release#ember-cli-qunit-420-or-greater

0xadada:
Build failed with 

```
ember test
Could not start watchman
Visit https://ember-cli.com/user-guide/#watchman for more info.
Building
WARNING: ember-decorators (used in ember-popper): You have not installed @ember-decorators/babel-transforms. It has been extracted to a separate addon. See instructions for installation: https://github.com/ember-decorators/babel-transforms#readme
ember-getowner-polyfill is not required for Ember 2.3.0 and later, please remove from your `package.json`.
/home/travis/build/simplabs/ember-simple-auth/addon-test-support/addon-test-support/index.js
  23:8  error  Parsing error: Unexpected token async
✖ 1 problem (1 error, 0 warnings)
cleaning up
cleaning up...
Built project successfully. Stored in "/home/travis/build/simplabs/ember-simple-auth/tmp/class-tests_dist-yW47WZPy.tmp".
not ok 1 Chrome 64.0 - ESLint | addon-test-support addon-test-support/index.js
    ---
        message: >
            addon-test-support/index.js should pass ESLint
            
            23:8 - Parsing error: Unexpected token async (null)
        Log: |
    ...
```
line 23, index.js
```js
`export async function authenticateSession(sessionData) {`

**Comment by cibernox** [comment]
@0xadada You need to tell eslint to parse javascript with a more recent version of the spec: https://github.com/cibernox/ember-basic-dropdown/blob/master/.eslintrc.js#L4

**Comment by 0xadada** [comment]
@cibernox indeed, https://github.com/simplabs/ember-simple-auth/pull/1519/commits/0d99f193bff5ee3b17ca4b301cc9056dae014e98 fixed the build problem.

File: README.md [hunk 7]
```diff
@@ -646,22 +646,69 @@ in `package.json`:
...
+      userId: 1,
+      otherData: 'some-data'
+    });
+    await visit('/super-secret-url');
+    assert.equal(currentURL(), '/super-secret-url', 'user is on super-secret-url');
+  });
+});
+```
+
+### ember-cli-qunit 4.2.0 or earlier
```
marcoow: Should this be `### ember-cli-qunit 4.1.0 or earlier`?

package.json
```diff
@@ -109,4 +109,4 @@
       "eslint-plugin-mocha"
     ]
   }
-}
+}
```
marcoow: 👀

marcoow:
@0xadada thanks also for the great writeups in https://github.com/simplabs/ember-simple-auth/pull/1519#issuecomment-359837396 and https://github.com/simplabs/ember-simple-auth/pull/1519#issuecomment-360649355. I guess we could somehow fix that but it is unrelated to what this PR does anyway. Of course it is hugely valuable to have such a great writeup of what exactly happens though 🎉

README.md
0xadada: Yup, i'll fix that.

0xadada:
@marcoow re: comparability with older ember-cli-qunit

> can we somehow change them so that that is the case (compat w/ember-cli-qunit before 4.2.0)?

I talked to @cibernox about this issue in comment https://github.com/simplabs/ember-simple-auth/pull/1519#issuecomment-358677384 and we weren't sure how to be compatible with both since we make use of `getContext` from `@ember/test-helpers`. Perhaps @rwjblue knows a trick? 

`import { getContext, settled } from '@ember/test-helpers';`

As it is now, i think its acceptable to introduce the new helpers, as it gives ESA consumers an opt-in choice to use the new helpers today. We could provide a codemod for down the road when we want to deprecate it, or then update these helpers to be backwards compatible.

cibernox:
@marcoow Those new helpers *do* require ember-cli-qunit 4.2 or greater, so I don't think we're ready to remove the old ones just yet. The best we could have is extract shared logic to a helper function, however given that the helpers are so simple I don't think it will help.

We can however test both the old and the new helpers simultaneously, that's not a problem.

0xadada:
@marcoow thoughts per @cibernox and I's previous two comments?

marcoow:
> We can however test both the old and the new helpers simultaneously, that's not a problem.

If we can do that, that'd be great and I think we should add that. I'm not sure how that would work though as there's only one dummy app that we run acceptance tests against and that dummy app can only use one version of ember-cli-qunit? But I'm probably missing something.

cibernox:
@marcoow I just realized that this addon uses mocha :(

Generally speaking, you can tests your test helpers just "by using them" inside an acceptance test.
The way I test simultaneously both the old and the new helpers is by creating two tests:

```js
// tests/acceptance/old-helpers.js
import { moduleForComponent, test } from 'ember-qunit';

moduleForAcceptance('Old test helpers');

test('authenticateSession', function(assert) {
  authenticateSession(); // this magical global is the old test helper
  visit('/private-url');
  andThen(function() {
    assert.equal(currentUrl(), '/private-url', 'The user was allowed');
  });
});
```
```js
// tests/acceptance/new-helpers.js
import { module, test } from 'qunit';
import { setupApplicationTest } from 'ember-qunit';
import { visit, currentUrl } from '@ember/test-helpers';
import { authenticateSession } from 'ember-simple-auth/test-support';
 
module('New test helpers', function(hooks) {
  setupApplicationTest(hooks);

  test('authenticateSession', async function(assert) {
    await authenticateSession(); // this imported function is the new test helpers
    await visit('/private-url');
    assert.equal(currentUrl(), '/private-url', 'The user was allowed');
  });
});
```

In qunit 4.2 and greater both testing approaches coexist.`moduleForAcceptance`, `moduleForComponent` and `moduleFor` are not even deprecated yet, so they will coexist for a while still.

marcoow:
> I just realized that this addon uses mocha :(

I'd say _"mocha :)"_ actually ;)

> Generally speaking, you can tests your test helpers just "by using them" inside an acceptance test.

Yes, that's what we're doing and if we can simply duplicate the test and use the new set of helpers in that that's great.

cibernox:
@Turbo87 I might your help to test the new test helpers in a test suite that doesn't use qunit. `getContext` won't work on a test that doesn't use `setup(Application|Rendering)Test`.

Turbo87:
@cibernox yeah, I think testing the new helpers might not be possible until `ember-mocha` is updated to rely on the new `ember-test-helpers` versions. you can try to upgrade to the latest `ember-cli-mocha` beta release, that is depending on `ember-test-helpers`, and then at least the function and module should be available. it will probably still not work though as for the current-style tests I think nobody is calling `setContext()` yet. Worth experimenting with though.

cibernox:
I already updated it to the latest beta earlier this week, but it's still not quite there.

Turbo87:
any idea why? should be on the latest `ember-test-helpers` dep then, right?

cibernox:
It is, but without something akin to `setupApplicationTest` we don't have the `context`, which other helpers need (p.e `this.owner`)

Turbo87:
> it will probably still not work though as for the current-style tests I think nobody is calling setContext() yet.

yep, I think that's what I meant with ^

rwjblue:
Nah, it’s easy enough to replicate for testing purposes here. What Ember-QUnit is doing is super simple, and can easily be done in a beforeEach in a mocha module:

```
import { setupContext, setupRenderingContext } from ‘@ember/test-helpers’;

describe(‘new world order’, function() {
  let context;
  beforeEach(function() {
    context = {};
    return setupContext(context).then(() => setupRenderingContext(context));
  });
});

```

0xadada:
@rwjblue i took a whack at the test you stubbed out over here, on a branch: https://github.com/0xadada/ember-simple-auth/blob/new-ember-qunit-release--qunit-test-in-mocha/tests/acceptance/authentication-qunit-test.js#L48 the 2nd test blows up

```
"before each" hook for "can be visited when the session is authenticated" ...

Error: Assertion Failed: You cannot use the same root element (DIV) multiple times in an Ember.Application
    at new EmberError (assets/vendor.js:24930:25)
    at assert (assets/vendor.js:25173:15)
    at Class.setup (assets/vendor.js:57838:84)
    at assets/test-support.js:38265:20
    at tryCatcher (assets/vendor.js:63465:21)
    at invokeCallback (assets/vendor.js:63643:33)
    at publish (assets/vendor.js:63629:9)
    at assets/vendor.js:54710:9
    at invoke (assets/vendor.js:20718:24)
    at Queue.flush (assets/vendor.js:20611:25)
```

I'm not sure how to bring the testing aspect of this home.

marcoow:
@0xadada: I think you have to return the promise returned from e.g. `invalidateSession` so this

```js
it('cannot be visited when the session is not authenticated', function() {
  invalidateSession().then(() => {
    visit('/protected');

    return andThen(() => {
      expect(currentPath()).to.eq('login');
    });
  });
});
```

becomes

```js
it('cannot be visited when the session is not authenticated', function() {
  return invalidateSession().then(() => {
    visit('/protected');

    return andThen(() => {
      expect(currentPath()).to.eq('login');
    });
  });
});
```

0xadada:
@marcoow Updated test (https://github.com/0xadada/ember-simple-auth/commit/8476e9c6d86c3bdbce8b38f0a50e0424ee792636#diff-f3800bba3aa2a8fce48351e2ffbe2298R38), but didn't work. Any mocha experts know how i can debug the problem? It seems its silently failing and i can't figure out what it is.

marcoow:
Hm, my understanding is this should be working:

```js
describe('Acceptance: Authentication (for QUnit helpers)', function() {
  let context;
  let server;

  beforeEach(async function() {
    context = {};
    return setupContext(context).then(() => setupApplicationContext(context));
  });

  afterEach(function() {
    tryInvoke(server, 'shutdown');
    teardownApplicationContext(context);
    teardownContext(context);
  });
```

However, it is not and I'm not sure why…

rwjblue:
@marcoow - Not sure if related, but the afterEach's in your example also need to be chaining promise (`teardownApplicationContext` and `teardownContext` are also async).

rwjblue:
@0xadada - You also can't use `andThen` in the newer style tests, that should definitley be throwing an error...

0xadada:
@rwjblue i converted away from `andThen` to bare promise `then` and `catch` clauses, @marcoow i added the `teardown*` helpers in https://github.com/0xadada/ember-simple-auth/commit/20b7782df27d65dda22535f6ff59985d9ccbcd82. If you both like it, i'll merge that sub-branch down into this PR.

marcoow:
Yes, I like that! 🎉 I left a comment on the commit - I think there's some potential to simplify it further. I'm looking forward to get this in!

0xadada:
@marcoow i gave it another shot (in https://github.com/0xadada/ember-simple-auth/commit/2b8fba709230c4359e8e2a31a99df7ae1d6ed7c8) but its failing again. I'm truly stuck on the "you cannot reuse the same element" error.

marcoow:
You need to `return visit(…` as well, e.g.

```js
it('cannot be visited when the session is not authenticated', function() {
  return invalidateSession().then(function() {
    return visit('/protected').then(() => {
      expect(currentPath()).to.eq('login');
    });
  });
});
```

Also changing the setup to:

```js
beforeEach(function() {
  application = startApp();
  context = {};
  return setupContext(context).then(() => setupApplicationContext(context));
});

afterEach(function() {
  tryInvoke(server, 'shutdown');
  return teardownApplicationContext(context).then(() => teardownContext(context));
});
```

results in

```
ok 37 Chrome 64.0 - Acceptance: Authentication (for QUnit helpers) the protected route cannot be visited when the session is not authenticated
not ok 38 Chrome 64.0 - Acceptance: Authentication (for QUnit helpers) the protected route can be visited when the session is authenticated
  …
ok 39 Chrome 64.0 - Acceptance: Authentication (for QUnit helpers) the login route can be visited when the session is not authenticated
not ok 40 Chrome 64.0 - Acceptance: Authentication (for QUnit helpers) the login route cannot be visited when the session is authenticated
  …
```

so at least 2/4 tests running. The 2 others are failing with `owner.setupRouter is not a function` - I'm not sure yet why that is…

0xadada:
looks like 
```js
export async function authenticateSession(sessionData) {
  const { owner } = getContext();
  const session = owner.lookup(SESSION_SERVICE_KEY);

  // setupRouter does not exist
  owner.setupRouter(); 

  ensureAuthenticator(owner);
  await session.authenticate(TEST_CONTAINER_KEY, sessionData);
  return settled();
}
```
screen shot 2018-02-07 at 4 41 47 pm

rwjblue:
Need to use `setApplication` instead of `setResolver` in test-helper.js

Turbo87:
I dont think that will work with `ember-mocha` yet...

0xadada:
I got it past the `setupRouter` error, but now failing because the session service isn't available

https://github.com/0xadada/ember-simple-auth/commit/03de24d44460bea56e7285436bc9154b37f7f86b#r27423260

🤷‍♂️

0xadada:
https://github.com/0xadada/ember-simple-auth/commit/d8a825c9c98009655dfe9e728b979a9c2db2eb15

ember-simple-auth internal `session` service is undefined. 
screen shot 2018-02-09 at 6 31 55 am

0xadada:
@rwjblue [like this](https://github.com/0xadada/ember-simple-auth/commit/dddf8da62f89ddc04655b150d2e2b66da0294eb3#diff-f3800bba3aa2a8fce48351e2ffbe2298R46)?

```js
import Application from '../app';
import { setApplication } from '@ember/test-helpers';

setApplication(Application.create({ autoboot: false }));
```
`Error: Assertion Failed: You cannot make a new Ember.Application using a root element that is a descendent of an existing Ember.Application`

rwjblue:
Yes, but we also need to use the imported config (to set the root element to `#ember-testing`).

https://github.com/ember-cli/ember-addon-output/blob/master/tests/dummy/config/environment.js#L42-L43
https://github.com/ember-cli/ember-addon-output/blob/master/tests/test-helper.js#L6

---

If you are still hitting that error, what is the stack?

0xadada:
screen shot 2018-02-10 at 3 05 00 pm
at that debugger, the stack for `error` is
```
"Error: Assertion Failed: You cannot use the same root element (DIV) multiple times in an Ember.Application
    at new EmberError (http://localhost:4200/assets/vendor.js:24799:25)
    at assert (http://localhost:4200/assets/vendor.js:25042:15)
    at Class.setup (http://localhost:4200/assets/vendor.js:56214:86)
    at Class.setupEventDispatcher (http://localhost:4200/assets/vendor.js:22689:18)
    at Class._bootSync (http://localhost:4200/assets/vendor.js:22644:14)
    at Class.didBecomeReady (http://localhost:4200/assets/vendor.js:23396:20)
    at invoke (http://localhost:4200/assets/vendor.js:20667:24)
    at Queue.flush (http://localhost:4200/assets/vendor.js:20587:25)
    at DeferredActionQueues.flush (http://localhost:4200/assets/vendor.js:20740:31)
    at Backburner.end (http://localhost:4200/assets/vendor.js:20872:42)"
```

reproduce
1. checkout https://github.com/0xadada/ember-simple-auth/tree/new-ember-qunit-release--qunit-test-in-mocha
2. `yarn && ember s` -> localhost:4200/tests
3. open the inspector

0xadada:
I accidentally closed this pR, but new PR is opened here https://github.com/simplabs/ember-simple-auth/pull/1536

PR #1521, opened 2018-01-19
Update mocha

cibernox:


cibernox:
I believe the error in CI is random. Can anyone trigger a rebuild to check it? =

marcoow:
I restarted the job 👍

marcoow:
Hm - seems to fail consistently. It's a bit hard to find out what's actually happening as the log is so long that you only see it completely when you view it raw where it's much harder to read 😞 

The `yarn.lock` diff is huge though - did you modify `package.json` manually and run `yarn` or did you run `yarn upgrade ember-cli-mocha@^0.15.0-beta.1` (which is the correct one)?

marcoow:
Nevermind - we're [ignoring `yarn.lock` on Travis anyway](https://github.com/simplabs/ember-simple-auth/blob/master/.travis.yml#L29). We should still keep the `yarn.lock` change minimal though.

cibernox:
I made the changes to yarn.lock minimal. But IDK if the failure is legitimate tho. I don't see why it would fail in Ember 1.12. That is really ancient.

marcoow:
Hm, seems to be related to ember-bootstrap actually:

```
[31mBuild failed.[39m
[31mTemplate Compiler Error (TemplateCompiler) in ember-bootstrap/templates/components/bs-accordion/body.hbs[39m

[31mCannot read property 'original' of undefined[39m

[31m[39m
[31mStack Trace and Error Report: /tmp/error.dump.b7bc0f062f2a472119805a62e262fabb.log[39m
[2K[1G[31merror[39m Command failed with exit code 1.
[2K[1G[34minfo[39m Visit [1mhttps://yarnpkg.com/en/docs/cli/run[22m for documentation about this command.
[34m
Result: false[39m
```

Yes, 1.12 is super ancient but for now we still support it 😞. The plan is to relase ESA 2.0 which drops support for anything < Ember 2.0 once Ember 3.0 is released.

cibernox:
I'll investigate it tomirrow.

PR #1522, opened 2018-01-19
Add an empty line

cibernox:
The intent is to verify that the failing tests in other branches is also in master.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,6 +20,7 @@ authorization mechanisms__.
 
 **Basic Information**
 
+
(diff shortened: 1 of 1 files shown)


cibernox:
@marcoow This is also failing to the error is also in master. It seems that it's some sort of transitive dependency thing.

marcoow:
Yep, seems to be caused by ember-bootstrap. I'll fix it.

cibernox:
This can be closed now

Issue #979, opened 2016-05-14
Deprecate Authorizers


marcoow:
Authorizers are a bit too complex and the implementation is not great currently. While they seem to offer a lot of flexibility they actually don't deliver as they are quite tightly coupled to the authenticator as they depend on values being retrieved by them. Instead we could probably provide a better mechanism for adding for injecting headers into outgoing requests as that's the use care in a majority of the projects out there. I'm thinking about an API somewhat like this for Ember Data adapters, e.g.:

``` js
import JSONAPIAdapter from 'ember-data/adapters/json-api';
import DataAdapterMixin from 'ember-simple-auth/mixins/data-adapter-mixin';
import config from '../config/environment';

const { comuted, inject: { service } } = Ember;

export default JSONAPIAdapter.extend(DataAdapterMixin, {
  session: service(),
  host: config.apiHost,
  headers: computed('session.data.authenticated.token', function() {
    let token = this.get('session.data.authenticated.token');
    return { Authorization: `Bearer ${token}` };
  })
});
```

I'm not sure whether Ember Data currently supports `headers` being a computed property but I guess that could probably be added. It might be cool if it could even support it returning a promise as there's the use case that a single use token for each request might be needed that would have to be retrieved as sort of a pre-flight request in which case the computed property would return a promise that resolves with the header then.

/cc @pangratz

srvance:
Ember Data does support computed properties for `headers`. What value does DataAdapterMixin add in that case? If `headers` would have to be declared in each adapter, this seems it would lead to a proliferation of boilerplate in cases where you had model-specific adapters. This is particularly bad when you have multiple data sources as ED has no layer of adapter between application and model.

marcoow:
When you're using multiple adapters in the vast majority of cases they share a common base adapter. If that's not the case you could use a mixin to share the code that sets the custom headers.

srvance:
I agree. If you're writing them all yourself, that's certainly an option.

I've recently been working with `ember-data-github` which tries to supply it's own token-passing mechanism based on a `headers` CP. I could monkey patch its adapter base class in an initializer or extend the model adapters adding an initializer in each one. The former is a bit ugly and the latter a bit laborious. I would see the latter as even worse without authorizers.

While the abstraction of an authorizer doesn't add a lot, I see it as useful for these kinds of cases. I would postulate that authorizers haven't been used very much because of a lack of documentation more than a bad design.

marcoow:
[...]

PR #1532, opened 2018-01-31
Deprecate Authorizers

stevenwu:
- [x] deprecate `session#authorize`
- [x] tests
- [x] add examples

#979


tests/unit/services/session-test.js
```diff
@@ -2,13 +2,25 @@ import ObjectProxy from '@ember/object/proxy';
...
+import { registerDeprecationHandler } from '@ember/debug';
 import { describe, beforeEach, it } from 'mocha';
 import { expect } from 'chai';
 import sinon from 'sinon';
 import Session from 'ember-simple-auth/services/session';
 
 import createWithContainer from '../../helpers/create-with-container';
 
+let warnings;
+registerDeprecationHandler((message, options, next) => {
```
marcoow: Should this be called in the respective `beforeEach` (or the `test` even)?

tests/unit/authorizers/oauth2-bearer-test.js
```diff
@@ -2,17 +2,34 @@ import { describe, beforeEach, it } from 'mocha';
 import { expect } from 'chai';
 import sinon from 'sinon';
 import OAuth2BearerAuthorizer from 'ember-simple-auth/authorizers/oauth2-bearer';
+import { registerDeprecationHandler } from '@ember/debug';
+
+let warnings;
+registerDeprecationHandler((message, options, next) => {
```
marcoow: Should this be called in the respective `beforeEach` (or the `test` even)?

tests/unit/authorizers/devise-test.js
```diff
@@ -2,17 +2,34 @@ import { describe, beforeEach, it } from 'mocha';
 import { expect } from 'chai';
 import sinon from 'sinon';
 import Devise from 'ember-simple-auth/authorizers/devise';
+import { registerDeprecationHandler } from '@ember/debug';
+
+let warnings;
+registerDeprecationHandler((message, options, next) => {
```
marcoow: Should this be called in the respective `beforeEach` (or the `test` even)?

README.md
```diff
@@ -502,6 +502,47 @@ export default Base.extend({
 });
 ```
 
+### Upgrading to 2.0
```
marcoow: As 2.0 isn't released yet, I'd maybe call this _"Deprecation of Authorizers"_ or so. We should also add a warning and a link to this section at the top of the _"Authorizers"_ section.

README.md
```diff
@@ -502,6 +502,47 @@ export default Base.extend({
...
+
+Starting from Ember Simple Auth 2.0, authorizers and the session service's `authorize` will be deprecated. You will need to configure your data adapter to include authorization data in request headers. Refer to the [Ember Guides](https://guides.emberjs.com/v3.0.0/models/customizing-adapters/#toc_headers-customization) for details on header configuration.
```
marcoow: I'd maybe not make this paragraph so specific to Ember Data (as authorizers could be used without Ember Data as well). Also authorizers are deprecated before 2.0 already and will be removed from 2.0. So maybe something like this

_"Authorizers and the session service's `authorize` method are deprecated and will be removed from Ember Simple Auth 2.0. The concept seemed like a good idea in the early days of Ember Simple Auth but proved to provide little value for quite some added complexity. In order to replace authorizers in an application, simply get the session data from the session service and inject it where it's needed manually."_

_"In most cases, authorizers are currently used together with Ember Data adapters (refer to the [Ember Guides](https://guides.emberjs.com/v3.0.0/models/customizing-adapters/#toc_headers-customization) for details on adapters). Replacing authorizers in these scenarios is pretty straight forward:"_

(also, max 80 chars line length please 🙏 )

addon/authorizers/base.js
```diff
@@ -18,9 +19,20 @@ import EmberObject from '@ember/object';
...
   @public
 */
 export default EmberObject.extend({
+  init() {
+    deprecate(`Ember Simple Auth: Authorizers are deprecated in favour of setting headers directly.`,
+      false,
+      {
+        id: 'ember-simple-auth.baseAuthorizer',
+        until: '2.0.0',
+        url: 'https://github.com/simplabs/ember-simple-auth#authorizers'
```
marcoow: I think we should link to the new section with the deprecation notice here.

README.md
```diff
@@ -502,6 +502,47 @@ export default Base.extend({
...
+      let { email, token } = this.get('session.data.authenticated');
+      let authData = `token="${token}", email="${email}"`;
+
+      return {
+        Authorization: `Token ${authData}`,
+      };
+    }
+  ),
+});
+```
```
marcoow: Have you checked whether these examples actually work yet? I'm not sure they do and we might have to add a new method to the `DataAdapterMixin`.

stevenwu: @marcoow The examples don't work 🙂.

What method do you want to add to `DataAdapterMixin`? This works.
```js
// data-adapter-mixin.js
ajaxOptions() {
    let { beforeSend } = hash;
    hash.beforeSend = (xhr) => {
    xhr.setRequestHeader('Authorization', this.get('authHeaderValue'));
    if (beforeSend) {
    ...
}
```
```js
// adapters/application.js
authHeaderValue: computed('session.data.authenticated.access_token', function() {
    let { access_token } = this.get('session.data.authenticated');
    return `Bearer ${access_token}`;
}),
```

marcoow: I think the best thing to do for now is to add a generic `authorize` method that receives the request as an argument:

```js
// data-adapter-mixin.js
ajaxOptions() {
  let { beforeSend } = hash;
  hash.beforeSend = (xhr) => {
    this.authorize(xhr);
    if (beforeSend) {
    ...
}
```

```js
// adapters/application.js
authHeaderValue: computed('session.data.authenticated.access_token', function() {
  let { access_token } = this.get('session.data.authenticated');
  xhr.setRequestHeader('Authorization', `Bearer ${access_token}`);
}),
```

That way we'd allow multiple headers, passing the token via the query string etc.

README.md
```diff
@@ -502,6 +504,47 @@ export default Base.extend({
...
+In most cases, authorizers are used with Ember Data adapters (refer to the
+[Ember Guides](https://guides.emberjs.com/v3.0.0/models/customizing-adapters/#toc_headers-customization)
+for details on adapters). Replacing authorizers in these scenarios is
+straightforward.
+
+Examples:
+
+```js
+// OAuth 2
+export default DS.JSONAPIAdapter.extend({
```
marcoow: This would also have to use the `DataAdapterMixin` which calls the `authorize` method.

README.md
```diff
@@ -502,6 +504,47 @@ export default Base.extend({
...
+  authorize(xhr) {
+    let { access_token } = this.get('session.data.authenticated');
+    xhr.setRequestHeader('Authorization', `Bearer ${access_token}`);
+  }
+});
+```
+
+```js
+// Devise
+export default DS.JSONAPIAdapter.extend({
```
marcoow: Like the above example, this would have to use the `DataAdapterMixin` as well.

README.md
```diff
@@ -502,6 +504,47 @@ export default Base.extend({
...
+[Ember Guides](https://guides.emberjs.com/v3.0.0/models/customizing-adapters/#toc_headers-customization)
+for details on adapters). Replacing authorizers in these scenarios is
+straightforward.
+
+Examples:
+
+```js
+// OAuth 2
+export default DS.JSONAPIAdapter.extend({
+  session: service('session'),
```
marcoow: The `DataAdapterMixin` injects the `session` service already but I guess it makes sense to leave it in here to avoid confusion.

addon/authorizers/base.js
```diff
@@ -18,9 +19,20 @@ import EmberObject from '@ember/object';
   @class BaseAuthorizer
   @module ember-simple-auth/authorizers/base
   @extends Ember.Object
+  @deprecated BaseAuthorizer:class
   @public
 */
 export default EmberObject.extend({
+  init() {
```
marcoow: This should call `_super`:

```js
init() {
  this._super(...arguments);

  deprecate(…
```

addon/mixins/data-adapter-mixin.js
```diff
@@ -75,23 +75,30 @@ export default Mixin.create({
...
+
       if (beforeSend) {
         beforeSend(xhr);
       }
     };
     return hash;
   },
 
+  authorize() {
+    assert('The `authorize` method should be overridden in your application adapter. It should accept a single argument, the xhr response.');
```
marcoow: _"the ~~xhr response~~ request object"_

tests/unit/mixins/data-adapter-mixin-test.js
```diff
@@ -43,10 +43,11 @@ describe('DataAdapterMixin', () => {
...
     });
 
-    it('asserts the presence of authorizer', function() {
+    it('asserts `authorize` is overridden', function() {
       adapter.set('authorizer', null);
       expect(function() {
         adapter.ajaxOptions();
+        hash.beforeSend();
       }).to.throw(/Assertion Failed/);
     });
```
marcoow: I think we should have a test for `authorize` actually being called when a request is made.

tests/unit/authorizers/devise-test.js
```diff
@@ -2,17 +2,32 @@ import { describe, beforeEach, it } from 'mocha';
...
   beforeEach(function() {
+    registerDeprecationHandler((message, options, next) => {
+      // in case a deprecation is issued before a test is started
+      if (!warnings) {
+        warnings = [];
+      }
+
+      warnings.push(message);
+      next(message, options);
+    });
```
marcoow: Would it make sense to move this into the test instead as it's only needed for that one test anyway (same thing applies below for the Devise test of course)?

stevenwu: The handler isn't registered early enough when it's setup inside the test, but I'll see if there are other options.

marcoow: I think you'd just have to move creation of the authorizer into the test as well (or just create another one in the test)?

addon/authorizers/base.js
marcoow: 👍

tests/unit/authorizers/devise-test.js
marcoow: 👍

tests/unit/mixins/data-adapter-mixin-test.js
marcoow: 👍

Turbo87:
OMG, finally! 🙏  :D

thanks @stevenwu!

Issue #1531, opened 2018-01-30
implement support for new style qunit API ember-qunit test hooks


0xadada:
If a consuming application is using `ember-cli-qunit` [4.2.0 or greater](https://github.com/ember-cli/ember-cli-qunit/blob/master/CHANGELOG.md#v420-2017-12-17), they may want to migrate to the
[more modern testing syntax](https://dockyard.com/blog/2018/01/11/modern-ember-testing). In that case, they will need to be imported from the `ember-simple-auth` addon namespace rather than from the consuming application namespace.

```js
// tests/acceptance/…
import { currentSession, authenticateSession, invalidateSession} from 'ember-simple-auth/test-support';
```

**Note** The helpers imported from the addon namespace (i.e. `import { functionName } from 'ember-simple-auth/test-support';`) differ slightly from those in the consumer application namespace in that they should not take the `application` argument.

The new-style helpers have the following function signatures:
* `currentSession()` returns the current session.
* `authenticateSession(sessionData)` authenticates the session asynchronously; the optional `sessionData` argument can be used to mock an authenticator response (e.g. a token or user).
* `invalidateSession()` invalidates the session asynchronously.

New tests using `authenticateSession` helper will look like this:

```js
import { module, test } from 'qunit';
import { setupApplicationTest } from 'ember-qunit';
import { currentURL, visit } from '@ember/test-helpers';
import { authenticateSession } from 'ember-simple-auth/test-support';

module('Acceptance | super secret url', function(hooks) {
  setupApplicationTest(hooks);

  test('authenticated users can visit /super-secret-url', async function(assert) {
    await authenticateSession({
      userId: 1,
      otherData: 'some-data'
    });
    await visit('/super-secret-url');
    assert.equal(currentURL(), '/super-secret-url', 'user is on super-secret-url');
  });
});
```

Thanks to  @cibernox for the help

Related https://github.com/simplabs/ember-simple-auth/issues/1512
Related https://github.com/simplabs/ember-simple-auth/issues/1530

villander:
@0xadada i try this solution, but without success. I'm writing this test in an addon

```

[...]

PR #1536, opened 2018-02-10
[#1512] Added new ember-qunit 4.3.0+ test helper

0xadada:
If a consuming application is using `ember-cli-qunit` [4.2.0 or greater](https://github.com/ember-cli/ember-cli-qunit/blob/master/CHANGELOG.md#v420-2017-12-17), they may want to migrate to the
[more modern testing syntax](https://dockyard.com/blog/2018/01/11/modern-ember-testing). In that case, they will need to be imported from the `ember-simple-auth` addon namespace rather than from the consuming application namespace.

```js
// tests/acceptance/…
import { currentSession, authenticateSession, invalidateSession} from 'ember-simple-auth/test-support';
```

**Note** The helpers imported from the addon namespace (i.e. `import { functionName } from 'ember-simple-auth/test-support';`) differ slightly from those in the consumer application namespace in that they should not take the `application` argument.

The new-style helpers have the following function signatures:
* `currentSession()` returns the current session.
* `authenticateSession(sessionData)` authenticates the session; the optional `sessionData` argument can be used to mock an authenticator response (e.g. a token or user).
* `invalidateSession()` invalidates the session.

Based on suggestions by @cibernox added a new modern style helper, that
will be exported to the ember-simple-auth namespace

* [x] New test helper
* [x] Updated README.md to explain new helper
* [x] updated dependencies
* [x] Updated test-helper to use `setApplication`

Closes https://github.com/simplabs/ember-simple-auth/issues/1531
Fixes https://github.com/simplabs/ember-simple-auth/issues/1512

thanks @rwjblue @cibernox @marcoow @Turbo87 for helping me land this PR!

### DIFF
--- a/addon-test-support/index.js
+++ b/addon-test-support/index.js
@@ -1,0 +1,61 @@
+import { get } from '@ember/object';
+import { getContext, settled } from '@ember/test-helpers';
+import Promise from 'rsvp';
+import Test from 'ember-simple-auth/authenticators/test';
+
+const SESSION_SERVICE_KEY = 'service:session';
+const TEST_CONTAINER_KEY = 'authenticator:test';
+
+function ensureAuthenticator(owner) {
+  const authenticator = owner.lookup(TEST_CONTAINER_KEY);
+  if (!authenticator) {
+    owner.register(TEST_CONTAINER_KEY, Test);
+  }
+}
+
+/**
+ * Authenticates the session.
+ *
+ * @param {Object} sessionData Optional argument used to mock an authenticator
+ * response (e.g. a token or user).
+ * @return {Promise}
+ * @public
+ */
+export function authenticateSession(sessionData) {
+  const { owner } = getContext();
+  const session = owner.lookup(SESSION_SERVICE_KEY);
+  owner.setupRouter(); // router must initialize fully before authentication
+  ensureAuthenticator(owner);
+  return session.authenticate(TEST_CONTAINER_KEY, sessionData).then(() => {
+    return settled();
+  });
+}
+
+/**
+ * Returns the current session.
+ *
+ * @return {Object} a session service.
+ * @public
+ */
+export function currentSession() {
+  const { owner } = getContext();
+  return owner.lookup(SESSION_SERVICE_KEY);
+}
+
+/**
+ * Invalidates the session.
+ *
+ * @return {Promise}
+ * @public
+ */
+export function invalidateSession() {
+  const { owner } = getContext();
+  const session = owner.lookup(SESSION_SERVICE_KEY);
+  const isAuthenticated = get(session, 'isAuthenticated');
+  return Promise.resolve().then(() => {
+    if (isAuthenticated) {
+      return session.invalidate();
+    }
+  })
+    .then(() => settled());
+}
--- a/tests/unit/initializers/setup-session-test.js
+++ b/tests/unit/initializers/setup-session-test.js
@@ -19,6 +19,7 @@ describe('setupSession', () => {
       register() {},
       injection() {}
     };
+    Ember.testing = true;
   });
 
   it('registers the session', function() {
--- a/README.md
+++ b/README.md
@@ -646,21 +646,68 @@ in `package.json`:
 ## Testing
 
 Ember Simple Auth comes with a __set of test helpers that can be used in
-acceptance tests__:
+acceptance tests__.
 
-* `currentSession(app)`: returns the current session.
-* `authenticateSession(app, sessionData)`: authenticates the session; the
-  optional `sessionData` argument can be used to mock an authenticator
-  response - e.g. a token.
-* `invalidateSession(app)`: invalidates the session.
+### ember-cli-qunit 4.2.0 and greater
 
-The test helpers can be imported from the `helpers/ember-simple-auth`
-module in the application's namespace:
+If your app is using `ember-cli-qunit` [4.2.0 or
+greater](https://github.com/ember-cli/ember-cli-qunit/blob/master/CHANGELOG.md#v420-2017-12-17),
+you may want to migrate to the [more modern testing
+syntax](https://dockyard.com/blog/2018/01/11/modern-ember-testing). In that
+case, helpers can be imported from the `ember-simple-auth` addon namespace.
+
+```js
+// tests/acceptance/…
+import { currentSession, authenticateSession, invalidateSession} from 'ember-simple-auth/test-support';
+```
+
+The new-style helpers have the following function signatures:
+* `currentSession()` returns the current session.
+* `authenticateSession(sessionData)` authenticates the session asynchronously;
+  the optional `sessionData` argument can be used to mock an authenticator
+  response (e.g. a token or user).
+* `invalidateSession()` invalidates the session asynchronously.
+
+New tests using the async `authenticateSession` helper will look like this:
+
+```js
+import { module, test } from 'qunit';
+import { setupApplicationTest } from 'ember-qunit';
+import { currentURL, visit } from '@ember/test-helpers';
+import { authenticateSession } from 'ember-simple-auth/test-support';
+
+module('Acceptance | super secret url', function(hooks) {
+  setupApplicationTest(hooks);
+
+  test('authenticated users can visit /super-secret-url', async function(assert) {
+    await authenticateSession({
+      userId: 1,
+      otherData: 'some-data'
+    });
+    await visit('/super-secret-url');
+    assert.equal(currentURL(), '/super-secret-url', 'user is on super-secret-url');
+  });
+});
+```
+
+### ember-cli-qunit 4.1.0 and earlier
+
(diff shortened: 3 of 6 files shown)


rwjblue:
The newer style helpers do not work before Ember 2.4, I believe that you need to guard those tests (preventing them from running on 1.13 and 2.0).

For example [this](https://github.com/emberjs/ember-test-helpers/blob/7a92c721f8e55096a7bdb217e7cd294102aa82f8/tests/unit/setup-context-test.js#L25-L27) is what ember-test-helpers does to prevent its own test suite from failing on these older versions:

```js
import hasEmberVersion from 'ember-test-helpers/has-ember-version';

// ...snip...

describe('.....', function() {
  if (!hasEmberVersion(2, 4)) {
    return;
  }

  // ....rest of the tests here...
});
```

0xadada:
@rwjblue now guard against testing releases before 2.4 in 
https://github.com/simplabs/ember-simple-auth/pull/1536/commits/8dd43903b4a492fd7f3529103c49c83019b7a5d4

addon-test-support/index.js
```diff
@@ -0,0 +1,61 @@
...
+ *
+ * @param {Object} sessionData Optional argument used to mock an authenticator
+ * response (e.g. a token or user).
+ * @return {Promise}
+ * @public
+ */
+export function authenticateSession(sessionData) {
+  const { owner } = getContext();
+  const session = owner.lookup(SESSION_SERVICE_KEY);
+  owner.setupRouter(); // router must initialize fully before authentication
```
rwjblue: The prior versions of `authenticateSession` didn't do this, so I'm not 100% sure why this is needed.  Can you try to remove it and see what happens?

tests/acceptance/authentication-qunit-test.js
```diff
@@ -0,0 +1,91 @@
...
+
+  describe('the protected route', function() {
+    if (!hasEmberVersion(2, 4)) {
+      // guard against running test module on unsupported version (before 2.4)
+      return;
+    }
+
+    it('cannot be visited when the session is not authenticated', function() {
+      return invalidateSession().then(function() {
+        return visit('/protected').then(() => {
```
rwjblue: Can probably detangle these few tests a tad bit by returning early, something like:

```js
    it('cannot be visited when the session is not authenticated', function() {
      return invalidateSession()
        .then(() => visit('/protected'))
        .then(() => {
          expect(currentURL()).to.eq('/login');
        });
      });
    });
```

addon-test-support/index.js
0xadada: I added `setupRouter` because visits (`visit()`) to routes protected by `AuthenticatedRouteMixin` fail without it.

TL;DR `hasRoute` is undefined when the authenticationSucceeded event is triggered by the `ApplicationRouteMixin`. I did a [full writeup of the series of calls](https://github.com/simplabs/ember-simple-auth/pull/1519#issuecomment-359837396) describing what occurs on the previous PR. @jordpo found a workaround was to invoke `visit` first to setup the router, but by calling `setupRouter` in the test-helper, it would initialize the routerMicroLib and fix the problem.

0xadada:
@rwjblue looks nicer this way https://github.com/simplabs/ember-simple-auth/pull/1536/commits/872d97cd3f7b8d3b6a18c5f0607b2758b6a08cc0

addon-test-support/index.js
rwjblue: Thank you for pointing me at that writeup!  Ultimately, this still seems like a bug being identified. IMHO, it is not logical for the `ApplicationRouteMixin` to call `.transitionTo` before the first navigation (e.g. before the first `visit()` call)...

/cc @marcoow @Turbo87

marcoow: > IMHO, it is not logical for the ApplicationRouteMixin to call .transitionTo before the first navigation (e.g. before the first visit() call)...

Yep, that's true. The problem is the [`ApplicationRouteMixin`'s `sessionAuthenticated` method](http://ember-simple-auth.com/api/classes/ApplicationRouteMixin.html#method_sessionAuthenticated) is called when the session is authenticated. Whether it makes sense to always do that automatically is certainly something that's debatable but that's how it is for now and we can't change that for now. We could maybe add a check into that method to just bail out early when the router is not ready (which I assume would require using private API to find out as well though 😞)

0xadada:
@marcoow The `setupRouter` function already does that, so I opted not to add the additional complexity.

-------- Original Message --------

addon-test-support/index.js
0xadada: @marcoow The `setupRouter` function already does that, so I opted not to add the additional complexity.

marcoow: @0xadada: what I meant is we can add a check for whether the router is ready [here](https://github.com/simplabs/ember-simple-auth/blob/1.4.2/addon/mixins/application-route-mixin.js#L108) and just return if it is not.

tests/acceptance/authentication-qunit-test.js
```diff
@@ -0,0 +1,91 @@
...
+import { expect } from 'chai';
+import Pretender from 'pretender';
+import {
+  invalidateSession,
+  authenticateSession,
+  currentSession
+} from 'ember-simple-auth/test-support';
+import config from '../../config/environment';
+
+describe('Acceptance: Authentication (for QUnit helpers)', function() {
```
cibernox: Just a nit-pick, but those helpers are not QUnit specific (and as a matter of fact, in there they are working in mocha).
I'd name the module something different, like "New-style Test Helpers"

addon-test-support/index.js
cibernox: Is this the last loose end?

marcoow: Yes, this is the last loose end. I guess we can leave it as is and see whether we can refactor the `ApplicationRouteMixin` separately so the error situation would never occur - that should be done in a separate PR anyways.

rwjblue: Yes, totally agreed @marcoow. Calling `setupRouter` here should be fine for now. 

Let's open an issue to track refactoring the mixin so this isn't needed though...

tests/acceptance/authentication-qunit-test.js
0xadada: @cibernox I'll rename the file and `describe('Acceptance: New-style Test Helpers)'.

0xadada:
Y'all, it feels so great to contribute back to the community. I feel like i've actually earned the ESA sticker i got at Ember Conf last year.

screen shot 2018-02-13 at 1 31 49 pm

Thanks to everyone who helped me land this PR, such a great community.
🎉

marcoow:
Thanks so much for the effort @0xadada! This wasn't easy but I'm glad it's in now (once the build completes). Please say hi in Portland if you're there!

addon-test-support/index.js
marcoow: follow-up issue is here: https://github.com/simplabs/ember-simple-auth/issues/1539

marcoow:
…and the build failed 😞 I restarted it but it doesn't look like a random failure…

marcoow:
The problem is that `Ember.testing` is `false` [here](https://github.com/simplabs/ember-simple-auth/blob/master/addon/initializers/setup-session.js#L10).

0xadada:
@marcoow i'm confused about what changed between yesterday and today that would've suddenly started causing this?

marcoow:
This must be some change in an upstream dependency. We ignore the `yarn.lock` file on CI so you'll get the latest of everything which sometimes can lead to errors.

0xadada:
@marcoow  the same error is failing on the `master` branch on my local system as well

0xadada:
looks like once [my tests `tearDownContext(context)`](https://github.com/simplabs/ember-simple-auth/pull/1536/files#diff-44b9f882bba69fbf5ee9455fdc5c781dR37) runs, it invokes [`@ember/test-helpers#tearDownContext`](https://github.com/emberjs/ember-test-helpers/blob/master/addon-test-support/%40ember/test-helpers/teardown-context.js#L30) which sets `Ember.testing = false`, and borks the other tests that use that.

tests/unit/initializers/setup-session-test.js
```diff
@@ -19,6 +19,7 @@ describe('setupSession', () => {
       register() {},
       injection() {}
     };
+    Ember.testing = true;
```
0xadada: Not really feelin' this, but `tearDownContext()` in the `new-style-helpers-test` blows this test up.

rwjblue: Seems fine to me. I do dislike that Ember-test-helpers has to do that...

marcoow:
So here we go - thanks for all the hard work @0xadada!

marcoow:
This has been released with the 1.5.0 release: https://github.com/simplabs/ember-simple-auth/releases/tag/1.5.0

Issue #1539, opened 2018-02-13
Do not attempt to transition before routing is ready


marcoow:
Currently, the `ApplicationRouteMixin` [maps the session service's events to methods on the mixin (and thus the application route) in `init`](https://github.com/simplabs/ember-simple-auth/blob/master/addon/mixins/application-route-mixin.js#L79). This leads to the problem that when the session is authenticated in an acceptance test before any route was visited (and the routing system is ready), the session service's event will be triggered which will cause a method on the application route to be called which will attempt a transition although the routing system is not ready yet. See [this writeup of the series of steps by @0xadada](https://github.com/simplabs/ember-simple-auth/pull/1519#issuecomment-359837396).

We could maybe only map the session service's events to methods on the application route (via the `ApplicationRouteMixin`) in the `activate` method (as opposed to `init`) which I assume would only be called after the routing system is ready but that might potentially lead to errors in edge cases… This needs some further investigation but is generally something that should be cleaned up.

alexander-alvarez:
Spent the last couple of days researching this because it's affecting our application.

How can I help?

0xadada:
@alexander-alvarez  what does your test look like?

alexander-alvarez:
@0xadada https://github.com/alexander-alvarez/ember-cli-qunit-service-registering/compare/572fae2#diff-6902ce6b3192309e4080d20541f6fbfd

marcoow:
This should not happen when you're using the new acceptance test helpers introduced with 1.5.0. In order to fix this, the first thing to try would be moving the [mapping of session service events from the `init` method](https://github.com/simplabs/ember-simple-auth/blob/master/addon/mixins/application-route-mixin.js#L79) to the `activate` method I think - if anyone wants to jump in and try that.

0xadada:
@alexander-alvarez I'm confused. Your test component invalidates the session upon `init` hook, then tests that `yes` is output, when `no` _should_ be output because the session is not authenticated (because you called `invalidate`).

0xadada:
@alexander-alvarez per @marcoow, try the new test helper:
`tests/integration/components/my-component-test.js`:
```js
import { authenticateSession } from 'ember-simple-auth/test-support';
```
then
```js
-this.owner.register('service:session', Service.extend({
-      isAuthenticated: true,
-      invalidate() {
-        assert.ok(true);
-    }
-}));
+authenticateSession({
+        userId: 1,
+        otherData: 'some-data'
+});
[...]

PR #1542, opened 2018-02-15
[BUGFIX]: Subscribing to session events in activate, not init

alexander-alvarez:
Closes https://github.com/simplabs/ember-simple-auth/issues/1539

**/cc** @marcoow  @0xadada 

I applied the fix to my addon and it worked :+1: 
https://github.com/alexander-alvarez/ember-cli-qunit-service-registering/compare/test-case-failing

Let me know if there are other test cases we can introduce into this repository

### DIFF
--- a/addon/mixins/application-route-mixin.js
+++ b/addon/mixins/application-route-mixin.js
@@ -74,7 +74,7 @@ export default Mixin.create({
     return Configuration.routeAfterAuthentication;
   }),
 
-  init() {
+  activate() {
     this._super(...arguments);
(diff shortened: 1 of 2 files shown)


alexander-alvarez:
The fastboot tests that failed in CI, passed (twice) on my local

marcoow:
Debugging FastBoot tests is a pain unfortunately but this needs to be investigated. Sadly not being able to reproduce locally doesn't make this easier 😞

marcoow:
#1547 fixed this with a different approach.

Issue #1543, opened 2018-02-15
Allow new test helpers be used in rendering tests


alexander-alvarez:
From what I can tell [`owner.setupRouter()`](https://github.com/simplabs/ember-simple-auth/blob/master/addon-test-support/index.js#L27) is the only thing preventing this from working in rendering tests.

What if we did this instead?
```js
if (owner.setupRouter) {
  owner.setupRouter();
}
```

Thoughts?
@0xadada

0xadada:
Thats literally what we do in the new test helper
https://github.com/simplabs/ember-simple-auth/blob/master/addon-test-support/index.js#L27

alexander-alvarez:
@0xadada I'm not following. I'm proposing we avoid calling `setupRouter` if the `owner.setupRouter` doesn't exist as it doesn't when you use `setupRenderingTest`

0xadada:
ah, i follow. @alexander-alvarez that seems reasonable

marcoow:
I'm not sure this is a good idea tbh. The actual session service is interconnected with routes, session stores etc. - all of which you don't want to become active in rendering tests (ake component integration tests). I think the way to go is to mock the session service in these tests.

alexander-alvarez:
@marcoow  I wanted to propose the idea since it wasn't too much work to implement and it seemed to me that the session service didn't have anything explicitly tying it to other pieces.

perfectly fine mocking out the session service as an alternative, but I need help pushing forwards https://github.com/simplabs/ember-simple-auth/pull/1542 as an alternative because without either currently I'm SOL

marcoow:
Yes, we certainly should get #1542 (or something similar) in. That is really mostly internal cleanup though - if you use the new acceptance test helpers introduced in 1.5.0, the router not being set up should not be a problem actually.

I'm going ahead and close this one and #1542 certainly is on our list of things to look at as soon as someone finds the time!

alexander-alvarez:
> That is really mostly internal cleanup though - if you use the new acceptance test helpers introduced in 1.5.0, the router not being set up should not be a problem actually.

@marcoow I can't mock out the session service in integration/rendering tests because when `Application.create(config.APP)` (or it's equivalent) is called in the default `tests/test-helper.js`, the application gets created which in turn automatically does `this.get('session')` and resolves the real session service which is cached going forwards and resolved in the tests (see https://github.com/emberjs/ember-test-helpers/issues/324) instead of the mocked one

[...]

PR #1544, opened 2018-02-15
[FEATURE] Introduce 'setupSessionTest' test-helper

alexander-alvarez:
In response to https://github.com/simplabs/ember-simple-auth/issues/1543, this PR improves `authenticateSession` so it can be used outside of `ApplicationTest` contexts.

Also adds a new `setupSessionTest` helper, so that we can use all helpers in any context that has a properly setup owner (e.g. `setupTest`, `setupRenderingTest`), not just `ApplicationTests`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -658,7 +658,7 @@ case, helpers can be imported from the `ember-simple-auth` addon namespace.
 
 ```js
 // tests/acceptance/…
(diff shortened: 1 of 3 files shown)


marcoow:
Thanks for the effort but see https://github.com/simplabs/ember-simple-auth/issues/1543#issuecomment-366163839 - I don't think this should be added tbh.

PR #1547, opened 2018-02-16
Convert `setupSessionRestoration()` to regular initializer

Turbo87:
This should hopefully address https://github.com/simplabs/ember-simple-auth/issues/1543#issuecomment-366265082

Instead of eagerly looking up the application route instance and `session`:

- we reopen the application route factory instead of the instance
- we lookup the `session` once we actually needed it in the `beforeModel()` hook

Please note that I have not actually tried this in an app, but the acceptance tests seem pretty happy about the change. Feedback welcome!
 
/cc @rwjblue @alexander-alvarez


Turbo87:
turns out https://github.com/rwjblue/ember-getowner-polyfill only includes `getOwner()` but not `setOwner()` which I had used in the unit tests for the initializer function 🤔

Turbo87:
yehaa, it's 🍏

Turbo87:
@marcoow fixed

PR #1557, opened 2018-03-02
Ignore Docs in Published NPM Package

### LINKED ISSUE #1556: Documentation's CSS Being Loaded in Production
I noticed the following errors today on my greenfield project using v1.5.1:

screen shot 2018-03-02 at 1 12 40 pm

I tracked these down to usage in ESA's docs. Perhaps we ought to exclude the docs from the npm package?

jherdman:
Resolves #1556

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -7,6 +7,7 @@
 /fastboot-dist
 /fastboot-tests
 /guides
+/docs
 **/.gitkeep
 .bowerrc
 .editorconfig


jherdman:
If this is kosher, would it be possible to get a patch release with these changes to silence the ugly warnings noted in #1556?

marcoow:
Unfortunately this is not as straight forward as it would break our deployment of the dummy app to ember-simple-auth.now.sh (and maybe other things).

Turbo87:
since https://github.com/simplabs/ember-simple-auth/issues/1556 is closed I assume this is not actually necessary?

marcoow:
Unfortunately we cannot easily do this anyway as explained above…

PR #1565, opened 2018-03-15
add empty "ember-simple-auth! instance initializer

### LINKED ISSUE #1564: "No instance intializer named ember-simple-auth" Acting like it doesn't exist
I'm running into an issue I haven't seen before. On app launch I'm getting an 

`Assertion Failed: No instance initializer named 'ember-simple-auth'`

I've got it being called in one of my customer initializers, `app/intializers/*.js`.

```
import Ember from 'ember';

export function initialize(appInstance) {
	var router = appInstance.lookup('router:main');
	var segment = appInstance.lookup('service:segment');

	//-----------------------------------------
	// When pages transition, track view
	//-----------------------------------------
	router.on('didTransition', function() {
		Ember.run.debounce(this, function(){
			const routeName    = this.currentRouteName;
			const routeHandler = appInstance.lookup('route:' + routeName);

			const segmentName       = routeHandler.get('_segmentName');
			const segmentProperties = routeHandler.get('_segmentProperties');

			//-----------------------------------------
			// If we have an override, use it
			//-----------------------------------------
			if( segmentName ){
				segment.trackPageView(segmentName, segmentProperties);
				return;
			}

			//-----------------------------------------
			// If not just track normally
			//-----------------------------------------
			segment.trackPageView();
		}, 150);
	});
}

export default {
	name  : 'segment',
	after : 'ember-simple-auth',
	initialize
};

```
I'm new to Ember, and it probably shows, but after doing some google sleuthing I'm not further than I was hours ago. Any help is appreciated.

ember-cli - 2.7.0
ember-simple-auth - 1.5.1
ember - 2.7.0

marcoow:
#1547 removed the `ember-simple-auth` instance initializer but that is actually a breaking change as applications might depend on the existence of that initializer, e.g.

```js
export default {
  name  : 'segment',
  after : 'ember-simple-auth',
  initialize
};
```

### TODO

- [ ] figure out how we can only show a deprecation when any instance initializer in the application has a `after: 'ember-simple-auth'` or `before: 'ember-simple-auth'` but not if that's not the case.

closes #1564

### DIFF
--- a/app/instance-initializers/ember-simple-auth.js
+++ b/app/instance-initializers/ember-simple-auth.js
@@ -1,0 +1,8 @@
+// This is only needed for backwards compatibility and will be removed in the
+// next major release of ember-simple-auth. Unfortunately, there is no way to
+// deprecate this without hooking into Ember's internals…
+export default {
+  name: 'ember-simple-auth',
+
+  initialize() {}
+};


Turbo87:
> figure out how we can only show a deprecation when any instance initializer in the application has a after: 'ember-simple-auth' or before: 'ember-simple-auth' but not if that's not the case.

I'm not aware of any way to do that, might not be possible to do properly... I guess if we mention the initializer removal in the changelog for the next major that should be enough.

can you add a small code comment to the initializer that explain that it still exists for backwards compat reasons and should be removed for the next major release?

marcoow:
@Turbo87: I added a comment. I guess we could hook into Ember's internals but of course that'd be a horrible solution. It seems like there should be a `deprecated` property on initializers and instance initializers that would trigger Ember to throw a deprecation whenever the initializer or instance initializer is depended on.

PR #1598, opened 2018-05-07
Tweak authorizer deprecation example to be closer to original behavior

john-griffin:
Authorizers would previosuly omit the authorization header if no token was available. The original example in the readme caused some issues in certain apps that relied on this behavior.


john-griffin:
@marcoow thanks, updated with `isPresent` check.

PR #1604, opened 2018-05-16
Use proper argument spread to avoid non-transpiled issues

### LINKED ISSUE #1638: Arrow functions can't be bound
https://github.com/simplabs/ember-simple-auth/blob/620982458e8bc5debcd188e8c3b59bea7ebc1421/addon/mixins/application-route-mixin.js#L82-L90

According to spec, this should not work, nor should passing arguments that way. See http://www.ecma-international.org/ecma-262/6.0/#sec-arrow-function-definitions-runtime-semantics-evaluation

I think it _might_ work because the arrow functions are still transpiled by Babel, which is why I ran into this bit because we are in the lucky position to having a project where we don't have to support older browsers.

Possible correction:
```js
A([
    ['authenticationSucceeded', 'sessionAuthenticated'],
    ['invalidationSucceeded', 'sessionInvalidated'],
]).forEach(([event, method]) => {
    this.get('session').on(event, (...args) => {
        this[method](...args);
    });
});
```

rtablada:
Based on the arrow function spec, `arguments` is inherited just like `super` and `this`.
This means that the code here is actually spreading the arguments from `_subscribeToSessionEvents` instead of the actual event handler.

This could lead to unexpected results especially since older versions of Babel did not follow spec and allowed use of `...arguments.
As newer versions of Babel follow spec and arrow functions are not transpiled (due to targets), this causes different behavior across transpilation targets.

This PR fixes this.

---

I have also checked and didn't see any other offending `...argument` calls since this is the only one inside of an arrow function.

closes #1638

### DIFF
--- a/addon/mixins/application-route-mixin.js
+++ b/addon/mixins/application-route-mixin.js
@@ -1,6 +1,5 @@
 import Mixin from '@ember/object/mixin';
 import { A } from '@ember/array';
-import { bind } from '@ember/runloop';
 import { computed } from '@ember/object';
 import { getOwner } from '@ember/application';
 import { inject } from '@ember/service';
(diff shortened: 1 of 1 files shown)


rtablada:
I'm actually 🤔 that `bind` isn't needed at all here.

PR #1606, opened 2018-05-17
Add `Remember me` button to dummy login

stevenwu:


tests/dummy/app/templates/components/login-form.hbs
```diff
@@ -13,6 +13,12 @@
     <label for="password">Password</label>
     {{input value=password placeholder='Enter Password' class='form-control' type='password'}}
   </div>
+  <div class="checkbox">
+    <label>
```
jessica-jordan: If we add the `for="rememberMe"` attribute and the `elementId="rememberMe"` property to the label and the input respectively, we can also make this form more accessible by explicitly labelling the checkbox for screen readers (See also: https://www.w3.org/WAI/tutorials/forms/labels/)

stevenwu:
Thanks @jessica-jordan. `id` works so I used that instead.

Issue #1619, opened 2018-07-01
Use without mixins


mike-north:
Is there a plan for allowing consumption of ember-simple-auth in a codebase that bans or restricts use of mixins? Examples include a codebase making use of ES6 classes and/or TypeScript, where mixins are largely unsupported (and will be for the foreseeable future)

One ultra-low-effort solution to this problem might be to expose the functions on the object passed to `Mixin.create( )` for `ApplicationRouteMixin`, `AuthenticatedRouteMixin`, etc... for direct consumption.

```ts
import { beforeModel as simpleAuthBeforeModel } from 'ember-simple-auth/mixins/authenticated-route-mixin'

export default class extends Route {
   beforeModel() {
       super.beforeModel(...arguments);
       simpleAuthBeforeModel(this, a, b, c);
   }
}

```

geekygrappler:
I agree we should try and move away from mixins at some point.

marcoow:
The big advantage of mixins is that they make incorporating ESA's into apps very easy. Of course @mike-north's approach above works well but especially for mixins that add multiple methods (and properties) that's tedious and error-prone. We should investigate closely what options we have for dropping mixins without making integrating ESA much harder than it is today.

Just thinking out loud here but what if we just switched to classes?

```ts
import AuthenticatedRoute from 'ember-simple-auth/routes/authenticated'

export default class extends AuthenticatedRoute {
}
```

There are likely other alternatives…

Romanior:
One possible way, as well just thinking out loud, could be to wrap Route class with custom creation function, for more cleaner user's class and simple integration, for instance:

```javascript
import { createAuthenticatedRoute } from 'ember-simple-auth/routes/authenticated'

class ApplicationRoute extends Route {}

export default createAuthenticatedRoute(ApplicationRoute, { /* optional config */ });
```

mike-north:
> We should investigate closely what options we have for dropping mixins without making integrating ESA much harder than it is today.

I think we can find a way that is no more difficult, but possibly more explicit (a good thing imo) than just adding (mystery) mixins in several places. Let's aim for easiest to understand and maintain, and fight the temptation to optimize for smallest # of lines.

Something like this is easy for developers to understand, and they're free to use inheritance for their own needs instead of extending from some base class

```ts
import { redirectIfUnauthenticated } from 'ember-simple-auth/authenticated-route'

class HomeRoute extends Route {

   beforeModel(transition) {
       super.beforeModel(transition);
[...]

PR #1623, opened 2018-07-09
move fastboot CP into util

geekygrappler:
I was having a look at what I could do to help with #1619. I don't really get the full scope of how we could move away from mixins, but I did notice that `isFastBoot` method was repeated 5 times, so I moved it to a util function.

### DIFF
--- a/addon/utils/is-fastboot.js
+++ b/addon/utils/is-fastboot.js
@@ -1,0 +1,13 @@
+import { computed } from '@ember/object';
+import { getOwner } from '@ember/application';
+import { assert } from '@ember/debug';
+
+export default function isFastBoot() {
+  return computed(function() {
+    const container = getOwner(this);
+    assert('You may only use isFastBoot() on a container-aware object', container && typeof container.lookup === 'function');
+
+    const fastboot = container.lookup('service:fastboot');
+    return fastboot ? fastboot.get('isFastBoot') : false;
+  });
+}
--- a/addon/mixins/application-route-mixin.js
+++ b/addon/mixins/application-route-mixin.js
@@ -6,6 +6,7 @@ import { getOwner } from '@ember/application';
 import { inject } from '@ember/service';
 import Ember from 'ember';
 import Configuration from './../configuration';
+import isFastBoot from 'ember-simple-auth/utils/is-fastboot';
 
 /**
   The mixin for the application route, __defining methods that are called when
@@ -56,11 +57,7 @@ export default Mixin.create({
   */
   session: inject('session'),
 
-  _isFastBoot: computed(function() {
-    const fastboot = getOwner(this).lookup('service:fastboot');
-
-    return fastboot ? fastboot.get('isFastBoot') : false;
-  }),
+  _isFastBoot: isFastBoot(),
 
   /**
     The route to transition to after successful authentication.
--- a/addon/mixins/authenticated-route-mixin.js
+++ b/addon/mixins/authenticated-route-mixin.js
@@ -4,6 +4,7 @@ import { assert } from '@ember/debug';
 import { computed } from '@ember/object';
 import { getOwner } from '@ember/application';
 import Configuration from './../configuration';
+import isFastBoot from 'ember-simple-auth/utils/is-fastboot';
 
 /**
   __This mixin is used to make routes accessible only if the session is
@@ -35,11 +36,7 @@ export default Mixin.create({
   */
   session: service('session'),
 
-  _isFastBoot: computed(function() {
-    const fastboot = getOwner(this).lookup('service:fastboot');
-
-    return fastboot ? fastboot.get('isFastBoot') : false;
-  }),
+  _isFastBoot: isFastBoot(),
 
   /**
     The route to transition to for authentication. The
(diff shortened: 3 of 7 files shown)


addon/utils/is-fastboot.js
```diff
@@ -0,0 +1,10 @@
+import { computed } from '@ember/object';
+import { getOwner } from '@ember/application';
+
+export default function isFastBoot() {
+  return computed(function() {
+    const fastboot = getOwner(this).lookup('service:fastboot');
```
mike-north: This CPM must be used on a container-aware object (usually something that the container instantiates, like Service/Route/Controller/Component/Model/Adapter/Serializer, but **NOT** an `Ember.Object.extend({})` that you explicitly create)

We should [assert](https://www.emberjs.com/api/ember/3.1/functions/@ember%2Fdebug/assert) that `getOwner(this)` returns something truthy reasonable
```js
const container = getOwner(this);
assert('You may only use isFastBoot() on a container-aware object',
   container && .lookup === 'function');
```

addon/utils/is-fastboot.js
```diff
@@ -0,0 +1,10 @@
...
+
+    return fastboot ? fastboot.get('isFastBoot') : false;
```
marcoow: I think this should fail if the lookup for `service:fastboot` lookup fails and not silently fall back to returning `false`.

geekygrappler:
Added assertions for both of those things. @marcoow is an assertion correct or should I do more i.e. throw an `Error`?

addon/utils/is-fastboot.js
geekygrappler: 👍

addon/utils/is-fastboot.js
geekygrappler: 👍

geekygrappler:
and the assertions cause the tests to fail so I'll have to look into that later.

addon/utils/is-fastboot.js
```diff
@@ -0,0 +1,14 @@
+import { computed } from '@ember/object';
+import { getOwner } from '@ember/application';
+import { assert } from '@ember/debug';
+
+export default function isFastBoot() {
+  return computed(function() {
+    const container = getOwner(this);
+    assert('You may only use isFastBoot() on a container-aware object', container && container.lookup === 'function');
+    const fastboot = container.lookup('service:fastboot');
+    assert('You may only use isFastBoot() when with a fastboot service present');
```
mike-north: I think @marcoow was talking about returning false rather than throwing an assertion error

mike-north:
In tackling some deeper refactoring, I'm finding that there's significant value in having an isFastboot function in addition to the CP. Maybe we can

```js

export function isFastBoot(container) { container.lookup('service:fastboot') .... }

export function isFastbootProp() {
    return new Ember.computed( .... )
}

```

addon/utils/is-fastboot.js
mike-north: Agree, although the cases where that may happen are quite limited now that he’s pulling the service directly from the container. 

Still — could be a useful error if you’re working in an engine whose container doesn’t have the service.

geekygrappler:
I updated this to throw an Error if fastboot service is not present. @marcoow please can you check it over, because this is a behaviour change I do not understand. I am not really sure what the `fastboot` service is. Does this mean people running `ESA` in an app must also have a `fastboot` service? That doesn't make sense to me.

geekygrappler:
@mike-north I'll add a function for `isFastBook()` in a separate PR after this gets merged. Or you can do it if you want. I'm on a ⛵️ for the next 12 days and I don't know if this means I'll have more or less time to contribute....

marcoow:
@geekygrappler: I only now realize that we had the behavior where it falls back to returning `true` previously as well and you just copied that over. ~I still think it makes sense to keep the assertion though as we're likely to only hide other errors of we leave it out.~ (see below 😞)

The `fastboot` service is a service that `ember-cli-fastboot` adds to the app and that can be used to check and access all kinds of things, e.g. the `isFastBoot` property which is `true` when running in Node and `false` in the browser or for reading headers etc. That said, I realize that we cannot even add the above assertion as that would essentially make `ember-cli-fastboot` a dependency of ESA which of course we don't want. Sorry for the confusion 🙏

geekygrappler:
I changed this back to just return false if the `fastboot` service doesn't exist.

addon/utils/is-fastboot.js
```diff
@@ -0,0 +1,13 @@
...
+import { getOwner } from '@ember/application';
+import { assert } from '@ember/debug';
+
+export default function isFastBoot() {
+  return computed(function() {
+    const container = getOwner(this);
+    assert('You may only use isFastBoot() on a container-aware object', container && typeof container.lookup === 'function');
+
+    const fastboot = container.lookup('service:fastboot');
+    return fastboot ? fastboot.get('isFastBoot') : false;
```
mike-north: This can also be written as 
```js
return fastboot && fastboot.get('isFastBoot');
```
I'm not asking that you change it, just a FYI

Issue #1626, opened 2018-07-10
[Cleanup] Remove stubs at the end of each test scenario


mike-north:
Currently sinon is used without un-stubbing and un-mocking. This runs the risk of introducing non-determinism into the test suite (i.e., "Test 32 passes, but only when Test 14 runs before it").

Sinon's [sandbox feature](http://sinonjs.org/releases/v6.1.3/sandbox) is an easy way to ensure that things are restored at the end of each module -- ensuring that all tests start out with a clean environment.

The easy mode way to do this is

##### Before
```js
import { describe, it } from 'mocha';
import sinon from 'sinon';

describe('My test module', () => {
   it('works properly', () => {
        sinon.stub(....);
   });
});

```

##### After
```js
import { describe, beforeEach, afterEach, it } from 'mocha';
import sinonjs from 'sinon'; // 1. Rename import to sinonjs

describe('My test module', () => {
   let sinon; // 2. Define a sandbox called sinon
   beforeEach(() => { sinon = sinonjs.sandbox.create(); }); // 3. Set it up before each test
   afterEach(() => { sinon.restore(); });  // 4. Tear it down at the end of each test
   it('works properly', () => {
        sinon.stub(....); // << this continues to work with no change required
   });
});

```

marcoow:
I *think* we currently restore each stubbed method individually but the sandbox approach is much better of course. Maybe that didn't exist back in the day…

Issue #1625, opened 2018-07-10
[Cleanup] Refactor interesting functions away from mixins


mike-north:
Currently there are places where functions live on mixins for little or no reason (there may have been very good reasons before, but no longer). 

While leaving the public API surface intact, we should refactor these out so that they **may** be consumed directly

##### Before:

```js
export default Mixin.extend({
  add(x) {
    return this.get('total') + x;
  }
});
```

##### After:

```js
export function add(obj, x) {
  return obj.get('total') + x;
}

export default Mixin.extend({
  add(x) {
    return add(this, x);
  }
});
```

marcoow:
mixins are deprecated now – see #2198

PR #1627, opened 2018-07-10
OAuth2Callback mixin de-entanglement

mike-north:
* Improve location stubbing -- there's no reason this has to be entangled with a mixin
* Add sinon sandboxing (#1626)
* Break pure functions out of OAuth2 callback mixin (#1625)


tests/unit/utils/location-test.js
```diff
@@ -0,0 +1,30 @@
...
+  it('works', function() {
+    expect(LocationUtil.default()).to.be.ok;
+    expect(LocationUtil.default().hash).to.be.a('string');
+  });
+  it('works when stubbed', function() {
+    expect(foo.hash).to.be.a('string');
+    sinon.stub(LocationUtil, 'default').returns({ hash: 66 });
+    expect(LocationUtil.default().hash).to.be.a('number');
+    expect(foo.hash).to.be.a('number');
+  });
```
marcoow: I'm not sure this test case is valuable as it tests the stub only which could be removed if this test case is removed?

mike-north: It was failing before I started adding sandboxes to the existing tests, and I generally like to leave tests which reveal bugs in place (to protect against regression).

Happy to remove this if you feel otherwise.

marcoow: I'd remove this test as I think it doesn't test anything that's not within this file?

mike-north:
This is ready for another look

PR #1631, opened 2018-07-16
add docs on expected responses for OAuth2 auth

### LINKED ISSUE #1630: Clarity on How to Structure a Server Response for Seession Authentication
Hey y'all,

I'm still wrestling a bit with how to exactly structure my API responses to better sync up with Ember Simple Auth.

I'm running the following right now:

```javascript
let { email, password } = this.getProperties('email', 'password');
this.get('session').authenticate('authenticator:oauth2', email, password).catch((reason) => { 
});
```

For an invalid response, my server is returning:

```
{
  status_code: "401",
  error: {
    "message": "Invalid email or password"
  },
  result: []
}
```

For a valid response:

```
{ 
  access_token: "64ce61a4dd8a5b786283a355206a65f2840da480d7c60f10f07a8aa51872a4f5"
  token_type: "bearer"
  expires_in: "432000"
  created_at: "1531683221"
  status_code: 200
  message: "Signed in successfully."
  result: {
    id: 1,
    email: "sample_email@site.com"
  }
}
```

So, when I get that invalid response from the server, I'm getting a js error of `TypeError: resason is undefined`. I'm wondering if this is happening because of an error of ESA looking for a message or response that its not getting from the server and thus not kicking the proper response back to the catch section. Is there any insight on what's happening and what the server response should be? 

After all of this, I'd be happy to contribute some documentation on the matter!

marcoow:
While the recommended way of implementing OAuth 2.0 auth is using a ready-to-go middleware, some people choose to write their own and we've seen some confusion around the exact spec of the expected responses in the past. This adds some info on that to the docs to hopefully prevent that confusion in the future.

closes #1630

### DIFF
--- a/addon/authenticators/oauth2-password-grant.js
+++ b/addon/authenticators/oauth2-password-grant.js
@@ -214,6 +214,36 @@ export default BaseAuthenticator.extend({
     method also schedules refresh requests for the access token before it
     expires.__
 
+    The server responses are expected to look as defined in the spec (see
+    http://tools.ietf.org/html/rfc6749#section-5). The response to a successful
+    authentication request should be:
+
+    ```json
+    HTTP/1.1 200 OK
+    Content-Type: application/json;charset=UTF-8
+
+    {
+      "access_token":"2YotnFZFEjr1zCsicMWpAA",
+      "token_type":"bearer",
+      "expires_in":3600, // optional
+      "refresh_token":"tGzv3JOkF0XG5Qx2TlKWIA" // optional
+    }
+    ```
+
+    The response for a failing authentication request should be:
+
+    ```json
+    HTTP/1.1 400 Bad Request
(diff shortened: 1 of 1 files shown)


hiimtaylorjones:
Thanks for this!

addon/authenticators/oauth2-password-grant.js
```diff
@@ -214,6 +214,36 @@ export default BaseAuthenticator.extend({
     method also schedules refresh requests for the access token before it
     expires.__
 
+    The server responses to are expected to look as defined in the spec (see
```
jessica-jordan: Woops, just missed that one: I think it would be

```
The server responses are expected to look [...]
```

marcoow: fixed 👍

PR #1636, opened 2018-07-18
Upgrade `broccoli-file-creator`

dcyriller:
As it raises deprecations in newer versions of ember-cli.

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,7 +30,7 @@
   },
   "dependencies": {
     "base-64": "^0.1.0",
-    "broccoli-file-creator": "^1.1.1 || ^2.0.0",
+    "broccoli-file-creator": "^2.0.0",
     "broccoli-funnel": "^1.2.0 || ^2.0.0",
     "broccoli-merge-trees": "^2.0.0 || ^3.0.0",
     "ember-cli-babel": "^6.8.2",
--- a/yarn.lock
+++ b/yarn.lock
@@ -1325,7 +1325,7 @@ broccoli-file-creator@^1.0.0, broccoli-file-creator@^1.1.1:
     rsvp "~3.0.6"
     symlink-or-copy "^1.0.1"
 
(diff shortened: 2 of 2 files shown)


dcyriller:
I'm curious: why specifying permissive package versions?

marcoow:
I **think** you should be able to upgrade to the latest version by running

```
yarn upgrade broccoli-file-creator@"^1.1.1 || ^2.0.0"
```

In general I think it's a bad practice in the JS world to always raise any dependency to the latest available version. ember-simple-auth works perfectly fine with `broccoli-file-creator@^1.1.1` as well as `broccoli-file-creator@^2.0.0` so there is no need to **require** `broccoli-file-creator@^2.0.0` and grow everyone's `node_modules` further (as they might have other dependencies that might require `broccoli-file-creator@^1.1.1` and now they have to have both versions of the same package).

marcoow:
To be clear: the above only applies to `dependencies` of **packages**. For a package's `devDependencies` or **any** dependencies in an **application**, it's totally fine to use the latest version of everything all the time (though you'll typically be doing that via a lock file). Just for `dependencies` of **packages** I don't think it's great practice as you're blowing up the `node_modules` folders of everyone using your package as explained above.

dcyriller:
Thanks @marcoow for your explanation. It's clear.

I'm closing the PR then.

marcoow:
For some context: as we discussed on Slack, `broccoli-file-creator@1.x` does not work with Broccoli 2 while `broccoli-file-creator@2.x` does. While of course, ESA **allows** `broccoli-file-creator@2.x`, ESA could already be used with Broccoli 2 but this permissive dependency makes upgrading a project to Broccoli 2 unnecessarily hard.

dcyriller:
@marcoow the tests were failing for a reason I don't fully understand.

I extracted in a separate PR the update of `ember-try` config. It can be found [here](https://github.com/simplabs/ember-simple-auth/pull/1642).

PR #1640, opened 2018-07-19
Bugfix: Don't try to bind arrow functions

pichfl:
Arrow functions have neiter a `this` that could be bound nor their own `arguments` object.

Replaces #1638


marcoow:
Ugh, so sorry - I just realized this was already addressed by #1604 which should have been merged already 😞 Thanks for the effort but I'll merge that one as it came first 🙏

pichfl:
Should have looked myself 😅

PR #1642, opened 2018-07-23
Update `ember-try` config

dcyriller:
Run `ember init` to regenerate `config/ember-try.js` file. The config gets modified this way:
* Add `ember-source-channel-url` dependency to dynamically pick ember
versions
* Add ember-2.18-lts scenario
* Use yarn to run the test in travis

This was required to make https://github.com/simplabs/ember-simple-auth/pull/1636 tests pass.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,7 +27,7 @@ install:
   - yarn install --no-lockfile
 
 script:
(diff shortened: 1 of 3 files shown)


Turbo87:
Thanks!

PR #1650, opened 2018-07-27
Improve tests

marcoow:
This adds Ember 3.0 to the test scenarios and runs the tests in parallel instead of one after the other which should hopefully help getting the build times down 🙏 

Another advantage of this is that if any of the scenarios fail, we don't have to re-run **all** scenarios and wait for ca. 20min but can re-run only that **one** scenario wich is much faster of course.

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,6 +21,7 @@
     "start": "ember serve",
     "test": "ember test",
     "test:all": "ember try:each",
+    "test:one": "ember try:one",
     "test:fastboot": "ember fastboot:test",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -21,6 +21,7 @@
     "start": "ember serve",
     "test": "ember test",
     "test:all": "ember try:each",
+    "test:one": "ember try:one",
```
jessica-jordan: This is cool, I didn't know that one before :+1:

PR #1628, opened 2018-07-10
[WIP] ApplicationRoute - mixin de-entanglement

mike-north:
This is a super-rough hack that I'll refine later on, but I wanted to leave it here for the purposes of discussion. 

By refactoring two functions out of the mixin, I'm now able to do the `ApplicationRouteMixin`'s job on a service or any other ember singleton.

## TODO:
- [ ] Figure out intuitive function names
- [ ] Break these two top-level functions out into a single `./utils/___`
- [ ] Docs
- [ ] Tests for non-mixin use

PR #1655, opened 2018-08-17
Authenticated route: allow easier use w/o mixin

mike-north:
Part of effort described in #1619 
Supercedes #1628

### DIFF
--- a/addon/mixins/authenticated-route-mixin.js
+++ b/addon/mixins/authenticated-route-mixin.js
@@ -4,7 +4,33 @@ import { assert } from '@ember/debug';
 import { computed } from '@ember/object';
 import { getOwner } from '@ember/application';
 import Configuration from './../configuration';
-import isFastBoot from 'ember-simple-auth/utils/is-fastboot';
+import isFastBootCPM, { isFastBoot } from '../utils/is-fastboot';
+
+/**
+ * If the user is unauthenticated, invoke `callback`
+ *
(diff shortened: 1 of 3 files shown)


addon/mixins/authenticated-route-mixin.js
```diff
@@ -4,7 +4,33 @@ import { assert } from '@ember/debug';
...
+import isFastBootCPM, { isFastBoot } from '../utils/is-fastboot';
+
+/**
+ * If the user is unauthenticated, invoke `callback`
+ *
+ * @param {ApplicationInstance} owner The ApplicationInstance that owns the service (and possibly fastboot and cookie) service(s)
+ * @param {Transition} transition Transition for the user's original navigation
+ * @param {(...args: []any) => any} callback Callback that will be invoked if the user is unauthenticated
+ */
+function runIfUnauthenticated(owner, transition, callback) {
```
mike-north: Once we have more of the simple-auth without mixins (#1619) story in place, we can start to carefully name these functions and export them from their respective modules. For now, we may want to leave these as module-private, and make a single "designing the public API" pass at the end

marcoow: yep, I guess that makes sense. That will also probably be the hardest part about this - making sure ESA is still - despite to the relatively large number of things it does now - relatively easy to consume/integrate.

PR #1656, opened 2018-08-17
Fix linting situation

mike-north:
Fixes the following issues
- [x] there were numerous lint errors on `master`
- [x] node-js stuff (`ember-cli-build.js`, etc...) was not linted at all
- [x] travis wasn't running the linter in CI


config/ember-try.js
```diff
@@ -7,7 +7,7 @@ module.exports = function() {
     getChannelURL('release'),
     getChannelURL('beta'),
     getChannelURL('canary')
-  ]).then((urls) => {
+  ]).then((/* urls */) => {
```
mike-north: This was an unused variable. source channel URLs are being retrieved, but are not used. I'll take care of this in a subsequent PR

server/app.js
```diff
@@ -1,7 +1,11 @@
-const express = require('express');
-const app = express();
-const server = require('./index.js')(app);
+'use strict';
 
-app.listen(3000, () => {
+var express = require('express');
+var app = express();
+// eslint-disable-next-line no-unused-vars
+var server = require('./index.js')(app);
```
mike-north: Took this

https://github.com/simplabs/ember-simple-auth/blob/84ca953ec7ac2f3db41ebb378b3c4d68208a64bd/server/.eslintrc.js#L3

as a cue that node 0.x support is important here for some reason. I'd also be happy to update the `.eslintrc.js` in this folder to permit node features that weren't available in `0.x` if that's the desirable way to address this conflict

tests/dummy/config/targets.js
```diff
@@ -4,7 +4,7 @@ const browsers = [
   'last 1 Safari versions'
 ];
 
-const isCI = !!process.env.CI;
+const isCI = Boolean(process.env.CI);
```
mike-north: This came along with the default ruleset. I'd also be happy to set up an exception for `!!`, since it's one of the more easily readable type coersions.

package.json
```diff
@@ -74,10 +74,12 @@
     "ember-resolver": "^5.0.1",
     "ember-sinon": "~2.2.0",
     "ember-source": "~3.3.0",
+    "ember-source-channel-url": "^1.1.0",
```
mike-north: Was being used here

https://github.com/simplabs/ember-simple-auth/blob/84ca953ec7ac2f3db41ebb378b3c4d68208a64bd/config/ember-try.js#L3

but not actually included with the project. It may have been a dependency of something at one time, but a fresh install from the lockfile in master reveals that it doesn't end up being in the project

```
yarn why ember-source-channel-url
yarn why v1.9.4
[1/4] 🤔  Why do we have the module "ember-source-channel-url"...?
[2/4] 🚚  Initialising dependency graph...
[3/4] 🔍  Finding dependency...
error We couldn't find a match!
✨  Done in 1.30s.
```

marcoow: 👍

server/app.js
marcoow: We're still supporting Ember 1.12 🙀 - I'm not sure actually that implies we **have** to support Node 0 as well…

tests/dummy/config/targets.js
marcoow: I think `Boolean(…)` is fine actually - might be a bit more readable for beginners 👍

PR #1657, opened 2018-08-17
Get ember-try builds from ember-source-channel-url

mike-north:


config/ember-try.js
```diff
@@ -7,7 +7,7 @@ module.exports = function() {
     getChannelURL('release'),
     getChannelURL('beta'),
     getChannelURL('canary')
-  ]).then((/* urls */) => {
+  ]).then(([releaseUrl, betaUrl, canaryUrl]) => {
```
marcoow: This [does not pass the linter](https://travis-ci.org/simplabs/ember-simple-auth/jobs/419803028#L546):

```
Destructuring is not supported until Node.js 6.0.0. The configured version range is '^4.5 || 6.* || >= 7.*'  node/no-unsupported-features/es-syntax
```

😞

marcoow: This [does not pass the linter](https://travis-ci.org/simplabs/ember-simple-auth/jobs/419803028#L546):

```
Destructuring is not supported until Node.js 6.0.0. The configured version range is '^4.5 || 6.* || >= 7.*'  node/no-unsupported-features/es-syntax
```

😞

mike-north: haha good thing we added that 🙌. I'll fix

PR #1662, opened 2018-08-24
add documentation how to use it with ember-fetch

luxferresum:
I'm not sure why there is the `authorize()` method at all. (backward compatibility?) Actually I would always prefer the `headers` property. The `DataAdapterMixin` is still useful for the invalidation feature.

I think this will get more important when people use the possibility to get rid of jQuery.


marcoow:
This is messed up partly by me and partly by the github outage I think - sorry @luxferresum, I'll create a replacement 😞

PR #1126, opened 2016-12-09
Remove unncessary transition.abort calls on auth mixins

juanazam:
`transition.abort` were removed on https://github.com/simplabs/ember-simple-auth/pull/992, but were re added with fast boot merge.

PR #1667, opened 2018-08-30
UnauthenticatedRouteMixin: Remove reference to unused transition

geekygrappler:
- clean up on isle 4! Left over clean up from #1126


addon/mixins/unauthenticated-route-mixin.js
```diff
@@ -62,7 +62,6 @@ export default Mixin.create({
    `beforeModel` method is actually executed.
 
     @method beforeModel
-    @param {Transition} transition The transition that lead to this route
```
marcoow: I don't think we should be removing this - this method actually accepts a `transition` argument although that might no longer be used.

geekygrappler:
Updated

PR #1670, opened 2018-08-31
refactor(authentication-route-mixin): use router service

geekygrappler:
- use router service instead of the routes `transitionTo` method

@mike-north in line with one of your todos from #1619 (https://github.com/simplabs/ember-simple-auth/issues/1619#issuecomment-403178107) is this correct? If so I can do this wherever else `router.transitionTo` is used.

### DIFF
--- a/addon/mixins/authenticated-route-mixin.js
+++ b/addon/mixins/authenticated-route-mixin.js
@@ -62,6 +62,11 @@ export default Mixin.create({
   */
   session: service('session'),
 
+  _router: computed(function() {
+    let owner = getOwner(this);
+    return owner.lookup('service:router') || owner.lookup('router:main');
+  }),
+
   _isFastBoot: isFastBootCPM(),
 
   /**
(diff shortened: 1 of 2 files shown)


addon/mixins/authenticated-route-mixin.js
```diff
@@ -61,6 +61,7 @@ export default Mixin.create({
     @public
   */
   session: service('session'),
+  router: service('router'),
```
mike-north: I agree w/ @marcoow's concern

> The problem with this is that it only works if the router service exists. 

For versions of Ember that predate the router service, you can get ahold of the router singleton directly via `getOwner(this).lookup('router:main')`

geekygrappler: Updated to use router service or get the router singleton.

addon/mixins/authenticated-route-mixin.js
```diff
@@ -61,6 +61,10 @@ export default Mixin.create({
     @public
   */
   session: service('session'),
+  router: computed(function() {
+    return this.get('routerService') ? this.get('routerService') : getOwner(this).lookup('router:main');
+  }),
+  routerService: service('router'),
```
marcoow: This will raise if the router service does not exist which is the case for older versions of Ember. I think what you'll want to do instead is:

```js
router: computed(function() {
  let owner = getOwner(this);
  return owner.lookup('service:router') || owner.lookup('router:main');
}),
```

which will use the router service if it exists and fall back to the main router if it doesn't.

geekygrappler: Hi @marcoow or @mike-north can one of you check that this can be merged now  that we have a robust  call for the router?

PR #1679, opened 2018-09-21
Fixes Handling of ClientID for Password Grant

### LINKED ISSUE #1678: ClientID implementation for OAuth2 Password Grant is Wrong
The `clientId` should be passed as a query string, not as the username of an `Authorization` header.

- https://alexbilbie.com/guide-to-oauth-2-grants/
- https://www.oauth.com/oauth2-servers/access-tokens/password-grant/
- https://auth0.com/docs/api-auth/tutorials/password-grant

ekristen:
* Fixes how `client_id` is passed on Password Grant for OAuth2.
* Adds support for `client_secret` and passing it as well for Password Grant if set.

References on Password Grants:
- https://www.oauth.com/oauth2-servers/access-tokens/password-grant/
- https://auth0.com/docs/api-auth/tutorials/password-grant
- https://alexbilbie.com/guide-to-oauth-2-grants/

Fixes #1678

### DIFF
--- a/addon/authenticators/oauth2-password-grant.js
+++ b/addon/authenticators/oauth2-password-grant.js
@@ -126,6 +126,14 @@ export default BaseAuthenticator.extend({
 
   _refreshTokenTimeout: null,
 
+  _clientIdHeader: computed('clientId', function() {
+    const clientId = this.get('clientId');
+    if (!isEmpty(clientId)) {
+      const base64ClientId = window.base64.encode(clientId.concat(':'));
+      return { Authorization: `Basic ${base64ClientId}` };
+    }
+  }),
+
   /**
     When authentication fails, the rejection callback is provided with the whole
     Fetch API [Response](https://fetch.spec.whatwg.org/#response-class) object
@@ -338,10 +346,14 @@ export default BaseAuthenticator.extend({
   makeRequest(url, data, headers = {}) {
     headers['Content-Type'] = 'application/x-www-form-urlencoded';
 
-    if (this.get('clientId') !== null) {
-      data['client_id'] = this.get('clientId');
+    if (this.get('sendClientIdAsQueryParam')) {
+      const clientId = this.get('clientId');
+      if (!isEmpty(clientId)) {
+        data['client_id'] = this.get('clientId');
+      }
     }
-    if (this.get('refreshAccessTokens')) {
+
+    if (data['grant_type'] === 'password' && this.get('refreshAccessTokens')) {
       data['offline_token'] = true;
     }
 
@@ -354,6 +366,13 @@ export default BaseAuthenticator.extend({
       headers,
       method: 'POST'
     };
+
+    if (!this.get('sendClientIdAsQueryParam')) {
+      const clientIdHeader = this.get('_clientIdHeader');
+      if (!isEmpty(clientIdHeader)) {
+        merge(options.headers, clientIdHeader);
+      }
+    }
 
     return new RSVP.Promise((resolve, reject) => {
       fetch(url, options).then((response) => {
--- a/addon/authenticators/oauth2-password-grant.js
+++ b/addon/authenticators/oauth2-password-grant.js
@@ -9,6 +9,7 @@ import {
   merge,
   assign as emberAssign
 } from '@ember/polyfills';
+import { deprecate } from '@ember/application/deprecations';
 import Ember from 'ember';
 import BaseAuthenticator from './base';
 import fetch from 'fetch';
@@ -34,14 +35,15 @@ export default BaseAuthenticator.extend({
   init() {
     this._super(...arguments);
     deprecate(`Ember Simple Auth: Client ID as Authorization Header is deprecated in favour of Client ID as Query String Parameter.`,
-          false,
-          {
-            id: 'ember-simple-auth.OAuth2PasswordGrantAuthenticator',
-            until: '2.0.0',
-            url: 'https://github.com/simplabs/ember-simple-auth#deprecation-of-authorizers',
-          }
+      false,
+      {
+        id: 'ember-simple-auth.OAuth2PasswordGrantAuthenticator',
+        until: '2.0.0',
+        url: 'https://github.com/simplabs/ember-simple-auth#deprecation-of-authorizers',
+      }
     );
-  }
+  },
+
   /**
     Triggered when the authenticator refreshed the access token (see
     [RFC 6749, section 6](http://tools.ietf.org/html/rfc6749#section-6)).
--- a/addon/authenticators/oauth2-password-grant.js
+++ b/addon/authenticators/oauth2-password-grant.js
@@ -115,15 +115,6 @@ export default BaseAuthenticator.extend({
 
   _refreshTokenTimeout: null,
 
-  _clientIdHeader: computed('clientId', function() {
-    const clientId = this.get('clientId');
-
-    if (!isEmpty(clientId)) {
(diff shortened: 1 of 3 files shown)


ekristen:
Fixes https://github.com/simplabs/ember-simple-auth/issues/1678

addon/authenticators/oauth2-password-grant.js
```diff
@@ -53,6 +53,18 @@ export default BaseAuthenticator.extend({
...
+    The client_secret to be sent to the authentication server. This should be set
+    along with the client_id if the client needs to authenticate itself against
+    the authorization server in addition to the user
+
+    @property clientSecret
+    @type String
+    @default null
+    @public
+  */
+  clientSecret: null,
```
marcoow: This was intentionally left out - as soon as you publish your client secret as part of your publicly accessible Ember app it's no longer a secret.

ekristen: You are right, I got a little over zealous, I'll remove it.

ekristen:
@marcoow if you can provide guidance on how to feature flag something (I'm pretty new to the ember world) I'll gladly fix this.

marcoow:
Sure, don't worry - happy to help. So what we'll have to do is:

* add a property `sendClientIdAsQueryParam` to `addon/authenticators/oauth2-password-grant.js` and default that to `false`
* enable the new behavior that you add in this MR only when `sendClientIdAsQueryParam` is `true`
* show a deprecation when `sendClientIdAsQueryParam` is `false` (from `init` in `addon/authenticators/oauth2-password-grant.js`)
* when we release the next major version, we can remove the flag, the deprecation and the old behavior

The idea behind this is that people who update to the next major version get the deprecation and can opt-in to the new behavior and validate that it works with their OAuth 2.0 server. Once the next major version is released, they just remove the flag. At the same time, we don't have to release this as a breaking change (yet) though.

ekristen:
@marcoow I'm not sure how to add deprecation notices but the rest should be in properly now.

marcoow:
Here's an example for a deprecation: https://github.com/simplabs/ember-simple-auth/blob/76e0eacf94f1d1d4d1194b3ef93ceae0bc5e000e/addon/authorizers/base.js#L28

addon/authenticators/oauth2-password-grant.js
```diff
@@ -336,6 +346,17 @@ export default BaseAuthenticator.extend({
...
+    if (this.get('sendClientIdAsQueryParam')) {
+      const clientId = this.get('clientId');
+      if (!isEmpty(clientId)) {
+        data['client_id'] = this.get('clientId');
+      }
+    }
+
+    if (data['grant_type'] === 'password' && this.get('refreshAccessTokens')) {
+      data['offline_token'] = true;
+    }
```
marcoow: This seems unrelated?

ekristen: I can open a second pull request if you like. It's actually required if you want access tokens to be refreshed by an Oauth2 server that's been implemented to spec. Generally you have to pass the offline_token as well as have the server allow for tokens to be refreshed.

ekristen:
I'll look at how to add tests for this.

addon/authenticators/oauth2-password-grant.js
marcoow: Hm, I can't find any reference to _"offline_token"_ in https://tools.ietf.org/html/rfc6749… Anyway, I'd put this in a separate PR for simplicity.

ekristen: https://tools.ietf.org/html/rfc6749#page-47

ekristen: I'll have to dig a bit more, but it looks it might not be a spec thing, but instead an opinionated option that many oauth2 services implement. The most common being offline_token in the body, but Auth0 implements it as a scope. 

I think it's important to look at adding because many require it.

marcoow:
Hm, I didn't read through all of that tbh but Cmd+F "offl" comes back with 0 matches…

marcoow:
@ekristen: any update on the tests? Would be great to get this in.

ekristen:
@marcoow forgot to mention you previously ^

ekristen:
@marcoow following up

addon/authenticators/oauth2-password-grant.js
```diff
@@ -31,6 +32,18 @@ const keys = Object.keys || emberKeys; // Ember.keys deprecated in 1.13
   @public
 */
 export default BaseAuthenticator.extend({
+  init() {
+    this._super(...arguments);
+    deprecate(`Ember Simple Auth: Client ID as Authorization Header is deprecated in favour of Client ID as Query String Parameter.`,
```
marcoow: This should not actually be in the `init` as using the authenticator as such is totally fine - just leaving `sendClientIdAsQueryParam` as `false` is deprecated.

addon/authenticators/oauth2-password-grant.js
```diff
@@ -31,6 +32,18 @@ const keys = Object.keys || emberKeys; // Ember.keys deprecated in 1.13
...
+      false,
+      {
+        id: 'ember-simple-auth.OAuth2PasswordGrantAuthenticator',
```
marcoow: The deprecation ID should be specific to the deprecation - something like `ember-simple-auth.oauth2-password-grant-authenticator.client-id-as-authorization`.

addon/authenticators/oauth2-password-grant.js
```diff
@@ -31,6 +32,18 @@ const keys = Object.keys || emberKeys; // Ember.keys deprecated in 1.13
...
+        until: '2.0.0',
+        url: 'https://github.com/simplabs/ember-simple-auth#deprecation-of-authorizers',
```
marcoow: This links to the wrong deprecation.

tests/unit/authenticators/oauth2-password-grant-test.js
```diff
@@ -117,7 +117,7 @@ describe('OAuth2PasswordGrantAuthenticator', () => {
       authenticator.authenticate('username', 'password');
     });
 
-    it('sends an AJAX request to the token endpoint with client_id Basic Auth header', function(done) {
+    it('(deprecated) sends an AJAX request to the token endpoint with client_id Basic Auth header', function(done) {
```
marcoow: There should also be a test that asserts that the deprecation is triggered at the right point.

addon/authenticators/oauth2-password-grant.js
ekristen: Done. Moved the `authenticate`, triggers only if set to false.

addon/authenticators/oauth2-password-grant.js
ekristen: Done.

addon/authenticators/oauth2-password-grant.js
ekristen: Fixed. Added additional docs to README.

tests/unit/authenticators/oauth2-password-grant-test.js
ekristen: Tests for deprecation notice now.

ekristen:
Hopefully it's ok now? Tests pass for me locally.

tests/unit/authenticators/oauth2-password-grant-test.js
marcoow: Sorry - I meant adding a separate test case to test this. We're generally testing one _"thing"_ per test case.

PR #1697, opened 2018-10-22
[DOC] add documentation how to use it with ember-fetch

luxferresum:
I'm not sure why there is the authorize() method at all. (backward compatibility?) Actually I would always prefer the headers property. The DataAdapterMixin is still useful for the invalidation feature.

I think this will get more important when people use the possibility to get rid of jQuery.

Looks like I've overwritten the changes for #1662. Sorry for that! Here again. I hope the change itself it welcome, otherwise sorry for opening this up again.

### DIFF
--- a/README.md
+++ b/README.md
@@ -582,6 +582,21 @@ export default JSONAPIAdapter.extend(DataAdapterMixin, {
 });
 ```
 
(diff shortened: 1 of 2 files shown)


marcoow:
duplicate of #1698

PR #1698, opened 2018-10-22
[DOC] add documentation how to use it with ember-fetch

marcoow:
This replaces #1662 and adds a not on usage with `ember-fetch`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -584,6 +584,22 @@ export default JSONAPIAdapter.extend(DataAdapterMixin, {
 });
 ```
 
+When using with `ember-fetch` the `authorize` will not be called and the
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -584,6 +584,22 @@ export default JSONAPIAdapter.extend(DataAdapterMixin, {
 });
 ```
 
+When using with `ember-fetch` the `authorize` will not be called and the
```
jessica-jordan: What do you think about:

```suggestion
When used with `ember-fetch` the `authorize` method will not be called and the
```

?

marcoow:
@jessica-jordan: I updated the same thing in the API docs as well 👍

PR #1712, opened 2018-11-19
Allows ember-fetch 6.x

lougreenwood:
Fixes #1705

PR #1716, opened 2018-11-22
fix: Enable 6.0.0 of ember-fetch

jasonmit:


jasonmit:
https://github.com/simplabs/ember-simple-auth/pull/1712

PR #1722, opened 2018-12-09
Fix deprecation warning when running with Ember 3.6

richard-viney:
In Ember 3.6 the `off()` function must be given the specific handler function to remove.

### DIFF
--- a/addon/internal-session.js
+++ b/addon/internal-session.js
@@ -49,7 +49,7 @@ export default ObjectProxy.extend(Evented, {
 
     let authenticator = this._lookupAuthenticator(this.authenticator);
     return authenticator.invalidate(this.content.authenticated, ...arguments).then(() => {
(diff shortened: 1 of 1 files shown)


mydea:
Any plans on when a new release will be made? :)

richard-viney:
May have missed a depreciation here, am seeing this now: `DEPRECATION: You attempted to remove a function listener which did not exist on the instance, which means it was an inherited prototype listener, or you attempted to remove it before it was added. Prototype function listeners have been deprecated, and attempting to remove a non-existent function listener this will error in the future. [deprecation id: events.inherited-function-listeners] See https://emberjs.com/deprecations/v3.x#toc_events-inherited-function-listeners for more details.`

Will do another small fix for this.

richard-viney:
#1725

PR #1725, opened 2018-12-20
Remove unnecessary calls to off() that trigger deprecation warnings

richard-viney:
Followup to #1722.

### DIFF
--- a/addon/internal-session.js
+++ b/addon/internal-session.js
@@ -167,8 +167,6 @@ export default ObjectProxy.extend(Evented, {
 
   _bindToAuthenticatorEvents() {
     const authenticator = this._lookupAuthenticator(this.authenticator);
-    authenticator.off('sessionDataUpdated', this, this._onSessionDataUpdated);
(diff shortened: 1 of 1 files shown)


Turbo87:
why do you think that these are unnecessary? wouldn't it lead to memory-leaks if we removed them?

richard-viney:
Believe it's ok because the original code was immediately putting the event handlers back on after just removing them.

I tested that multiple calls to `on()` with identical arguments doesn't result in the specified function being called multiple times, i.e. it doesn't allow the same `[context, function]` pair to be registered more than once. That means that the `off()` calls removed in this PR weren't ever doing much anyway.

Thoughts?

josemarluedke:
@marcoow @Turbo87 @jessica-jordan Can we get a release containing this change? Thanks.

robclancy:
Someone press the release button...

jessica-jordan:
@josemarluedke @robclancy I'm on it!

quaertym:
@jessica-jordan Please push to npm.

PR #1745, opened 2019-03-07
Re-Enable mandatory fastboot tests

### LINKED ISSUE #1719: Re-enable *mandatory* fastboot tests in CI
The fastboot CI job has been failing on master for about a month, probably due to a combination of

- ember-data#master being incompatible with recent releases of ember (you'll get a rollup build error when attempting to build)
- ember-fetch's custom rollup stuff not working nicely with fastboot testing infrastructure

To address this, I've improved our CI configuration (#1718) to detect this kind of breakage in nightly builds, by ensuring that tests are run both _with and without_ taking the lockfile into account.

Unfortunately, getting fastboot testing passing again will take some non-trivial digging. Here's what I've found in my limited time
- in https://github.com/kaliber5/ember-fastboot-addon-tests most dependency versions are "borrowed" from the library being tested. ember-data is an exception, which defaults to `ember-data#master`. I've opened a PR here which at least gets the first test running again https://github.com/kaliber5/ember-fastboot-addon-tests/pull/26
- ember-fetch's custom rollup stuff seems to be failing when it fails to find (I think) babel-preset-env in the fastboot test app. I haven't had a chance to dig into this any further yet

In the mean time, to allow meaningful development to continue while some of this infrastructure is updated/repaired, I've marked the CI job as an allowed failure.

marcoow:
They were previously disabled due to repeated failures (caused by infrastructure, not actually broken code).

closes #1719

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,6 @@ jobs:
   fail_fast: true
   allow_failures:
     - env: SCRIPT="test:one ember-canary"
-    - env: SCRIPT=test:fastboot
 
   include:
     # runs linting and tests with current locked deps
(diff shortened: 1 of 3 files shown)


marcoow:
doesn't work for now :disappointed:

PR #1764, opened 2019-03-22
Engine Support

marcoow:
This adds an engine to the dummy app so we can validate/test we support engines correctly.

### TODO

- [x] add tests
- ~ideally we can automatically `transitionToExternal` instead of `transitionTo` in the `AuthenticatedRouteMixin` if we detect we're running in an engine (and the login route is **not** in the enging)~ – this actually makes no sense as the authentication route could be **within** the engine
- ~disable engines tests in ember versions < 2.12~ – this is no longer necessary as we dropped support for Ember versions < 3.0
- [x] add docs


tests/dummy/lib/my-engine/addon/templates/index.hbs
```diff
@@ -0,0 +1,7 @@
+<div class="page-header">
+  <h1>Engine Page</h1>
+</div>
+<div class="alert alert-warning">
+  This is a protected page in an engine only visible to authenticated users!
+</div>
+<a {{action 'logout'}} class="btn btn-danger" href="#" role="button" data-test-logout-button>Logout</a>
```
sdebarros, line 7: Why is this not a button? 😄

README.md
```diff
@@ -555,6 +556,57 @@ in `package.json`:
...
+import Route from '@ember/routing/route';
+import AuthenticatedRouteMixin from 'ember-simple-auth/mixins/authenticated-route-mixin';
+
+export default Route.extend(AuthenticatedRouteMixin, {
+  triggerAuthentication() {
+    this.transitionToExternal('login');
+  }
+});
+
+```
```
sdebarros, line 608: I think it would be nice if the examples also used Octane syntax, as the rest of the README examples.

marcoow: You're right – fixed

tests/dummy/lib/my-engine/addon/templates/index.hbs
marcoow: I have no idea ;) fixed

Issue #1714, opened 2018-11-20
makeRequest doesn't return XHR object


gtb104:
We upgraded from 1.1.0 to 1.7.0, and now we can't establish a WebSocket connection. After the FastBoot compatibility PR #1035, `makeRequest()` no longer returns the jQuery XHR object, but a Promise instead. We were using that XHR object to pull out some header information (CSRF token specifically) and saving it off for other requests to use. Is there a new way to get the header info from `makeRequest()`? If not, the `request.headers` property should also be returned from the `fetch()` call.

gtb104:
It would be helpful if https://github.com/simplabs/ember-simple-auth/blob/master/addon/authenticators/oauth2-password-grant.js#L362-L364 returned the `response` object with a `responseJSON` property like it does in the `if` and the `catch`.
```
            } else {
              response.responseJSON = json;
              resolve(response);
            }
```

marcoow:
We can't change the value that `makeRequest` resolves with now as that would be a breaking change actually. I do agree we should do that though as currently the API is a bit inconsistent. I think the way to go here would be to add a `resolveWithResponse` property similar to [`rejectWithResponse`](https://github.com/simplabs/ember-simple-auth/blob/master/addon/authenticators/oauth2-password-grant.js#L159) that would allow us to introduce this in a non-breaking way.

marcoow:
cleaning up old issues

gtb104:
@marcoow, Any timeline on if/when your alternate suggestion would be implemented?

marcoow:
No - if you'd like to contribute a PR implementing just that, that would be ❤️

PR #1798, opened 2019-04-02
Add resolveWithResponse to oauth2-password-grant

gtb104:
This is an implementation suggestion to cover issue #1714.

This adds the `resolveWithResponse` property to `oauth2-password-grant`. When enabled, this will allow for the return of the full response object upon successful authentication. This allows the consumer to pull details from the response header if desired.

### DIFF
--- a/addon/authenticators/oauth2-password-grant.js
+++ b/addon/authenticators/oauth2-password-grant.js
@@ -170,6 +170,21 @@ export default BaseAuthenticator.extend({
   rejectWithResponse: false,
 
   /**
+    When authentication succeeds, the callback is provided with the whole
+    Fetch API [Response](https://fetch.spec.whatwg.org/#response-class) object
+    instead of the responseJSON only.
+
+    This is useful for cases when the backend provides additional context not
+    available in the responseJSON.
+
+    @property resolveWithResponse
+    @type Boolean
+    @default false
+    @public
+  */
+  resolveWithResponse: false,
+
+  /**
     Restores the session from a session data object; __will return a resolving
     promise when there is a non-empty `access_token` in the session data__ and
     a rejecting promise otherwise.
@@ -279,13 +294,14 @@ export default BaseAuthenticator.extend({
       const data = { 'grant_type': 'password', username: identification, password };
       const serverTokenEndpoint = this.get('serverTokenEndpoint');
       const useResponse = this.get('rejectWithResponse');
+      const resolveWithResponse = this.get('resolveWithResponse');
       const scopesString = makeArray(scope).join(' ');
       if (!isEmpty(scopesString)) {
         data.scope = scopesString;
       }
       this.makeRequest(serverTokenEndpoint, data, headers).then((response) => {
         run(() => {
-          if (!this._validate(response)) {
+          if (!this._validate(response, resolveWithResponse)) {
             reject('access_token is missing in server response');
           }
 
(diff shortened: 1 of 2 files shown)


addon/authenticators/oauth2-password-grant.js
```diff
@@ -391,7 +409,12 @@ export default BaseAuthenticator.extend({
               response.responseJSON = json;
               reject(response);
             } else {
-              resolve(json);
+              if (resolveWithResponse) {
+                response.responseJSON = json;
```
marcoow: I don't think we should be doing this here. The `response` has a `json` method that should be used instead anyway.

gtb104: So it would be `resolve(response)` and the consumer of the response could use `response.json()` to get JSON instead of using `responseJSON`?

marcoow: Yes, exactly - we shouldn't modify the original response object, that was a bad idea when I first introduced it above.

gtb104: Hmmm. It doesn't appear that you can `resolve(response)` from within the promise returned from `response.text()`, as calling `response.json()` throws errors like `Uncaught TypeError: Already read`. This is caused from calling more than one Body method on a response.

I tried
```
  if (resolveWithResponse) {
    resolve(response);
  } else {
      ...
    }
  }
}
```
but that would cause issues with everywhere that uses `makeRequest`.

I'm wondering if sticking with `response.responseJSON = json;` for now is the way to go as it seems to me that doing it any other way would be pretty invasive.

marcoow: Hm, maybe it's possible to rewind the already read properties? I **think** I remember we did that in some project where we needed to read multiple times… Or maybe we can clone the response before reading and then resolve with the original one?

The problem with adding a `responseJSON` property is that changes the API from _"returns `fetch`'s response"_ to _"returns `fetch`'s response with an additional `responseJSON`"_ property. That might seem trivial but it's something we need to maintain over time as well as something that could break, e.g. if `fetch` ever introduced its own `responseJSON` property that was somehow different from ours though.

gtb104: I'm not sure how to "rewind" the read properties, but we can `clone` the response. The only real issue with doing that is all the places that use `makeRequest` have to be able to handle JSON or Response data, which makes this PR more invasive than I intended. I'll see how `clone` pans out.

I'm not sure the direction that this project is going, but I could see a case for maintaining `responseJSON` until the next major release. At that point the API could be changed where the response is only a `Response` object.

marcoow: This change will introduce what's going to be supported in the next major release so we should make sure it's what we want to maintain for the long term. I'd clone the `response` - that seems to be a pretty reasonable approach.

marcoow:
Sorry, I just realized this doesn't actually work. The authenticator's `authenticate` method is called by the session (when `sessionService.authenticate('authenticator:my-authenticator', …)` is invoked). The session expects the authenticator to return a promise that fulfills with a POJO which will then be the session data. That is the same for all authenticators and is part of the authenticator API. Of course, not all authenticators will necessarily make a request (you could have an authenticator that wraps the [Web Authentication API](https://developer.mozilla.org/en-US/docs/Web/API/Web_Authentication_API) for example). So that means we cannot expect the `authenticate` method to resolve with a request object and we also cannot have different authenticators resolve with different values since then the session can no longer work with arbitrary authenticators. That's different in the case when the returned promise rejects since then the error that the session received is simply forwarded to the caller of `sessionService.authenticate(…)` which of course knows which concrete authenticator is used in this invocation and can handle the concrete error accordingly.

Sorry I only realized this now after you put in substantial work already - should have brought this up much earlier 🤦‍♂

Issue #1799, opened 2019-04-02
Deprecation warnings on Ember 3.9


richard-viney:
On Ember 3.9:

`DEPRECATION: Setting a computed property as volatile has been deprecated. Instead, consider using a native getter with native class syntax. [deprecation id: computed-property.volatile] See https://emberjs.com/deprecations/v3.x#toc_computed-property-volatile for more details.`

marcoow:
Yes, this should be refactored 👍

richard-viney:
Looking at https://github.com/emberjs/rfcs/blob/master/text/0370-deprecate-computed-volatile.md it suggests that fixing this deprecation will require transitioning to native classes. However, after looking at the three places volatiles are used in this addon, I think they can just be turned into function calls?

This would be a public API change on `BaseAuthenticator.tokenRefreshOffset`, the other two places are in private APIs.

richard-viney:
It looks like there's also a deprecation warning from `computed-property.override` due to `_isFastBoot` sometimes being set to `false` but having a default initialisation of `reads('_fastBoot.isFastBoot')`.

josemarluedke:
Hitting this as well. Just installed the addon and got this deprecation.

wallacyalva:
the same happens to me

chrism:
Hi, I'm getting two deprecations mentioned in the original post and another one relating to `_isFastBoot`.

```
DEPRECATION: The <(unknown):ember196>#_isFastBoot computed property was just overriden. This removes the computed property and replaces it with a plain value, and has been deprecated. If you want this behavior, consider defining a setter which does it manually. [deprecation id: computed-property.override] See https://emberjs.com/deprecations/v3.x#toc_computed-property-override for more details.
```

Thanks.

charuwts:
Excited to see these being addressed, is there a status on when ESA will have a release with these fixes?

jordpo:
having the same issue with the deprecations - any news on this?

ryedeer:
[...]

PR #1829, opened 2019-04-22
Fix volatile deprecations

herzzanu:
Handles this https://deprecations.emberjs.com/v3.x/#toc_computed-property-volatile as well as removing the unneeded `ember-getowner-polyfill`.

### DIFF
--- a/addon/authenticators/oauth2-password-grant.js
+++ b/addon/authenticators/oauth2-password-grant.js
@@ -109,21 +109,17 @@ export default BaseAuthenticator.extend({
     multiple tabs, we need to prevent the tabs from sending refresh token
     request at the same exact moment.
 
-    __When overriding this property, make sure to mark the overridden property
-    as volatile so it will actually have a different value each time it is
-    accessed.__
-
     @property tokenRefreshOffset
     @type Integer
     @default a random number between 5 and 10
     @public
   */
-  tokenRefreshOffset: computed(function() {
+  get tokenRefreshOffset() {
     const min = 5;
     const max = 10;
 
     return (Math.floor(Math.random() * (max - min)) + min) * 1000;
-  }).volatile(),
+  },
 
   _refreshTokenTimeout: null,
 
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -38,7 +38,6 @@
     "ember-cli-is-package-missing": "^1.0.0",
     "ember-cookies": "^0.3.0 || ^0.4.0",
     "ember-fetch": "^2.1.0 || ^3.0.0 || ^4.0.0 || ^5.0.0 || ^6.0.0",
-    "ember-getowner-polyfill": "^1.1.0 || ^2.0.0",
```
marcoow: This is needed as we support very old versions of Ember still (1.12 😱)

herzzanu: Oops, I realised that too late, let me add it back :)

marcoow:
@herzzanu can you also resolve the conflict in `yarn.lock`?

marcoow:
@herzzanu: this still doesn't seem to work though… Have you had a look at why the tests are failing?

herzzanu:
Seems tests are passing only fro ember 3.9 and fail for any previous versions. Looking for a solution.

josemarluedke:
Hitting this deprecation as well. Anything I can do to move this PR further ?

jordpo:
The test failures appear unrelated. Can you revert your changes to the yarn.lock file and see if that fixes anything? @herzzanu

herzzanu:
@jordpo no luck unfortunately :/

krasnoukhov:
This will fix #1799

jordpo:
I'm not sure you can use `get` with the classic syntax. Is there documentation that says that we can?

herzzanu:
I don't see why not. https://deprecations.emberjs.com/v3.x/#toc_computed-property-volatile
Apparently it seems to only work on ember >= 3.9

Turbo87:
resolved by #1885 and #1886

PR #1839, opened 2019-05-09
Migrate away from volatile cps to methods

scottkidder:
Volatile computed properties are now deprecated in 3.9.

This PR transitions each of the existing volatile computed properties into methods. Only one of them was public, so the volatile computed property still exists with a deprecation message (I didn't put much though into the deprecation message, documentation link, version, etc).

Comments appreciated!

### DIFF
--- a/addon/authenticators/oauth2-password-grant.js
+++ b/addon/authenticators/oauth2-password-grant.js
@@ -104,11 +104,6 @@ export default BaseAuthenticator.extend({
   refreshAccessTokens: true,
 
   /**
-    The offset time in milliseconds to refresh the access token. This must
-    return a random number. This randomization is needed because in case of
-    multiple tabs, we need to prevent the tabs from sending refresh token
-    request at the same exact moment.
-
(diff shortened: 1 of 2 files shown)


backspace:
Thanks for opening this, I saw #1799 and thought I’d make a PR myself, but you already did!

I wonder, though, whether it would be preferable to use `get tokenRefreshOffset` and the like, as suggested in the [deprecation documentation](https://deprecations.emberjs.com/v3.x/#toc_computed-property-volatile). For example, I [did this yesterday](https://github.com/backspace/prison-rideshare-ui/commit/6d3504f68b45aebeb8a95c42029da8882afe95df) in a project I maintain.

I could open another PR that uses this approach if you don’t have time at the moment.

backspace:
#1829 does it that way, actually!

scottkidder:
I didn't use that approach to maintain backwards compatibility but it would be nice!

Turbo87:
resolved by #1885 and #1886

PR #1841, opened 2019-05-11
Remove computed property overrides

backspace:
This fixes the deprecation warnings described here, from Ember 3.8:
https://deprecations.emberjs.com/v3.x/#toc_computed-property-override

You can see in the Travis logs that the warnings are gone by comparing the log of [this most recent run on `master`](https://travis-ci.org/simplabs/ember-simple-auth/jobs/530936734), which has 222 instances of “overriden” (sic), to the log of [this PR run](https://travis-ci.org/backspace/ember-simple-auth/jobs/531228151), which has 0!

I chose to follow a pattern of adding a property to `this` with the same attribute name with `Override` appended. This allows the `get` to check if the overridden property exists before sending the default if it doesn’t. It’s repetitive and a bit unwieldy. Since it’s the same pattern for each computed property, it could be extracted. Let me know if you think that’s desirable or if there are any other changes you’d like to see.

Thanks for this project, it’s great to have around!

### DIFF
--- a/addon/mixins/application-route-mixin.js
+++ b/addon/mixins/application-route-mixin.js
@@ -66,8 +66,19 @@ export default Mixin.create({
     @default 'index'
     @public
   */
-  routeAfterAuthentication: computed(function() {
-    return Configuration.routeAfterAuthentication;
+  routeAfterAuthentication: computed({
+    get() {
+      if (this.hasOwnProperty('routeAfterAuthenticationOverride')) {
+        return this.routeAfterAuthenticationOverride;
+      } else {
+        return Configuration.routeAfterAuthentication;
+      }
+    },
+
+    set(key, value) {
+      this.routeAfterAuthenticationOverride = value;
+      return value;
+    }
   }),
 
   init() {
--- a/addon/mixins/application-route-mixin.js
+++ b/addon/mixins/application-route-mixin.js
@@ -1,6 +1,5 @@
 import Mixin from '@ember/object/mixin';
 import { A } from '@ember/array';
-import { computed } from '@ember/object';
 import { getOwner } from '@ember/application';
 import { inject } from '@ember/service';
 import Ember from 'ember';
--- a/addon/mixins/application-route-mixin.js
+++ b/addon/mixins/application-route-mixin.js
@@ -66,20 +66,7 @@ export default Mixin.create({
     @default 'index'
     @public
   */
-  routeAfterAuthentication: computed({
-    get() {
-      if (this.hasOwnProperty('routeAfterAuthenticationOverride')) {
-        return this.routeAfterAuthenticationOverride;
-      } else {
-        return Configuration.routeAfterAuthentication;
-      }
-    },
-
-    set(key, value) {
-      this.routeAfterAuthenticationOverride = value;
-      return value;
-    }
-  }),
+  routeAfterAuthentication: Configuration.routeAfterAuthentication,
 
   init() {
     this._super(...arguments);
--- a/addon/mixins/authenticated-route-mixin.js
+++ b/addon/mixins/authenticated-route-mixin.js
@@ -62,22 +62,6 @@ export default Mixin.create({
   */
   session: service('session'),
 
-  _router: computed({
-    get() {
-      if (this.hasOwnProperty('_routerOverride')) {
-        return this._routerOverride;
-      } else {
-        let owner = getOwner(this);
-        return owner.lookup('service:router') || owner.lookup('router:main');
-      }
-    },
-
-    set(key, value) {
-      this._routerOverride = value;
-      return value;
-    }
-  }),
-
   _isFastBoot: isFastBootCPM(),
 
   /**
@@ -105,6 +89,12 @@ export default Mixin.create({
       return value;
     }
   }),
+
+  init() {
+    this._super(...arguments);
+    let owner = getOwner(this);
+    this._router = owner.lookup('service:router') || owner.lookup('router:main');
+  },
 
   /**
     Checks whether the session is authenticated and if it is not aborts the
--- a/addon/mixins/authenticated-route-mixin.js
+++ b/addon/mixins/authenticated-route-mixin.js
@@ -1,7 +1,6 @@
 import { inject as service } from '@ember/service';
 import Mixin from '@ember/object/mixin';
 import { assert } from '@ember/debug';
-import { computed } from '@ember/object';
 import { getOwner } from '@ember/application';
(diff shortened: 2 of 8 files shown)


richard-viney:
Great! As you point out, doing this with a computed property macro would end up being less verbose.

backspace:
Agreed, that’s my preference, but I’m not a maintainer so I’ll wait for confirmation that they’d be into having that kind of abstraction in the codebase 😀

addon/mixins/application-route-mixin.js
```diff
@@ -66,8 +66,19 @@ export default Mixin.create({
...
+      } else {
+        return Configuration.routeAfterAuthentication;
+      }
+    },
+
+    set(key, value) {
+      this.routeAfterAuthenticationOverride = value;
+      return value;
+    }
   }),
```
marcoow: Could this not be replaced with

```js
routeAfterAuthentication: Configuration.routeAfterAuthentication,
```

simply?  I can't remember why this was made a CP in the first place…

addon/mixins/authenticated-route-mixin.js
```diff
@@ -80,10 +75,27 @@ export default Mixin.create({
...
+      return value;
+    }
   }),
 
+  init() {
+    this._super(...arguments);
+    let owner = getOwner(this);
+    this._router = owner.lookup('service:router') || owner.lookup('router:main');
+  },
+
```
backspace: It looked to me that in other files, `init` was placed after the various computed property definitions, but it’s easily moved if I guessed wrong!

backspace:
I made those changes but maybe `getOwner(this)` doesn’t work inside `init`? I’ve run out of time for this at the moment but I’ll see if I can figure it out at some point 😞

backspace:
There’s more than one thing wrong but it’s true that `getOwner` won’t work in a mixin because `this` doesn’t yet exist. I looked at the arguments and the first one for a route mixin is `store`, so `getOwner(store)` could work there, but that seems scandalous 😳

addon/mixins/authenticated-route-mixin.js
```diff
@@ -80,10 +75,27 @@ export default Mixin.create({
...
+    set(key, value) {
+      this.authenticationRouteOverride = value;
+      return value;
+    }
   }),
 
+  init() {
+    this._super(...arguments);
+    let owner = getOwner(this);
+    this._router = owner.lookup('service:router') || owner.lookup('router:main');
```
basicBrogrammer: Can't we just inject the router service now ?

backspace: I don’t think so, given that this addon is compatible back to Ember 1.12.

addon/mixins/authenticated-route-mixin.js
marcoow: Yes, I think that's what we've been doing so far 👍

addon/mixins/authenticated-route-mixin.js
marcoow: We should release a 2.0 soon to get rid of all this ancient stuff 😞

marcoow:
I think using `getOwner(this)` in `init` in a mixin should be fine **after** `this._super(...arguments);`. I'm not really sure what the problem is here…

addon/mixins/authenticated-route-mixin.js
marcoow: @backspace the problem with this line is that it will always call `owner.lookup`, even when the test overrides the `_router` property (which is a problem in itself but shouldn't be refactored as part of this PR), e.g. https://github.com/simplabs/ember-simple-auth/blob/master/tests/unit/mixins/authenticated-route-mixin-test.js#L64

If you change this to

```js
this._router = this._router || owner.lookup('service:router') || owner.lookup('router:main');
```

it works as expected.

addon/mixins/authenticated-route-mixin.js
```diff
@@ -80,10 +75,27 @@ export default Mixin.create({
...
+        return this.authenticationRouteOverride;
+      } else {
+        return Configuration.authenticationRoute;
+      }
+    },
+
+    set(key, value) {
+      this.authenticationRouteOverride = value;
+      return value;
+    }
```
marcoow: I think this can just be

```js
authenticationRoute: Configuration.authenticationRoute,
```

addon/mixins/unauthenticated-route-mixin.js
```diff
@@ -62,8 +62,19 @@ export default Mixin.create({
...
+        return this.routeIfAlreadyAuthenticatedOverride;
+      } else {
+        return Configuration.routeIfAlreadyAuthenticated;
+      }
+    },
+
+    set(key, value) {
+      this.routeIfAlreadyAuthenticatedOverride = value;
+      return value;
+    }
```
marcoow: I think this can just be

```js
routeIfAlreadyAuthenticated: Configuration.routeIfAlreadyAuthenticated,
```

backspace:
I can’t explain it but I made 4f48367 to add some logging that shows that `getOwner(this)` is strangely returning `false` in `init`, you can see it [here in this Travis job](https://travis-ci.org/backspace/ember-simple-auth/jobs/551113299#L2345-L2346). But maybe it’s a red herring, as I think 18db3c4 addressed the failures.

I needed 4379cf3 to get the lint and floating dependencies tests running on Travis, maybe some dependency has dropped support for Node 6? I’ve since removed it, but before that, [the tests were passing](https://travis-ci.org/backspace/ember-simple-auth/builds/551115563) apart from the FastBoot one but the [failures](https://travis-ci.org/backspace/ember-simple-auth/jobs/551115578#L576-L577) look the same as [those](https://travis-ci.org/simplabs/ember-simple-auth/jobs/550143418#L1014-L1015) on the most recent `master` job.

I’m unsure about the Node 6 vs 8 thing but could easily [open a PR](https://github.com/simplabs/ember-simple-auth/compare/master...backspace:travis-node-8?expand=1) to change that if it’s desirable.

josemarluedke:
These changes look good! Any change to get them merged soon? Thanks.

josemarluedke:
I just came across a problem with `routeAfterAuthentication` not working in an Octane app. Using the changes in this branch in conjunction with changes from master, the problem goes away.

Turbo87:
closing in favor of https://github.com/simplabs/ember-simple-auth/pull/1900

josemarluedke:
I believe  #1900 was just a subset of the changes here. There are a few more deprecations going on.

PR #1859, opened 2019-06-23
Replaced .volatile properties with native getters

alexmasita:
My app throws the following deprecation notice.

```
deprecate.js:120 DEPRECATION: Setting a computed property as volatile has been deprecated. Instead, consider using a native getter with native class syntax. [deprecation id: computed-property.volatile] See https://emberjs.com/deprecations/v3.x#toc_computed-property-volatile for more details.
```
The pull request converts them to native getters.

### DIFF
--- a/addon/session-stores/cookie.js
+++ b/addon/session-stores/cookie.js
@@ -141,23 +141,21 @@ export default BaseStore.extend({
     return owner && owner.lookup('service:fastboot');
   }),
 
-  _secureCookies: computed(function() {
+  get _secureCookies() {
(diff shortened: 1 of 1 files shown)


marcoow:
Thanks for the effort but I'm closing this as a duplicate of #1829 and #1839. Unfortunately the approach can't work anyway though 😞

alexmasita:
Understood.  I see the backward compatibility issue.  I see a situation where you might want to have two repos, one for pre <= Ember 3.8 and one after >= Ember 3.9.

Issue #1838, opened 2019-05-08
Following the documentation, after login, the transition is triggered before the load of current user promise is resolved


remiuz:
We've integrated ember-simple-auth to our application. Following, the documentation, we load the user from the sessionAuthenticated callback :

```javascript
// app/routes/application.js
import Ember from 'ember';
import ApplicationRouteMixin from 'ember-simple-auth/mixins/application-route-mixin';

const { service } = Ember.inject;

export default Ember.Route.extend(ApplicationRouteMixin, {
  currentUser: service(),

  beforeModel() {
    return this._loadCurrentUser();
  },

  sessionAuthenticated() {
    this._super(...arguments);
    this._loadCurrentUser();
  },

  _loadCurrentUser() {
    return this.get('currentUser').load().catch(() => this.get('session').invalidate());
  }
});
```

In the example, you call this._super(...arguments); in sessionAuthenticated() and then you call the _loadCurrentUser() function. If we do that in our application, the transition will be achieved by the sessionAuthenticated() function from ApplicationRouteMixin before the user will be loaded. So the route model can be resolved before the user is loaded...

From our side, it was causing unexpected behaviors. 

We fixed in our project it in this way : 

```javascript
// app/routes/application.js
import Ember from 'ember';
import ApplicationRouteMixin from 'ember-simple-auth/mixins/application-route-mixin';

const { service } = Ember.inject;

export default Ember.Route.extend(ApplicationRouteMixin, {
  currentUser: service(),

  beforeModel() {
    return this._loadCurrentUser();
  },

  sessionAuthenticated() {
    const _super = this._super;

    this._loadCurrentUser()
      .then(() => {
        _super.call(this, ...arguments);
      });
  },

  _loadCurrentUser() {
    return this.get('currentUser').load().catch(() => this.get('session').invalidate());
  }
});
```

marcoow:
Hm, I see what you mean - that makes sense, yes. I think we'd want this in the guides:

```js
async sessionAuthenticated() {
  await this._loadCurrentUser();

  return this.super(...arguments);
},
```

A PR would be much appreciated!

nbenz:
I submitted a PR to update the documentation ^

This fixed some really gnarly `visit()` behavior in my acceptance tests, so I'm glad I stumbled across this issue.

gilles-gosuin:
I'm not sure documenting this is a good idea; it does not seem to solve my colleague @remiuz's issue.

This works like a charm:
```
[...]

PR #1887, opened 2019-07-08
[Docs] use async/await when loading current user

nbenz:
use async/await when loading current user so that the post-authentication transition won't happen until the user has finished loading

https://github.com/simplabs/ember-simple-auth/issues/1838

### DIFF
--- a/guides/managing-current-user.md
+++ b/guides/managing-current-user.md
@@ -150,9 +150,9 @@ export default Ember.Route.extend(ApplicationRouteMixin, {
     return this._loadCurrentUser();
   },
 
-  sessionAuthenticated() {
+  async sessionAuthenticated() {
+    await this._loadCurrentUser();
     this._super(...arguments);
-    this._loadCurrentUser();
   },
 
   _loadCurrentUser() {


guides/managing-current-user.md
```diff
@@ -146,13 +146,14 @@ const { service } = Ember.inject;
...
+    await this._loadCurrentUser();
+    return this._super(...arguments);
   },
 
-  sessionAuthenticated() {
-    this._super(...arguments);
-    this._loadCurrentUser();
+  async sessionAuthenticated() {
+    await this._loadCurrentUser();
+    return this._super(...arguments);
```
marcoow: I think the `return` is misleading as `_super` doesn't return anything and you'd probably want to call `_super` first.

nbenz: I copied this from your PR comment. :) but looking at the source code, I agree the `return` is unnecessary. Calling `_super` first is actually the bug this is trying to fix. You need to wait for the user to load before you have the mixin process whatever transition is currently happening.

nbenz: I also see in the [`beforeModel`](https://api.emberjs.com/ember/3.10/classes/Route/methods/beforeModel?anchor=beforeModel) docs that it will already pause transitions when a Promise is returned, so I think the current documentation was fine there.

marcoow: Ah, sorry - you're right of course 🤦‍♂

PR #1900, opened 2019-07-16
Rename `_router` property to `_authRouter` to avoid conflicts

backspace:
This is a subset of #1841 as requested by @locks:
https://discordapp.com/channels/480462759797063690/491905849405472769/600683165929177091


backspace:
It appears that if you restart a job that’s part of a stage that failed and it succeeds on the second attempt, the next stage doesn’t run 😬

pichfl:
@backspace I restarted the build on Travis 👍

backspace:
👍 understood, done, seems to be working!

PR #1917, opened 2019-07-31
Fix computed property override deprecations

josemarluedke:
This fixes computed property override deprecations. Initial work was done in #1841 but not merged.

### DIFF
--- a/addon/mixins/application-route-mixin.js
+++ b/addon/mixins/application-route-mixin.js
@@ -1,6 +1,5 @@
 import Mixin from '@ember/object/mixin';
 import { A } from '@ember/array';
-import { computed } from '@ember/object';
 import { getOwner } from '@ember/application';
 import { inject } from '@ember/service';
 import Ember from 'ember';
@@ -66,9 +65,7 @@ export default Mixin.create({
     @default 'index'
     @public
   */
-  routeAfterAuthentication: computed(function() {
-    return Configuration.routeAfterAuthentication;
-  }),
+  routeAfterAuthentication: Configuration.routeAfterAuthentication,
 
   init() {
     this._super(...arguments);
--- a/addon/mixins/authenticated-route-mixin.js
+++ b/addon/mixins/authenticated-route-mixin.js
@@ -80,9 +80,7 @@ export default Mixin.create({
     @default 'login'
     @public
   */
-  authenticationRoute: computed(function() {
-    return Configuration.authenticationRoute;
-  }),
+  authenticationRoute: Configuration.authenticationRoute,
 
   /**
     Checks whether the session is authenticated and if it is not aborts the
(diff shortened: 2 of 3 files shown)


addon/session-stores/adaptive.js
```diff
@@ -105,24 +105,20 @@ export default Base.extend({
...
-  _isLocalStorageAvailable: computed(function() {
-    try {
-      localStorage.setItem(LOCAL_STORAGE_TEST_KEY, true);
-      localStorage.removeItem(LOCAL_STORAGE_TEST_KEY);
-      return true;
-    } catch (e) {
-      return false;
-    }
-  }),
-
```
Turbo87: why did you convert these computed properties? where are they overridden?

addon/session-stores/cookie.js
```diff
@@ -135,12 +135,6 @@ export default BaseStore.extend({
 
   _cookies: service('cookies'),
 
-  _fastboot: computed(function() {
-    let owner = getOwner(this);
-
-    return owner && owner.lookup('service:fastboot');
-  }),
-
```
Turbo87: same as above, what is overriding this property?

addon/utils/is-fastboot.js
```diff
@@ -10,8 +10,18 @@ import ApplicationInstance from '@ember/application/instance';
...
+      if (this.hasOwnProperty('isFastBootOverride')) {
+        return this.isFastBootOverride;
+      } else {
+        return isFastBoot(getOwner(this));
+      }
+    },
+    set(key, value) {
+      this.isFastBootOverride = value;
+      return value;
+    }
```
Turbo87: I don't understand this change. Why would something call `this.set('isFastboot`, true/false)`? Seems to me that this kind of thing won't change during the runtime of the app 🤔

josemarluedke:
@Turbo87 I have pretty much implemented the same thing as #1841. However, I do agree with the change.

Applications will probably not override these 3 cases you mentioned, but, we override these properties in tests. Looking at the console while running tests, you will see that are many deprecations. 

IMO, if Ember is deprecating something, the internal tests of a lib should not be relying on them.

Happy to revert these changes if you think we don't need them.

Turbo87:
> we override these properties in tests

I don't think anything that is running in the browser should override the `isFastboot` flag to be `true`. Also, these flags are **private** properties and implementation details, they could change at any time so relying on them seems very brittle to me.

josemarluedke:
@Turbo87 Sorry for my wording, but by "we" I meant ember-simple-auth. Ember Simple Auth does override these properties in tests.

Turbo87:
oh, okay! sorry, wasn't aware of that. 😅 

I would still like to avoid making the production code more complicated just so that we can better mock things in the tests 🤔

josemarluedke:
@Turbo87 all right, removed the other changes. Keet only the changes around the configuration.

josemarluedke:
@Turbo87 @marcoow Ping. :D

Issue #1908, opened 2019-07-21
Deprecation warnings on Ember 3.9 (computed property was just overriden)


krachtstefan:
As mentioned in issue https://github.com/simplabs/ember-simple-auth/issues/1799, there is stil one deprecation warning with Ember 3.9:

`DEPRECATION: The <(unknown):ember242>#_isFastBoot computed property was just overriden. This removes the computed property and replaces it with a plain value, and has been deprecated. If you want this behavior, consider defining a setter which does it manually.`

The stacktrace is referencing the following line

https://github.com/simplabs/ember-simple-auth/blob/76e0eacf94f1d1d4d1194b3ef93ceae0bc5e000e/addon/session-stores/adaptive.js#L130

Turbo87:
@marcoow any idea how that could happen? setting a property on a freshly created `{}` object shouldn't cause any CPs to be overwritten 🤔

marcoow:
Yes, we're still using the `isFastBoot` CP macro (e.g. https://github.com/simplabs/ember-simple-auth/blob/d6071274d0cb8e3855e24bf2687c1a6395379440/addon/session-stores/local-storage.js#L29) - that should be relatively easy to replace though as there should be no need for that to be a CP anyway

Turbo87:
@marcoow I think the problem here is not that it's a CP, the problem is that we're overriding a property from a subclass with something potentially unrelated. we shouldn't be overriding at all without a good reason.

josemarluedke:
What is the reason for having this set `options._isFastBoot = false` assigned here anyway? Could that be just removed?

bhabanipetnet:
hello. Could we have an update in this `#_isFastBoot computed property` issue?

robclancy:
This was reported in 3.9 and now we are on 3.13. Ember doesn't even release that often.

vmalloc:
How easy is it to fix? I can try and craft a PR but I'm not 100% sure I understand what this needs to change to...

jordpo:
Also wondering about the status of this. Thanks!

devotox:
This was not fixed in any of the new versions.

After tracing the problem is that `options._isFastBoot` is passed into the `createStore` functions

vmalloc:
Just got this again after trying to upgrade to 2.0.0:
```
[...]

PR #1938, opened 2019-08-29
Use a computedProperty with get set.

colenso:
Since the `_isFastBoot` computedProperty is overridden in some places, I've used added a setter as per https://deprecations.emberjs.com/v3.x/#toc_computed-property-override

### DIFF
--- a/addon/utils/is-fastboot.js
+++ b/addon/utils/is-fastboot.js
@@ -5,13 +5,22 @@ import ComputedProperty from '@ember/object/computed';
 import { getOwner } from '@ember/application';
 import { assert } from '@ember/debug';
 import ApplicationInstance from '@ember/application/instance';
+import { isPresent } from '@ember/utils';
 
 /**
  * @return {ComputedProperty<boolean>}
  */
 export default function isFastBootCPM() {
-  return computed(function() {
-    return isFastBoot(getOwner(this));
+  return computed({
+    get() {
+      if (isPresent(this._isFastBootCPM)) {
+        return this._isFastBootCPM;
+      }
+      return isFastBoot(getOwner(this));
+    },
+    set(key, value) {
+      return this._isFastBootCPM = value;
+    }
   });
 }
 
--- a/addon/utils/is-fastboot.js
+++ b/addon/utils/is-fastboot.js
@@ -19,7 +19,8 @@ export default function isFastBootCPM() {
       return isFastBoot(getOwner(this));
     },
     set(key, value) {
-      return this._isFastBootCPM = value;
+      this.set('_isFastBootCPM',value);
+      return this._isFastBootCPM;
     }
   });
 }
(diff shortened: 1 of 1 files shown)


barelyknown:
Related to https://github.com/simplabs/ember-simple-auth/issues/1908

josemarluedke:
@Turbo87 @marcoow Any change for you to have a look at this?

Turbo87:
@josemarluedke so far nobody was able to explain why setting a property on a regular `{}` object causes a computed property to be overwritten. without understanding what is going on I am hesitant to merge a PR like this.

josemarluedke:
@Turbo87 I don't know either, I just know that this deprecation happens and it seems so simple to solve. Maybe some other folks can help us understand why this deprecation happens in this case. cc/ @pzuraq @rwjblue.

For reference: 
https://github.com/simplabs/ember-simple-auth/blob/711740610c5a10e72b3de3cc2b18b84430d60a98/addon/session-stores/adaptive.js#L130

pzuraq:
It looks like those options get passed to `LocalStorage.create` here: https://github.com/simplabs/ember-simple-auth/blob/76e0eacf94f1d1d4d1194b3ef93ceae0bc5e000e/addon/session-stores/adaptive.js#L145

This means that it will be assigned to the instance of the `LocalStore` class, similar to:

```js
class LocalStorage {
  static create(props) {
	let instance = new this();

    Object.assign(instance, props);
  }
}
```

So when it goes to assign the props, it's assigning `_isFastBoot` on the class, which is a CP, to `false`. Since the `_isFastBoot` is a CP, this triggers the CPs defaulting behavior (clobber, effectively erasing the CP), but we deprecated that because it's not the way native getters/setters work.

Turbo87:
ugh, okay, that makes sense. now I'm wondering why we're assigning the `_isFastBoot` flag at all if we have a proper CP in the class anyway.

git blame points to https://github.com/simplabs/ember-simple-auth/commit/75cd83b239f1950916450e1e4e89ac7056d823d9, but that commit is too big to actually be useful.

@marcoow any idea?

marcoow:
I think what's happening is that the adaptive store simply passes on all options that it receives to the concrete store it creates (based on whether `localStorage` is available or not). The tests simply set `_isFastBoot` as an easy way to simulate FastBoot or not. That can all be improved for sure and should be eventually but I think for now the easiest fix would simply be to do this in the `init`s of classes that currently use the util:

```js
this._isFastBoot = options._isFastBoot || isFastBoot();
```

That way could remove all of these CPs which don't really need to be CPs anyway as they never change anyway without having to change test tests (which can be improved later then).

xomaczar:
@marcoow would you like a new PR where `this._isFastBoot = options._isFastBoot || isFastBoot();` is used instead of CP with getter/setter like this PR is trying to accomplish?

marcoow:
@xomaczar yes, that would be great - sorry for the slow response 😞

PR #1941, opened 2019-09-09
Update with async/await example

geekygrappler:
Hello 👋,

I thought this was a nicer example because 😍async/await 😍.

### DIFF
--- a/README.md
+++ b/README.md
@@ -210,11 +210,17 @@ export default Ember.Controller.extend({
   session: Ember.inject.service('session'),
 
   actions: {
-    authenticate() {
+    async authenticate() {
       let { identification, password } = this.getProperties('identification', 'password');
-      this.get('session').authenticate('authenticator:oauth2', identification, password).catch((reason) => {
-        this.set('errorMessage', reason.error || reason);
-      });
+      try {
+        await this.session.authenticate('authenticator:oauth2', identification, password);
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -210,11 +210,17 @@ export default Ember.Controller.extend({
...
 
   actions: {
-    authenticate() {
+    async authenticate() {
       let { identification, password } = this.getProperties('identification', 'password');
-      this.get('session').authenticate('authenticator:oauth2', identification, password).catch((reason) => {
+      try {
+        await this.session.authenticate('authenticator:oauth2', identification, password);
+      } catch(e) {
         this.set('errorMessage', reason.error || reason);
```
pichfl: You missed converting the `reason` from the `.catch()` to the `catch (error) `.

```js
...
} catch (error) {
    this.set('errorMessage', error.error || error);
}

File: README.md [hunk 2]
```diff
@@ -210,11 +210,17 @@ export default Ember.Controller.extend({
...
-      });
+      }
+
+      if (this.session.isAuthenticated) {
+        // What to do with all this success?
+      }
```
marcoow: This *could* be in the `try` as in the case that authentication fails, an error would be thrown (which is not a great API 😞).

README.md
marcoow: I fixed it

PR #1956, opened 2019-09-25
Drop support for Node 6

marcoow:
…as it is EOL


marcoow:
Yes, this will be part of a 2.0 release. @sdebarros had included it in #1938 but merging this separately now allows us to get all the currently broken @dependabot-bot PRs in.

PR #1985, opened 2019-10-28
deprecate rejectWithResponse=false

marcoow:
This deprecates relying on the current default value of `false` for the `rejectWithResponse` in the OAuth 2.0 Password Grant and Devise authenticators. We want to remove the property in the future and need everyone to switch to the new behavior before we can do that. That deprecation should have been added long ago already 🤦‍♂

### DIFF
--- a/tests/unit/authenticators/devise-test.js
+++ b/tests/unit/authenticators/devise-test.js
@@ -1,4 +1,5 @@
 import { tryInvoke } from '@ember/utils';
+import { registerDeprecationHandler } from '@ember/debug';
 import {
   describe,
   beforeEach,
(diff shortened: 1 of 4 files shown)


marcoow:
@sdebarros we'll have to release this before #1933 - I only realized this was missing when reviewing your PR.

tests/unit/authenticators/devise-test.js
```diff
@@ -62,6 +63,25 @@ describe('DeviseAuthenticator', () => {
...
+      let warnings;
+      registerDeprecationHandler((message, options, next) => {
+        // in case a deprecation is issued before a test is started
+        if (!warnings) {
+          warnings = [];
+        }
+
+        warnings.push(message);
+        next(message, options);
+      });
```
sdebarros:
```
      let warnings = [];
      registerDeprecationHandler((message, options, next) => {
        warnings.push(message);
        next(message, options);
      });
```

tests/unit/authenticators/oauth2-password-grant-test.js
```diff
@@ -196,6 +196,32 @@ describe('OAuth2PasswordGrantAuthenticator', () => {
...
+      let warnings;
+      registerDeprecationHandler((message, options, next) => {
+        // in case a deprecation is issued before a test is started
+        if (!warnings) {
+          warnings = [];
+        }
+
+        warnings.push(message);
+        next(message, options);
+      });
```
sdebarros: Same here.

tests/unit/authenticators/devise-test.js
marcoow: fixed 👍

tests/unit/authenticators/oauth2-password-grant-test.js
marcoow: fixed 👍

PR #1986, opened 2019-10-28
Deprecate data adapter mixin elements

marcoow:
This deprecates the `DataAdapterMixin`'s `authorizer` property and `headersForRequest` method that will effectively be removed in 2.0 with the removal on authorizers.

### DIFF
--- a/tests/unit/mixins/data-adapter-mixin-test.js
+++ b/tests/unit/mixins/data-adapter-mixin-test.js
@@ -1,4 +1,5 @@
 import EmberObject from '@ember/object';
+import { registerDeprecationHandler } from '@ember/debug';
 import { describe, beforeEach, it } from 'mocha';
 import { expect } from 'chai';
 import sinonjs from 'sinon';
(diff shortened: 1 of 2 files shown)


tests/unit/mixins/data-adapter-mixin-test.js
```diff
@@ -123,6 +124,23 @@ describe('DataAdapterMixin', () => {
   });
 
   describe('#headersForRequest', function() {
+    it('shows a deprecation warning', function() {
+      let warnings;
+      registerDeprecationHandler((message, options, next) => {
+        // in case a deprecation is issued before a test is started
```
sdebarros: I'm not sure what this comment means here since the `registerDeprecationHandler` is being used inside a test 🤔 

Also, wouldn't it be the same to do:
```
      let warnings = [];
      registerDeprecationHandler((message, options, next) => {
        warnings.push(message);
        next(message, options);
      });
```

marcoow: I think you're right, yes - I copied this from elsewhere and should have paid a bit more attention probably ;) - changed it now as you suggested

PR #1993, opened 2019-10-30
require Node 10+

### LINKED ISSUE #1992: Drop support for Node 8
…in preparation for ESA 3.0. As we'll be publishing another major version soon we should use the opportunity to drop support for Node 8 which will soon be EOL and dropping support will be a breaking change of course.

marcoow:
closes #1992

### DIFF
--- a/package.json
+++ b/package.json
@@ -97,7 +97,7 @@
     "torii": "^0.10.0"
   },
   "engines": {
-    "node": "8.* || 10.* || >= 12.*"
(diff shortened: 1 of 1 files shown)


marcoow:
The build has [succeeded already actually](https://travis-ci.org/simplabs/ember-simple-auth/builds/605030089?utm_source=github_status&utm_medium=notification) - just the status hasn't been reflected here for some reason.

PR #1995, opened 2019-10-30
Fix deprecation until versions

marcoow:
When deprecating the `rejectWithResponse` for the `Devise` and `OAuth2PasswordGrant` authenticators, we put the wrong `until` version in the deprecation - that version should be `3.0.0` of course, not `2.0.0` which is the version the deprecations were introduced with.

This also updates all previous deprecations that we took over from 1.x and didn't actually remove in 2.0.0 to now say `until: '3.0.0'`.

### DIFF
--- a/addon/authenticators/devise.js
+++ b/addon/authenticators/devise.js
@@ -84,7 +84,7 @@ export default BaseAuthenticator.extend({
   */
   rejectWithXhr: computed.deprecatingAlias('rejectWithResponse', {
     id: `ember-simple-auth.authenticator.reject-with-xhr`,
-    until: '2.0.0'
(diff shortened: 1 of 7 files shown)


sdebarros:
@marcoow we should have updated all deprecation notices when releasing 2.0 😳https://github.com/simplabs/ember-simple-auth/blob/ad52181f2995c53bddbc217776a179eadd3e40a0/addon/authenticators/devise.js#L87 for instance.

marcoow:
> @marcoow we should have updated all deprecation notices when releasing 2.0 😳

Ouch, you're right of course 🤦‍♂ I fixed it now.

Issue #1998, opened 2019-11-01
How to migrate from using authorize


AddisonG:
I've opened a ticket on stackoverflow asking about this issue: https://stackoverflow.com/questions/58653616/how-to-replace-the-authorize-method-in-ember-simple-auth

I'd like some further instruction on how to do acceptance tests using the `authenticateSession` method.

marcoow:
We should add some explanation in the README for this.

AddisonG:
Thanks @marcoow  and @sdebarros. Works perfectly.

PR #2010, opened 2019-11-07
Remove authorize assertion and update migration example

sdebarros:
Related to the StackOverflow question in https://github.com/simplabs/ember-simple-auth/issues/1998.

We are showing a deprecation warning that tells the user to use `headers` instead of `authorize`, but we also have an assert in `authorize` if it's not overridden, this causes an error when using ajax.

### DIFF
--- a/README.md
+++ b/README.md
@@ -528,7 +528,7 @@ export default Base.extend({
 ### Deprecation of Authorizers
 
 Authorizers and the session service's `authorize` method are deprecated and
-will be removed from Ember Simple Auth 2.0. The concept seemed like a good idea
+will be removed from Ember Simple Auth 3.0. The concept seemed like a good idea
 in the early days of Ember Simple Auth, but proved to provide limited value for
 the added complexity. To replace authorizers in an application, simply get the
 session data from the session service and inject it where needed.
@@ -550,7 +550,7 @@ import DataAdapterMixin from "ember-simple-auth/mixins/data-adapter-mixin";
 const { JSONAPIAdapter } = DS;
 
 export default JSONAPIAdapter.extend(DataAdapterMixin, {
-  session: service(), 
+  session: service(),
   authorize(xhr) {
     let { access_token } = this.get('session.data.authenticated');
     if (isPresent(access_token)) {
--- a/README.md
+++ b/README.md
@@ -525,43 +525,47 @@ export default Base.extend({
 });
 ```
 
-### Deprecation of Authorizers
+### Deprecation of Authorizers and DataAdapterMixin authorize method
 
-Authorizers and the session service's `authorize` method are deprecated and
-will be removed from Ember Simple Auth 3.0. The concept seemed like a good idea
-in the early days of Ember Simple Auth, but proved to provide limited value for
-the added complexity. To replace authorizers in an application, simply get the
-session data from the session service and inject it where needed.
+Authorizers and the session service's `authorize` method where deprecated in Ember Simple Auth 1.6
+and will be removed in Ember Simple Auth 3.0. The concept seemed like a good idea in the early days of
+Ember Simple Auth, but proved to provide limited value for the added complexity.
 
-In most cases, authorizers are used with Ember Data adapters (refer to the
-[Ember Guides](https://guides.emberjs.com/v3.0.0/models/customizing-adapters/#toc_headers-customization)
-for details on adapters). Replacing authorizers in these scenarios is
-straightforward.
+Originally, we recommended replacing the use of authorizers by defining an `authorize` method on
+the application adapter that could use our `session` service to inject the authorization information.
 
-Examples:
+As of Ember Simple Auth 2.0, we've decided to also deprecate this `authorize` method, and instead recomend
+the use of Ember Data's `headers` property on the `application` adapter for this use case.
+
+To inject the `Authorization` information on a request, simply define the `headers` computed property on your
+adapter and get the needed session data from the session service. This works for both AJAX and fetch requests.
 
 ```js
 // OAuth 2
 import DS from 'ember-data';
 import { inject as service } from '@ember/service';
-import { isPresent } from '@ember/utils';
+import { computed } from '@ember/object';
 import DataAdapterMixin from "ember-simple-auth/mixins/data-adapter-mixin";
 
 const { JSONAPIAdapter } = DS;
 
 export default JSONAPIAdapter.extend(DataAdapterMixin, {
   session: service(),
-  authorize(xhr) {
-    let { access_token } = this.get('session.data.authenticated');
-    if (isPresent(access_token)) {
-      xhr.setRequestHeader('Authorization', `Bearer ${access_token}`);
+  headers: computed('session.data.authenticated.access_token', function() {
+    let headers = {};
+
+    if (this.get('session.isAuthenticated')) {
+      headers['Authorization'] = `Bearer ${this.get('session.data.authenticated.token')}`;
     }
-  }
+
+    return headers;
+  }),
 });
 
 // DataAdapterMixin already injects the `session` service. It is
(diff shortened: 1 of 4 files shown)


addon/mixins/data-adapter-mixin.js
```diff
@@ -125,7 +125,9 @@ export default Mixin.create({
   },
 
   authorize() {
-    assert('The `authorize` method should be overridden in your application adapter. It should accept a single argument, the request object.');
+    let [ major, _ ] = Ember.VERSION.split(".");
+
+    assert('The `authorize` method should be overridden in your application adapter. It should accept a single argument, the request object.', major >= 2);
```
sdebarros: Ember pre 2.0 doesn't have the `headers` property, so this should still be replaced on those versions.

Another option is to use `warn` instead of `assert` to just show a warning to users, instead of throwing an error.

marcoow:
I think with #1933 we're actually going to drop support for Ember Data < 3.0.0?

addon/mixins/data-adapter-mixin.js
marcoow: I think this would have to check the Ember Data version, not the Ember version?

sdebarros:
It can also be annoying that we are showing the deprecation warning even if the user does use `headers` instead of `authorize`:
https://github.com/simplabs/ember-simple-auth/blob/4d080e35f71e763a0341d562bdb30b55c3f4c1c9/addon/mixins/data-adapter-mixin.js#L113 

I thought of a few options but none is perfect:
- Do `this.get('headers')` and not show the deprecation if it's not empty (or if it has a value for `Authorization`), this would still show the deprecation if the header is not included because the session is not authenticated.
- Make `authorize` return some special value, and then check if it doesn't (was overridden) and show the deprecation warning in that case

sdebarros:
@marcoow yes, but as it is right now, in version 2 we are suggesting the use of `headers` and throwing an error if someone does so while still using Ajax, which can be confusing and doesn't allow for the migration before upgrading to version 3.

If it's simpler, we can just explain that on the README.

addon/mixins/data-adapter-mixin.js
sdebarros: Yes 🤦‍♀

README.md
```diff
@@ -531,44 +531,48 @@ export default Base.extend({
...
-In most cases, authorizers are used with Ember Data adapters (refer to the
-[Ember Guides](https://guides.emberjs.com/v3.0.0/models/customizing-adapters/#toc_headers-customization)
-for details on adapters). Replacing authorizers in these scenarios is
-straightforward.
+Originally, we recommended replacing the use of authorizers by defining an `authorize` method on
+the application adapter that could use our `session` service to inject the authorization information.
 
-Examples:
+As of Ember Simple Auth 2.0, we've decided to also deprecate this `authorize` method, and instead recomend
+the user of Ember's `headers` property on the application adapter for this use case.
```
marcoow:
```suggestion
the use of Ember Data's `headers` property on the `application` adapter for this use case.
```

README.md
```diff
@@ -578,34 +582,22 @@ import DataAdapterMixin from "ember-simple-auth/mixins/data-adapter-mixin";
...
-      headers['Authorization'] = `Bearer ${this.session.data.authenticated.token}`;
-    }
+    headers['Authorization'] = `Token token="${token}", email="${email}"`;
 
     return headers;
   }),
 });
 ```
 
+* The `headers` property is available from Ember 2.0, for older versions you can continue to use `authorizers`.
```
marcoow:
```suggestion
* The `headers` property is available from Ember Data 2.0, for older versions you can continue to use the `authorize` method.
```

addon/mixins/data-adapter-mixin.js
```diff
@@ -125,7 +126,9 @@ export default Mixin.create({
   },
 
   authorize() {
-    assert('The `authorize` method should be overridden in your application adapter. It should accept a single argument, the request object.');
+    let [ major_version ] = DS.VERSION.split(".");
+
+    assert('The `authorize` method should be overridden in your application adapter. It should accept a single argument, the request object.', major_version >= 2);
```
marcoow:
```suggestion
    assert('The `authorize` method should be overridden in your application adapter. It should accept a single argument, the request object.', major_version <= 2);
```

Isn't this what we want since in newer versions you should be using `headers` instead?

tests/unit/mixins/data-adapter-mixin-test.js
```diff
@@ -53,10 +54,14 @@ describe('DataAdapterMixin', () => {
...
-        hash.beforeSend();
-      }).to.throw(/Assertion Failed/);
+      let [ major_version ] = DS.VERSION.split(".");
+
+      if (major_version < 2) {
+        expect(function() {
+          adapter.ajaxOptions();
+          hash.beforeSend();
+        }).to.throw(/Assertion Failed/);
+      }
```
marcoow: I think the whole test needs to be wrapped as otherwise it will fail as there are no assertions.

marcoow:
> It can also be annoying that we are showing the deprecation warning even if the user does use headers instead of authorize:
> I thought of a few options but none is perfect:
> * Do this.get('headers') and not show the deprecation if it's not empty (or if it has a value for Authorization), this would still show the deprecation if the header is not included because the session is not authenticated.
> * Make authorize return some special value, and then check if it doesn't (was overridden) and show the deprecation warning in that case

I think we don't have a good way of knowing whether someone does actually use `headers` - even if the headers are non-empty that doesn't have to mean they are using it for authorization (and even if no `Authorization` header is present that doesn't mean they are not using it for authorization as they could have their own auth scheme and use a different header)…

addon/mixins/data-adapter-mixin.js
sdebarros: The `assert` only throws the error if the condition is false, and I think we want it not to be thrown on newer versions 🤔

sdebarros:
> > It can also be annoying that we are showing the deprecation warning even if the user does use headers instead of authorize:
> > I thought of a few options but none is perfect:
> > 
> > * Do this.get('headers') and not show the deprecation if it's not empty (or if it has a value for Authorization), this would still show the deprecation if the header is not included because the session is not authenticated.
> > * Make authorize return some special value, and then check if it doesn't (was overridden) and show the deprecation warning in that case
> 
> I think we don't have a good way of knowing whether someone does actually use `headers` - even if the headers are non-empty that doesn't have to mean they are using it for authorization (and even if no `Authorization` header is present that doesn't mean they are not using it for authorization as they could have their own auth scheme and use a different header)…

What we would really need to know is if they've overridden `authorize`, because that's when we would need to warn them, but I can only think of the hacky second option above ^^ 

I don't think it's a crucial thing, but it may be confusing to see the warning if you're not actually overriding the method.

addon/mixins/data-adapter-mixin.js
marcoow: 🤦‍♂

marcoow:
> Make authorize return some special value, and then check if it doesn't (was overridden) and show the deprecation warning in that case

Hm, I think that could work indeed - I think you could use a [`Symbol`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol)

tests/unit/mixins/data-adapter-mixin-test.js
sdebarros: It wasn't failing, maybe `mocha` doesn't have that issue? I moved the `if` outside anyway since it doesn't make sense to have an empty test.

marcoow: Hm, yeah - maybe I mixed that up with QUnit's behavior.

PR #2013, opened 2019-11-11
Modernize tests

marcoow:
This modernizes the test suite by:

* using the latest ember-mocha (vs. not using ember-mocha at all)
* using ember-mocha's `setupXYZ` helpers
* removing the manually-mocked container (and replacing it with the owner set up by ember-mocha's `setupXYZ` helpers)
* using `async`/`await` in tests
* removing the `createWithContainer` helper


tests/unit/authenticators/devise-test.js
```diff
@@ -50,46 +49,47 @@ describe('DeviseAuthenticator', () => {
...
       });
 
       describe('when the server returns incomplete data', function() {
-        it('fails when token is missing', function() {
+        it('fails when token is missing', async function() {
           server.post('/users/sign_in', () => [201, { 'Content-Type': 'application/json' }, '{ "email": "email@address.com" }']);
 
-          return authenticator.authenticate('email@address.com', 'password').catch((error) => {
+          try {
+            await authenticator.authenticate('email@address.com', 'password');
```
sdebarros: Might be good to do something like https://github.com/simplabs/ember-simple-auth/pull/2013/files#diff-bb12f19912e2ea2cf2254d67032dbc05R27, if `authenticate` breaks and stops failing with an incorrect response, the test would still pass.

marcoow: yes, good point - fixed 👍

sdebarros: That pattern is in several places 😅there should be a nicer way to achieve this.

marcoow: I fixed the others as well - cannot think of a better solution for now 🤔

tests/unit/authenticators/oauth2-password-grant-test.js
```diff
@@ -47,22 +46,22 @@ describe('OAuth2PasswordGrantAuthenticator', () => {
...
+              expect(data).to.eql({ 'access_token': 'secret token 2!', 'expires_in': 67890, 'refresh_token': 'refresh token 2!' });
             });
           });
 
           describe('when the access token is not refreshed successfully', function() {
-            it('returns a rejecting promise', function(done) {
-              authenticator.restore({ 'access_token': 'secret token!', 'expires_at': 1 }).catch(() => {
+            it('returns a rejecting promise', async function() {
+              try {
+                await authenticator.restore({ 'access_token': 'secret token!', 'expires_at': 1 });
```
sdebarros: ☝️

sdebarros: This file has a few more 😄

marcoow: ah, right - fixed 👍

PR #2017, opened 2019-11-12
Deprecate legacy response handling

marcoow:
This deprecates:

* returning anything but a `Response` object from the (overridable) `OAuth2PasswordGrant` authenticator's `makeRequest` method
* the custom `responseJSON` and `responseText` properties on the response object that we reject with after a failing request

### DIFF
--- a/addon/authenticators/oauth2-password-grant.js
+++ b/addon/authenticators/oauth2-password-grant.js
@@ -397,13 +397,43 @@ export default BaseAuthenticator.extend({
           try {
             let json = JSON.parse(text);
             if (!response.ok) {
-              response.responseJSON = json;
+              Object.defineProperty(
+                response,
+                'responseJSON',
+                {
+                  get() {
+                    deprecate(`Ember Simple Auth: The response's responseJSON property is deprecated.`,
+                      false,
+                      {
+                        id: 'ember-simple-auth.oauth2-password-grant-authenticator.response-json',
+                        until: '3.0.0',
+                      }
+                    );
+                    return json;
+                  }
+                }
+              );
               reject(response);
             } else {
               resolve(json);
             }
           } catch (SyntaxError) {
-            response.responseText = text;
+            Object.defineProperty(
+              response,
+              'responseText',
+              {
+                get() {
+                  deprecate(`Ember Simple Auth: The response's responseText property is deprecated.`,
+                    false,
+                    {
+                      id: 'ember-simple-auth.oauth2-password-grant-authenticator.response-text',
+                      until: '3.0.0',
+                    }
+                  );
+                  return text;
+                }
+              }
+            );
             reject(response);
           }
         });
--- a/addon/authenticators/oauth2-password-grant.js
+++ b/addon/authenticators/oauth2-password-grant.js
@@ -199,7 +199,7 @@ export default BaseAuthenticator.extend({
           reject();
         }
       } else {
-        if (!this._validate(data)) {
+        if (!this._validateResponse(data)) {
           reject();
         } else {
           this._scheduleAccessTokenRefresh(data['expires_in'], data['expires_at'], data['refresh_token']);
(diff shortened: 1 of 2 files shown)


tests/unit/authenticators/oauth2-password-grant-test.js
```diff
@@ -463,4 +490,31 @@ describe('OAuth2PasswordGrantAuthenticator', () => {
...
+      beforeEach(function() {
+        authenticator = OAuth2PasswordGrant.extend({
+          makeRequest() {
+            return fetch('/token', { method: 'POST' }).then((response) => response.json());
+          }
+        }).create();
+        server.post('/token', () => [200, { 'Content-Type': 'application/json' }, '{ "access_token": "secret token 2!" }']);
+      });
+
+      it('shows a deprecation warning when not resolcing with a Response object', function() {
```
sdebarros:
```suggestion
      it('shows a deprecation warning when not resolving with a Response object', function() {
```

addon/authenticators/oauth2-password-grant.js
```diff
@@ -346,10 +345,16 @@ export default BaseAuthenticator.extend({
             }));
           }
         });
+
+        const rejectWithResponse = this.get('rejectWithResponse');
+        deprecateNotRejectingWithResponse(rejectWithResponse);
```
sdebarros: I'm not sure it makes sense to add this deprecation warning to places where it wasn't before, since we are not actually returning the response. Both here in `invalidate` and in `_refreshAccessToken` we are not returning the response (`_refreshAccessToken` is "private" even), there's nothing the user can do about it.

marcoow: I think you're confusing this with the new deprecation added in this PR. This is the deprecation we already had which is triggered when the `rejectWithResponse` property is `false`.

sdebarros: Yes, but we are not really using the response in these two places (or sending it to the user)

marcoow: Yes, but the setting should **always** be `true` - if we're only checking for it in the `authenticate` method and trigger a deprecation there but someone overrides that method for example, they would maybe never see the deprecation although they rely on the deprecated behavior.

sdebarros: Ok, I'm not entirely sure about showing this warning so much 😄  but I see your point.

addon/authenticators/oauth2-password-grant.js
```diff
@@ -279,35 +279,34 @@ export default BaseAuthenticator.extend({
...
+            if (!isEmpty(expiresAt)) {
+              json = assign(json, { 'expires_at': expiresAt });
+            }
+
+            resolve(json);
+          }).catch(reject);
+        }).catch(reject);
       }, (response) => {
-        run(null, reject, useResponse ? response : (response.responseJSON || response.responseText));
+        run(null, reject, rejectWithResponse ? response : (response.responseJSON || response.responseText));
```
sdebarros: I think this code would now be accessible if the `fetch` in https://github.com/simplabs/ember-simple-auth/pull/2017/files#diff-931dc51edd167948e773858f72a47233R400 fails, and in that case we don't have the `responseJSON` or `responseText`. This is now done by `_parseResponse`

addon/authenticators/oauth2-password-grant.js
```diff
@@ -434,15 +424,22 @@ export default BaseAuthenticator.extend({
...
+            refreshToken = json['refresh_token'] || refreshToken;
+            const expiresAt = this._absolutizeExpirationTime(expiresIn);
+            const data = assign(json, { 'expires_in': expiresIn, 'expires_at': expiresAt, 'refresh_token': refreshToken });
+            this._scheduleAccessTokenRefresh(expiresIn, null, refreshToken);
+            this.trigger('sessionDataUpdated', data);
+            resolve(data);
+          }).catch(reject);
+        }).catch(reject);
       }, (response) => {
         warn(`Access token could not be refreshed - server responded with ${response.responseJSON}.`, false, { id: 'ember-simple-auth.failedOAuth2TokenRefresh' });
```
sdebarros: The same as https://github.com/simplabs/ember-simple-auth/pull/2017/files#r349651768, I think it doesn't make sense to have it here, this should probably go in https://github.com/simplabs/ember-simple-auth/pull/2017/files#diff-931dc51edd167948e773858f72a47233R442
I didn't find a test for this warning.

addon/authenticators/oauth2-password-grant.js
marcoow: I fixed that. That also required fixing the tests and adding previously missing responses to the failing tests (e.g. `return [201, { 'Content-Type': 'application/json' }, '{ "access_token": "secret token 2!" }'];`). These missing responses were previously causing an error already but because of the `response.responseJSON || response.responseText` which resulted to `undefined` on that error object, the tests didn't break 🤯

addon/authenticators/oauth2-password-grant.js
marcoow: see above

addon/authenticators/oauth2-password-grant.js
```diff
@@ -434,17 +424,24 @@ export default BaseAuthenticator.extend({
...
+            refreshToken = json['refresh_token'] || refreshToken;
+            const expiresAt = this._absolutizeExpirationTime(expiresIn);
+            const data = assign(json, { 'expires_in': expiresIn, 'expires_at': expiresAt, 'refresh_token': refreshToken });
+            this._scheduleAccessTokenRefresh(expiresIn, null, refreshToken);
+            this.trigger('sessionDataUpdated', data);
+            resolve(data);
+          }).catch(reject);
+        }).catch(reject);
+      }, (body) => {
+        warn(`Access token could not be refreshed - server responded with ${body}.`, false, { id: 'ember-simple-auth.failedOAuth2TokenRefresh' });
```
sdebarros: I think the warning is only shown now when the fetch fails vs before that it was shown when the request failed, is that OK? (I don't know how important this warning is)

PR #2034, opened 2019-11-22
switch to github actions

marcoow:
…as it will hopefully speed things up significantly as simplabs/ember-test-selectors#443 did.


.github/workflows/ci.yml
```diff
@@ -0,0 +1,77 @@
+name: CI
+
+on:
+  push:
+    branches:
+      - master
+      - 'v*'
+      - 2.x
+      - /^\d+\.\d+\.\d+/
```
Turbo87: maybe we should standardize on one way to name our version branches 😅

.github/workflows/ci.yml
```diff
@@ -0,0 +1,77 @@
...
+      - name: lint
+        run: yarn lint
+
+  tests:
+    name: Tests
+    runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
+        test-suite: ["test:one ember-3.0", "test:one ember-lts-3.4", "test:one ember-lts-3.8", "test:one ember-lts-3.12", "test:one ember-release", "test:one ember-beta", "test:node", "test:fastboot", "test:one ember-canary"]
```
Turbo87: is there a way to format this differently to make it more readable?

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,38 @@
...
+
+      - name: install dependencies
+        run: yarn install
+
+      - name: Deploy
+        env:
+          NOW_AUTH_TOKEN: ${{ secrets.NOW_TOKEN }}
+        run: |
+          now deploy --token=$NOW_AUTH_TOKEN --scope=e-simple-auth && now alias --token=$NOW_AUTH_TOKEN --scope=e-simple-auth
+          cd server && now deploy --token=$NOW_AUTH_TOKEN --scope=e-simple-auth && now alias --token=$NOW_AUTH_TOKEN --scope=e-simple-auth
```
Turbo87: do we need both?

.github/workflows/release.yml
```diff
@@ -0,0 +1,21 @@
...
+
+    steps:
+      - uses: actions/checkout@v1
+      - uses: actions/setup-node@v1
+        with:
+          node-version: 12.x
+
+      - run: npm publish
+        env:
+          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
```
Turbo87: I assume you've set it in the github repo secrets section?

.travis.yml
```diff
@@ -1,84 +0,0 @@
...
-    - stage: 'Basic Tests'
-      name: 'Lint'
-      env: SCRIPT=lint
-
-    - name: 'Fixed Dependencies'
-      env: SCRIPT="test"
-
-    - name: 'Floating Dependencies'
-      env: SCRIPT="test"
-      install: yarn install --no-lockfile
```
Turbo87: we don't seem to test this case anymore?

.github/workflows/ci.yml
marcoow: `/^\d+\.\d+\.\d+/` is for the release tags though - for ember-test-selectors we use `'v*'` but we're not using the prefix here yet. I copied it anyway as we might want to streamline the tags across our repositories. `2.x` is mainly relevant now as `master` contains all the changes for `3.0` already.

.github/workflows/ci.yml
marcoow: yes, done

.github/workflows/deploy.yml
marcoow: Yes, one deploys the Ember app, the other one deploys the mock server.

.github/workflows/release.yml
marcoow: yes 👍

.travis.yml
marcoow: added back in

.github/workflows/ci.yml
Turbo87: then it should be under `tags`, not `branches` 🤔

.github/workflows/deploy.yml
Turbo87: it seems to be the exact same call. will they be deployed to different locations? 🤔

marcoow: Yes, one has `"name": "simple-auth-server",` in the `package.json`, the other is `"name": "ember-simple-auth",` - see the `cd server`

Turbo87: I see 👍

.github/workflows/ci.yml
marcoow: You're right - we don't need this at all actually. I also renamed the `2.x` branch to `v2`.

marcoow:
🎉  Once we fix the FastBoot tests so we don't run into a 10 minute timeout anymore, this will make the build dramatically faster!

PR #2035, opened 2019-11-23
Added authorization_code flow authenticators for torii provider.

davideferre:
Extended torii provider for refreshing tokens in oauth2 authorization_code flow.

### DIFF
--- a/addon/authenticators/torii-authorization-code.js
+++ b/addon/authenticators/torii-authorization-code.js
@@ -1,0 +1,173 @@
+import ToriiAuthenticator from 'ember-simple-auth/authenticators/torii';
+import { warn } from '@ember/debug';
+import { keys as emberKeys } from '@ember/polyfills';
+import { run } from '@ember/runloop';
+import { inject as service } from '@ember/service';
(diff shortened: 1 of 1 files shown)


davideferre:
> Thanks for the contribution!
> 
> However, this looks like it is handling a very specific case that is likely only relevant to your application but not everyone's Torii-based authenticators? This should likely remain in your application code rather than in the base library itself. Please let me know if I'm missing something.

I've opened a PR because this is not only relevant to my application but it's a common problem for everyone that use an oauth2 autentication system with authorization_code flow (PKCE).
This flow is useful if you develop app or webapp and you don't want to let users share theirs passwords within the mobile app.
Useful links:
- https://hackernoon.com/strengthening-oauth2-for-mobile-f4f3925dbf18
- https://auth0.com/docs/flows/concepts/auth-code-pkce

marcoow:
Sure, refresh tokens are a good thing in general but I don't know what Torii providers people that use the Torii authenticator use and whether those providers/services use refresh tokens at all?

marcoow:
Closing this as it's out of scope for ESA.

PR #2076, opened 2019-12-30
Fix deprecations regarding Computed Property Overrides

### LINKED ISSUE #1908: Deprecation warnings on Ember 3.9 (computed property was just overriden)

marcoow:
This fixes the deprecations regarding overridden computed properties. Most of these CPs were only used in testing and should be properly mocked via the container anyway (e.g. `_fastBoot`).

### TODO

- [x] all deprecations in app code should be fixed but there are some left in the testing code (mainly the code around testing adaptive and cookie session stores which is relatively messy anyway 😞) - that should be fixed before releasing this as well.

closes #1908
closes #2044 
closes #1938

### DIFF
--- a/addon/internal-session.js
+++ b/addon/internal-session.js
@@ -194,7 +194,7 @@ export default ObjectProxy.extend(Evented, {
 
   _lookupAuthenticator(authenticatorName) {
     let owner = getOwner(this);
-    let authenticator = getOwner(this).lookup(authenticatorName);
+    let authenticator = owner.lookup(authenticatorName);
     setOwner(authenticator, owner);
     return authenticator;
   }
--- a/addon/internal-session.js
+++ b/addon/internal-session.js
@@ -5,7 +5,7 @@ import Evented from '@ember/object/evented';
 import { merge, assign as emberAssign } from '@ember/polyfills';
 import { set } from '@ember/object';
 import { debug, assert } from '@ember/debug';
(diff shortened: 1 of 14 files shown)


addon/internal-session.js
```diff
@@ -192,7 +192,10 @@ export default ObjectProxy.extend(Evented, {
     });
   },
 
-  _lookupAuthenticator(authenticator) {
-    return getOwner(this).lookup(authenticator);
+  _lookupAuthenticator(authenticatorName) {
+    let owner = getOwner(this);
+    let authenticator = getOwner(this).lookup(authenticatorName);
```
sdebarros:
```suggestion
    let authenticator = owner.lookup(authenticatorName);
```

pjcarly:
Great, thanks for the merge.

Could we get a new NPM release?

PR #2135, opened 2020-03-18
Modernize dummy app

marcoow:
This modernizes the dummy app by:

* using `async`/`await` instead of dealing with promises
* using closure actions instead of old actions or `sendAction`
* using ember-data's package imports
* using native inputs in the login form instead of the `input` helper

It also fixes handling of invalid credentials and other login errors in the `<LoginForm>` which is currently broken.

### DIFF
--- a/tests/dummy/app/components/login-form.js
+++ b/tests/dummy/app/components/login-form.js
@@ -6,15 +6,17 @@ export default Component.extend({
   session: service('session'),
 
   actions: {
-    authenticateWithOAuth2() {
-      let { identification, password } = this.getProperties('identification', 'password');
-      this.get('session').authenticate('authenticator:oauth2', identification, password)
-        .then(() => {
-          this.get('rememberMe') && this.set('session.store.cookieExpirationTime', 60 * 60 * 24 * 14);
-        })
-        .catch((reason) => {
-          this.set('errorMessage', reason.error);
-        });
+    async authenticateWithOAuth2() {
+      try {
+        let { identification, password } = this;
+        await this.get('session').authenticate('authenticator:oauth2', identification, password);
+
+        if (this.rememberMe) {
+          this.get('session').set('store.cookieExpirationTime', 60 * 60 * 24 * 14);
+        }
+      } catch (reason) {
+        this.set('errorMessage', reason.error);
+      }
     },
 
     authenticateWithFacebook() {
--- a/tests/dummy/app/components/login-form.js
+++ b/tests/dummy/app/components/login-form.js
@@ -32,6 +32,18 @@ export default Component.extend({
                             + `&response_type=${responseType}`
                             + `&scope=${scope}`
       );
-    }
+    },
+
+    updateIdentification(e) {
+      this.set('identification', e.target.value);
+    },
+
+    updatePassword(e) {
+      this.set('password', e.target.value);
+    },
+
+    updateRememberMe(e) {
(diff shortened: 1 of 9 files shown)


tests/dummy/app/templates/components/login-form.hbs
```diff
@@ -7,14 +7,14 @@
 <form {{action 'authenticateWithOAuth2' on='submit'}}>
   <div class="form-group">
     <label for="identification">Login</label>
-    {{input value=identification placeholder='Enter Login' class='form-control'}}
+    <input type="text" placeholder="Enter Login" class="form-control" onchange={{action 'updateIdentification'}} value={{this.identification}} />
```
Turbo87: hmm, what's the reason for this change? why not keep using the `<Input>` component?

tests/dummy/app/components/login-form.js
```diff
@@ -6,15 +6,18 @@ export default Component.extend({
...
+      try {
+        let { identification, password } = this;
+        await this.get('session').authenticate('authenticator:oauth2', identification, password);
+
+        if (this.rememberMe) {
+          this.get('session').set('store.cookieExpirationTime', 60 * 60 * 24 * 14);
+        }
+      } catch (response) {
+        let responseBody = await response.clone().json();
+        this.errorMessage = responseBody.error;
```
Turbo87: this looks quite different than the previous implementation 🤔

tests/dummy/app/services/session-account.js
```diff
@@ -1,22 +1,14 @@
...
-        }, reject);
-      } else {
-        resolve();
-      }
-    });
+  async loadCurrentUser() {
+    let accountId = this.get('session.data.authenticated.account_id');
+    if (accountId) {
+      let account = await this.get('store').find('account', accountId);
+      this.account = account;
```
Turbo87: should we not keep using `this.set()` here?

marcoow: yes, fixed

tests/dummy/app/templates/components/login-form.hbs
marcoow: I find using native inputs easier to follow and more explicit. What would be an advantage of the `<Input>` component?

tests/dummy/app/components/login-form.js
marcoow: Yes, this is actually broken on `master` currently…

Turbo87: then shouldn't we fix this in a dedicated PR?

I find this PR somewhat hard to review because it's changing so many things at the same time

marcoow: Yes, makes sense: #2138

tests/dummy/app/templates/components/login-form.hbs
```diff
@@ -4,17 +4,17 @@
 <a {{action 'authenticateWithGoogleImplicitGrant'}} class="btn btn-primary" href="#" role="button">Login with Google (Implicit Grant)</a>
 <hr/>
 <p>or login with OAuth 2.0 <em>Resource Owner Password Credentials</em>:</p>
-<form {{action 'authenticateWithOAuth2' on='submit'}}>
+<form onsubmit={{action 'authenticateWithOAuth2'}}>
```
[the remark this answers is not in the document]
marcoow: ok, can revert

PR #2138, opened 2020-03-18
Fix handling of login errors in the dummy app

marcoow:
This fixes handling of login errors in the dummy app. Since version 3.0 we no longer reject with the error response's (JSON) body but with the response itself – the dummy app had not been updated accordingly yet.

### DIFF
--- a/tests/dummy/app/components/login-form.js
+++ b/tests/dummy/app/components/login-form.js
@@ -14,8 +14,9 @@ export default Component.extend({
         if (this.rememberMe) {
           this.get('session').set('store.cookieExpirationTime', 60 * 60 * 24 * 14);
         }
-      } catch (reason) {
-        this.set('errorMessage', reason.error);
+      } catch (response) {
+        let responseBody = await response.clone().json();
+        this.set('errorMessage', responseBody);
       }
     },
 
--- a/tests/dummy/app/templates/components/login-form.hbs
+++ b/tests/dummy/app/templates/components/login-form.hbs
@@ -7,11 +7,11 @@
 <form {{action 'authenticateWithOAuth2' on='submit'}}>
   <div class="form-group">
     <label for="identification">Login</label>
-    <input type="text" placeholder="Enter Login" class="form-control" onchange={{action 'updateIdentification'}} value={{this.identification}} />
+    <input type="text" placeholder="Enter Login" class="form-control" onchange={{action 'updateIdentification'}} value={{this.identification}} data-test-identification />
   </div>
   <div class="form-group">
     <label for="password">Password</label>
-    <input type="password" placeholder="Enter Password" class="form-control" onchange={{action 'updatePassword'}} value={{this.password}} />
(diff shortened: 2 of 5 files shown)


Turbo87:
> Since version 3.0 we no longer reject with the error response's (JSON) body but with the response itself

I'm confused by this. Do we not reject with an `Error` instance?

see https://eslint.org/docs/rules/prefer-promise-reject-errors

marcoow:
No, unfortunately we do not reject with an `Error` anywhere 😞.

tests/dummy/app/components/login-form.js
```diff
@@ -12,8 +12,8 @@ export default Component.extend({
         .then(() => {
           this.get('rememberMe') && this.set('session.store.cookieExpirationTime', 60 * 60 * 24 * 14);
         })
-        .catch((reason) => {
-          this.set('errorMessage', reason.error);
+        .catch((response) => {
+          response.clone().json().then((responseBody) => this.set('errorMessage', responseBody.error));
```
Turbo87: why do we have to `.clone()` it?

tests/dummy/app/templates/components/login-form.hbs
```diff
@@ -7,11 +7,11 @@
 <form {{action 'authenticateWithOAuth2' on='submit'}}>
   <div class="form-group">
     <label for="identification">Login</label>
-    {{input value=identification placeholder='Enter Login' class='form-control'}}
+    {{input value=identification placeholder='Enter Login' class='form-control' data-test-identification="true"}}
```
Turbo87:
```suggestion
    {{input value=identification placeholder='Enter Login' class='form-control' data-test-identification=true}}
```

tests/dummy/app/templates/components/login-form.hbs
```diff
@@ -7,11 +7,11 @@
...
   </div>
   <div class="form-group">
     <label for="password">Password</label>
-    {{input value=password placeholder='Enter Password' class='form-control' type='password'}}
+    {{input value=password placeholder='Enter Password' class='form-control' type='password' data-test-password="true"}}
```
Turbo87:
```suggestion
    {{input value=password placeholder='Enter Password' class='form-control' type='password' data-test-password=true}}
```

tests/dummy/app/components/login-form.js
marcoow: We are [reading the response body in the authenticator](https://github.com/simplabs/ember-simple-auth/blob/master/addon/authenticators/oauth2-password-grant.js#L314) already and it can only be read once. We could of course clone in the authenticator before rejecting so you wouldn't have to do that when catching it but that might be wasteful in case you only need the response headers.

Turbo87: is this documented somewhere?

marcoow: added a note here: #2141

PR #2166, opened 2020-04-06
Zeit 2

### LINKED ISSUE #2165: Upgrade demo app to Zeit Now 2
Zeit Now 1.0 will be shut down in summer and we need to [upgrade](https://zeit.co/guides/migrate-to-zeit-now/) the demo servers we are hosting there. That includes the [HTTP mock server](https://github.com/simplabs/ember-simple-auth/blob/master/server/app.js) as well as the [FastBoot server](https://github.com/simplabs/ember-simple-auth/blob/master/fastboot-server.js).

marcoow:
closes #2165


marcoow:
I think we'll switch to heroku's free tier – should be much simpler (and cheaper because free)

PR #2167, opened 2020-04-06
remove config for removed checks

marcoow:
This removes ESLint configuration that disables the `ember/local-modules` and `ember/avoid-leaking-state-in-components` which have been removed in `eslint-plugin-ember@7.0` already.


marcoow:
🤦‍♂

PR #2184, opened 2020-04-15
Convert to Yarn workspace

marcoow:
As we're starting to make substantial modifications to the codebase that will also significantly change the way that ESA is used in applications (and likely drop support for some older versions of Ember again), we will require significantly different applications to test ESA in (e.g. one using the classic mixins integration, one using ESA via decorators (or another new mechanism), one app using typescript maybe etc.) This PR creates the foundation for that so that we can have as many Ember apps as we want as workspaces that all use the same addon.

### TODO

- [ ] the Procfiles are now both at the top level and called `Procfile.client` and `Procfile.server` – we'll need to update the heroku settings accordingly when merging this


packages/ember-simple-auth/node_modules/supports-color/package.json
```diff
@@ -0,0 +1,53 @@
+{
```
Turbo87, line 1: why do we add `node_modules` folders to this repo?

marcoow: Ugh, added that accidentally.

This is still an early draft though – I'll let everyone know when this is ready for review.

marcoow:
It's likely I missed some things and unfortunately this PR is huge but it'd be good to get a second (and third, fourth, …) pair of eyes on this.

packages/ember-simple-auth/node_modules/.bin/_mocha
```diff
@@ -0,0 +1 @@
+../mocha/bin/_mocha
```
pichfl, line 1: Are these checked in intentionally?

pichfl:
@marcoow I think some `packages/ember-simple-auth/node_modules/*` were added accidentally

packages/ember-simple-auth/node_modules/.bin/_mocha
marcoow: Ugh yep, added those accidentally – removed now

.ember-cli
```diff
@@ -1,9 +0,0 @@
-{
-  /**
-    Ember CLI sends analytics information by default. The data is completely
-    anonymous, but there are times when you might want to disable this behavior.
-
-    Setting `disableAnalytics` to true will prevent any data from being sent.
-  */
-  "disableAnalytics": false
-}
```
Turbo87, line 9: I'm surprised that this doesn't show up as "moved". Did you explicitly delete it?

.eslintignore
```diff
@@ -1,9 +1,3 @@
-/blueprints/*/files
-
-# unconventional js
-/blueprints/*/files/
-/vendor/
-
 # compiled output
 /dist/
 /tmp/
```
Turbo87, line 3: why do we keep these? do we expect to have these folders in the top-level folder?

.github/workflows/ci.yml
```diff
@@ -35,22 +41,24 @@ jobs:
...
 
   tests:
     name: Tests
     runs-on: ubuntu-latest
 
     strategy:
       matrix:
+        workspace:
+          - ember-simple-auth
+          - test-app
```
Turbo87, line 54: do we want to run these workspace tests for all Ember versions? I'm a bit worried the CI time will explode by doing that.

.github/workflows/ci.yml
```diff
@@ -35,22 +41,24 @@ jobs:
         run: yarn install
 
       - name: lint
-        run: yarn lint
+        run: yarn workspace ${{ matrix.workspace }} lint
```
Turbo87, line 44: I'm wondering if we should run the lint for all workspaces in just one job instead

packages/ember-simple-auth/.eslintignore
```diff
@@ -0,0 +1,18 @@
...
+/tmp/
+
+# dependencies
+/bower_components/
+
+# misc
+/coverage/
+
+# ember-try
+/.node_modules.ember-try/
```
Turbo87, line 18: what about `.bower_components.ember-try`? otherwise we probably don't need the `bower_components` above either, right?

packages/ember-simple-auth/ember-cli-build.js
```diff
@@ -9,11 +9,6 @@ var sourceTrees = [];
 module.exports = function(defaults) {
   var app = new EmberAddon(defaults, {
     storeConfigInMeta: true,
-    'ember-bootstrap': {
-      bootstrapVersion: 4,
-      importBootstrapFont: false,
-      importBootstrapCSS: true
-    },
```
Turbo87, line 16: why was this removed?

packages/ember-simple-auth/tests/unit/mixins/application-route-mixin-test.js
```diff
@@ -29,6 +31,7 @@ describe('ApplicationRouteMixin', () => {
       clear: sinon.stub()
     }));
 
+    this.owner.register('route:application', Route.extend(ApplicationRouteMixin));
```
Turbo87, line 34: why is this change necessary now?

.ember-cli
marcoow: fixed

.eslintignore
marcoow: fixed

.github/workflows/ci.yml
marcoow: I think we need to do this, otherwise we lose a bit of coverage as the tests in `packages/ember-simple-auth` are unit tests only and the ones in `packages/test-app` are application tests.

.github/workflows/ci.yml
marcoow: But this way we can run the jobs in parallel?

packages/ember-simple-auth/ember-cli-build.js
marcoow: It is no longer needed in `packages/ember-simple-auth` since the dummy app has been moved to `packages/test-app` and the dummy app in `packages/ember-simple-auth` is now empty.

packages/ember-simple-auth/.eslintignore
marcoow: I removed `bower_components` above 👍

packages/ember-simple-auth/tests/unit/mixins/application-route-mixin-test.js
marcoow: This was previously relying on the application route that's defined in the dummy app. The dummy app of `packages/ember-simple-auth` is now empty though as it has been moved to `packages/test-app`, thus there is no application controller anymore if we don't register one.

.github/workflows/ci.yml
Turbo87: yeah, but it has to perform the `yarn install` for every single job which most likely takes longer than the actual linting

marcoow: The linting jobs still finish in around 1m though: https://github.com/simplabs/ember-simple-auth/pull/2184/checks?check_run_id=608604147 which is substantially faster than the jobs than run tests. Assuming all jobs run in parallel I think keeping the CI config simpler trumps speeding up these jobs that are not a limiting factor for the CI pipeline anyway?

package.json
```diff
@@ -1,124 +1,15 @@
...
-    "configPath": "tests/dummy/config",
-    "versionCompatibility": {
-      "ember": ">=1.12"
-    }
-  },
-  "fastbootDependencies": [
-    "node-fetch"
+  "workspaces": [
+    "packages/*"
   ],
```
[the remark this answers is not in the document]
marcoow: What would you say would the advantage of that be?

package.json
```diff
@@ -1,124 +1,15 @@
...
-    "loader.js": "^4.5.1",
-    "marked": "^1.0.0",
-    "mermaid": "^7.0.3",
-    "mocha": "^7.0.1",
-    "morgan": "^1.10.0",
-    "my-engine": "./tests/dummy/lib/my-engine",
-    "rimraf": "^3.0.0",
-    "sinon-chai": "~3.5.0",
-    "torii": "^0.10.0"
+    "diagrams": "mmdc -i guides/assets/esa-initial-flow.txt -o guides/assets/esa-initial-flow.svg"
```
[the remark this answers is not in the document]
marcoow: Not sure what you mean – the chart is linked from the [guides](https://github.com/simplabs/ember-simple-auth/tree/master/guides)

.github/workflows/ci.yml
```diff
@@ -105,5 +119,32 @@ jobs:
...
+
+      - uses: actions/cache@v1
+        with:
+          path: ${{ steps.yarn-cache.outputs.dir }}
+          key: ${{ runner.os }}-yarn-test-${{ hashFiles('**/yarn.lock') }}
+          restore-keys: |
+            ${{ runner.os }}-yarn-
+
+      - name: install dependencies
+        run: yarn install
```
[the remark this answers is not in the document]
marcoow: Yes, probably a good idea but unrelated as you say 👍

package.json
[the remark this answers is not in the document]
patsy-issa: Ah missed the link in there 👍

The previous script generates a `png`  that the guides link to, the following change turns the output to `svg`, I would prefer if we keep the original `png` output for this pr and replace the png resources with svgs on its own.

package.json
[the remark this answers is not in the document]
patsy-issa: I think it makes it easier to see what workspaces exists in the project from `package.json` without needing to look under `/packages`

package.json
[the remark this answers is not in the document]
marcoow: This newer version of mermaid doesn't support PNG anymore I think.

package.json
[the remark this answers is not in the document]
marcoow: Hm, we can list the packages of course – I'm not sure that's a real problem anyone would have though? 🤔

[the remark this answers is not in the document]
patsy-issa: Not a problem per say more of a explicit vs automagic preference, no strong feelings either way ^^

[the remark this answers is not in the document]
marcoow: I don't really see added value tbh – let's keep it as is if you don't have a strong preference

package.json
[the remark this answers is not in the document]
marcoow: fixed in the guides

guides/auth-torii-with-github.md
```diff
@@ -31,7 +31,7 @@ configuration and base classes. For GitHub specifically, the authentication flow
    <access_token>` header request to, say, `https://api.github.com/user`,
    which returns the GitHub user’s user info, at which point it can be matched to the app's own `User` model. 
 
-You can find a sequence diagram for the full flow [here](./assets/esa-initial-flow.txt.png).
+You can find a sequence diagram for the full flow [here](./assets/esa-initial-flow.svg).
```
patsy-issa, line 34: Won't this be a broken link as the [file](https://github.com/simplabs/ember-simple-auth/tree/23981b7e39d137bd63440bdc114bcf7daa832ef8/guides/assets) was not generated manually?

marcoow: 🤦 fixed

PR #2198, opened 2020-05-08
Deprecate mixins

### LINKED ISSUE #2185: Quest: Drop mixins
ember-simple-auth has been heavily leveraging mixins form the beginning. Using mixins is what allowed ESA to be integrated easily be developers into their applications since the only thing that they'd need to do was to mix-in these mixins into some of the routes of the application and ESA would just work. However, with native JavaScript classes, mixins aren't really well supported and you have to fall back to constructs like

```js
export default class ProtectedRoute extends Route.extend(AuthenticatedRouteMixin) {
}
```

which mixes native classes and Ember's old object model and thus results in unintuitive and harder-to-understand-than-necessary code. Also, eventually mixins will likely be removed from Ember altogether and they already trigger [linter warnings](https://github.com/ember-cli/eslint-plugin-ember/blob/master/docs/rules/no-mixins.md) which means installing ESA into an app will trigger new linter warnings which isn't a great and particularly reassuring experience obviously.

For those reasons, ESA should drop its mixins and adopt a new approach. There are a few key attributes we're looking for in a new solution:

* Ease of use: as explained above, mixins made ESA super easy to use for everyone and we don't want to give that up when switching to a new mechanism. Ideally, integrating ESA into an app would still only require a few steps with little complexity to understand.
* Customizability: another advantage of mixins is that everything they add to the respective classes can be overridden again and thus customized. That gives developers an easy way to modify particular aspects of ESA's behavior so it perfectly fits their needs without having to go through lots of complexity. In fact, we even extracted and designed some of the methods on the ESA mixins specifically with extensibility and customizability in mind, e.g. https://github.com/simplabs/ember-simple-auth/blob/master/addon/mixins/authenticated-route-mixin.js#L118
* Sustainability: obviously we don't want to change the main mechanism for integrating ESA into an app frequently so we're looking for something that will be stable for the foreseeable future. That means obviously we cannot use anything that is already deprecated in Ember 3.x but should also be watching for what might be coming in Ember and/or JS in the future and be sure the mixin replacement we choose now will continue to work.

### Status Quo

ESA currently provides 5 mixins:

* [`ApplicationRouteMixin`](https://ember-simple-auth.com/api/classes/ApplicationRouteMixin.html): This is an optional mixin that when used, maps certain events of the session to methods on the application route (by overriding `init`) that it provides default implementations of.
* [`AuthenticatedRouteMixin`](https://ember-simple-auth.com/api/classes/AuthenticatedRouteMixin.html): This is mixed in to route classes that are to expect the session to be authenticated in order to be accessed. The mixin adds a custom `beforeModel` method that will call the original one (if present) if the session is authenticated or call `triggerAuthentication` otherwise that it provides a default implementation of which redirects to a configurable login route
* [`UnauthenticatedRouteMixin`](https://ember-simple-auth.com/api/classes/UnauthenticatedRouteMixin.html): This is the opposite of the `AuthenticatedRouteMixin` and works in the same way by adding a default implementation of `beforeModel` that checks whether the session is **not** authenticated and redirects otherwise (without a dedicated method for that)
* [`OAuth2ImplicitGrantCallbackRouteMixin`](https://ember-simple-auth.com/api/classes/OAuth2ImplicitGrantCallbackRouteMixin.html) is used in combination with the [`OAuth2ImplicitGrantAuthenticator`](https://ember-simple-auth.com/api/classes/OAuth2ImplicitGrantAuthenticator.html) and triggers authentication of the session via that authenticator when the route is visited (by overriding `activate`)
* [`DataAdapterMixin`](https://ember-simple-auth.com/api/classes/DataAdapterMixin.html) is used with Ember Data adapters and overrides the `handleResponse` method so that it automatically invalidates the session on 401 error responses

### Options

There are several options that might be worth looking into more deeply:

* Decorators: This seems to be the obvious choice for replacing mixins. Instead of the above example, one would simply write

```js
import { authenticated } from 'ember-simple-auth/decorators/routes';

@authenticated
export default class ProtectedRoute extends Route {
}
```

The advantage of decorators is they would still allow integrating ESA super easily but the disadvantage is the uncertainty around the decorators proposal and the fact that we might not be able to do everything we need to do with the final version of the proposal
* Custom mixin mechanism: Using a mechanism for mixins that does not rely on the Ember Object model but works with native JavaScript classes would be another option that minimises the changes in terms of how ESA is integrated into an application. The above example could look something like this with an approach like that:

```js
import { Protected } from 'ember-simple-auth/mixins/routes';

export default class ProtectedRoute extends Protected(Route) {
}
```

More information on the approach can be found [here](https://justinfagnani.com/2015/12/21/real-mixins-with-javascript-classes/). The disadvantage of this is that it is fairly exotic and while technically ESA could still be adopted in a way that is very similar to the current mixins-based mechanism, it would result in substantially increased complexity as it'd be unclear to many devs what they are doing exactly and how the mechanism works.
* Providing individual methods: Instead of providing ESA's functionality via mixins it would be possible of course to export the individual methods instead so that those would be called manually be the dev in the appropriate places instead of automatically via the mixins. The above example would look something like this with the approach:

```js
import { requireAuthentication } from 'ember-simple-auth/helpers/routes';

export default class ProtectedRoute extends Route {
  async beforeModel() {
    await requireAuthentication();
    …
  }
}
```

While this has the advantage of being fairly simple and straight-forward, the disadvantage is that it requires the dev to follow more steps when adapting ESA in an application which leads to more complexity simply as the amount of glue code that integrates ESA and the application is higer.

### Prior Art

@mike-north did some work some time ago to split the ESA codebase apart into [exports of individual methods that could be imported and called directly as a replacement for mixins](https://github.com/simplabs/ember-simple-auth/issues/1619). Unfortunately nobody ever found the time to implement it and the effort didn't lead anywhere…

marcoow:
This PR deprecates the mixins and introduces a new API that is exposed completely via the `session` service.

We discussed a few options for replacing the current mixins with a new mechanism in #2185 (thanks to everyone who chimed in!) but focussed mostly on concepts that would replace the mixins directly. Since we have dropped support for older versions of Ember though and now have new capabilities that we can leverage that were not available when ESA's current mixin mechanism was first introduced, I realized we could build a completely new API.

Historically, transitioning between routes was only really possible from route classes (or controllers) since there was no routing service. For that reason, we had to establish a connection between the session and its change state on one side and the routes on the other side. With the routing service that is no longer necessary though as we can transition between routes from the session directly using the service. That means we can just drop the `ApplicationRouteMixin` that implemented the above described connection between the routing layer and the session and handle all of that in the `session` itself. We still maintain customizability of course since any app can simply extend the session service and override the respective methods, similar to how you would do the same thing in the application route currently.

The `AuthenticatedRouteMixin`'s and `UnauthenticatedRouteMixin`'s main responsibilities were adding custom logic into the route classes' `beforeModel` methods that would check for the session to be authenticated or not and transition elsewhere etc. A mixin is a nice way to share that code and automatically put it in the right place. While exposing that API via the session service now will require a bit more code, it still doesn't put significant additional cognitive load on the developer and it's easy to extract the code into a base class if one wanted to do that:

```js
export default ProtectedRoute extends Route {
  @service session;

  beforeModel(transition) {
    this.session.requireAuthentication(transition, 'login');
  }
}
```

The opposite of this would be `this.session.prohibitAuthentication('index')` for disallowing a route to be visited when the session is authenticated already.

The `OAuth2ImplicitGrantCallbackRouteMixin` and `DataAdapterMixin` mixins are deprecated without 1:1 replacements. They provide very little value anyway and would usually only be used once per application. Thus, adding guidance in the docs for achieving the respective behavior should be sufficient.

### TODO

- [x] update README and guides
- [x] deprecate `DataAdapterMixin`
- [x] deprecate `OAuth2ImplicitGrantCallbackRouteMixin`
- [x] update the API docs template to show the deprecation status of mixins

closes #2185


packages/ember-simple-auth/addon/services/session.js
```diff
@@ -139,12 +161,61 @@ export default Service.extend(Evented, {
...
+    lifecycle, specifically in an instance initializer or the application
+    route's constructor.__
+
+    @method setup
+    @public
+  */
+  setup() {
+    this.get('session').on('authenticationSucceeded', () => this.handleAuthentication(Configuration.routeAfterAuthentication));
+    this.get('session').on('invalidationSucceeded', () => this.handleInvalidation(Configuration.rootURL));
+  },
```
jelhan, line 217: Why does the internal session communicates with the public session through events? Wondering if internal session can call `handleAuthentication` and `handeInvalidation` methods of public session service directly. This would also remove the need for this `setup()` method, wouldn't it?

packages/ember-simple-auth/addon/services/session.js
```diff
@@ -212,4 +283,99 @@ export default Service.extend(Evented, {
...
+    {{#crossLink "SessionService/attemptedTransition:property"}}{{/crossLink}}
+    property so that  it can be retried after the session is authenticated. If
+    the transition is aborted in Fastboot mode, the transition's target URL
+    will be saved in a `ember_simple_auth-redirectTarget` cookie for use by the
+    browser after authentication is complete.
+
+    @method requireAuthentication
+    @param {Transition} transition A transition that triggered the authentication requirement or null if the requirement originated independently of a transition
+    @param {String|Function} routeOrCallback The route to transition to in case that the session is not authenticated or a callback function to invoke in that case
+    @return {Boolean} true when the session is authenticated, false otherwise
```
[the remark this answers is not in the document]
marcoow: So this method and `prohibitAuthentication` are more or less directly replacements for the `AuthenticatedRouteMixin` and `UnauthenticatedRouteMixin`. The logic in these mixins is completely sync as well since all they do is check whether the session is authenticated or not and if that's the case or not, redirect appropriately. I think what you're suggesting is turning the session's `isAuthenticated` property into something that's async but I'd very much like to avoid that. If you cannot synchronously know whether the session is authenticated or not that's a huge change to ESA's API and probably introduces significant complexity.

If you want to check whether the authentication data you have stored in the session is still valid before every visit to a protected route, you could of course not use the `requireAuthentication` method but roll your own that makes a request that validates the token every time. I think the approach that we've been taking in ESA for years though (if we have auth data in the session, assume it's valid until we learn it is not, e.g. by receiving a 401 response) has proven to work pretty well.

packages/ember-simple-auth/addon/services/session.js
marcoow: I think the events between the internal session and the session service are an internal detail that I'd like not to change as part of this. There might be better ways to do this eventually of course. You are right though that we might not need the `setup` method since we're already instantiating the `session` service when the application starts up: https://github.com/simplabs/ember-simple-auth/blob/master/packages/ember-simple-auth/addon/initializers/setup-session-restoration.js#L16

marcoow: I removed the setup method.

packages/ember-simple-auth/addon/services/session.js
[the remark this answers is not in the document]
jelhan: > I think what you're suggesting is turning the session's `isAuthenticated` property into something that's async

You got me wrong here. `isAuthenticated` should be still a sync property. It should still reflect if valid authentication information are available. I only want to add official support for transparently establishing a local session if a SSO session exists.

Ember Simple Auth currently assumes that there are only two states:

1. The app has valid authentication credentials locally, which means that the user is authenticated.
2. The app does not have valid authentication credentials locally, which means the user is not authenticated.

A single sign-on service decouples the two states: The app may not have valid authentication credentials (e.g. an access token) locally but the user may have a SSO session. In that scenario the local session should be established transparently to the user.

The `requireAuthentication` method would be a good place to do this. An implementation might look like this:

```js
async requireAuthentication() {
  let { isAuthenticated } = this;

  // check if local session exists
  if (!isAuthenticated) {
    try {
      // try to establish local session transparently to the user
      await this.authenticate();
    } catch (error) {
      // wasn't able to establish a local session without user interaction
      // use default implementation to redirect user to SSO login page
      await super.requireAuthentication(...arguments);

      return false;
    }
  }

  return true;
}
```

Please note that is mainly a change to the documentation. If I didn't missed something it's only about two changes to the docs: 

1. Change the return values for `requireAuthentication` and `prohibitAuthentication` to `Promise`.
2. Change the integration examples in guides like the following:
   ```patch
   // app/routes/index.js
   import Route from '@ember/routing';
   
   export default IndexRoute extends Route {
   -  beforeModel(transition) {
   +  async beforeModel(transition) {
   -    this.session.requireAuthentication(transition, 'login');
   +    await this.session.requireAuthentication(transition, 'login');
     },
   }
   ```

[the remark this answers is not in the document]
marcoow: I think you're describing a very specific case here and I don't think that should be covered by ESA by default as it would not actually be relevant to > 90% of its users. Also there seems to be a mis-conception here: what you're describing really is trying to **authenticate** the session before calling `requireAuthentication` which is totally fine and you can do that but that's what ESA has authenticators for. Also an authenticator isn't required to store any credentials in the session (which you'd not even be able to do in case of a http-only cookie-based session). So I think what you're describing is really just an SSO **authenticator**. You could do something like this (without `requireAuthentication` being async):

```js
async beforeModel(transition) {
  await this.tryAuthenticationViaSSO();
  this.session.requireAuthentication(transition, 'login');
}

tryAuthenticationViaSSO() {
  if (!this.session.isAuthenticated) {
    try {
      await this.session.authenticate('authenticator:sso')
    } catch (e) {
      // redirect to login here or let requireAuthentication above handle that
    }
  }
}
```

The current implementation of `requireAuthentication` works essentially the same way the `AuthenticatedRouteMixin` works and I think we don't need to change that to accomodate the case you describe. The general strategy in ESA is to provide the primitives one needs to implement all kinds of (or really any) auth patterns and flows plus providing sensible default implementations for typical scenarios.

packages/ember-simple-auth/addon/mixins/application-route-mixin.js
```diff
@@ -5,9 +5,16 @@ import { A } from '@ember/array';
...
 import Ember from 'ember';
+import { deprecate } from '@ember/application/deprecations';
 import Configuration from './../configuration';
+
 import isFastBoot from 'ember-simple-auth/utils/is-fastboot';
 
+deprecate('Ember Simple Auth: The ApplicationRouteMixin is now deprecated; it can be safely removed.', false, {
+  id: 'ember-simple-auth.mixins.application-route-mixin',
+  until: '4.0.0'
+});
```
patsy-issa, line 16: We should add a reference to the README step to help users know where to look for instructions on the changes they need to make. As it stands the message can be understood as if one could simply remove the `ApplicationRouteMixin` without the need for further steps.

Procfile.server
```diff
@@ -1 +1 @@
-web: yarn workspace test-app start:demo-api
+web: yarn workspace classic-test-app start:demo-api
```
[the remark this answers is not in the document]
marcoow: Yes, actually this one is wrong since both the FastBoot app as well as the api server should be served from the new test app now – fixed.

packages/ember-simple-auth/addon/mixins/application-route-mixin.js
marcoow: That is actually the case.

patsy-issa: Would they not need to use the session service within their `beforeModel` hooks in order to achieve parity with the mixin usage?

packages/ember-simple-auth/addon/services/session.js
```diff
@@ -212,4 +271,99 @@ export default Service.extend(Evented, {
...
+  requireAuthentication(transition, routeOrCallback) {
+    let isAuthenticated = requireAuthentication(getOwner(this), transition);
+    if (!isAuthenticated) {
+      let argType = typeof routeOrCallback;
+      if (argType === 'string') {
+        triggerAuthentication(getOwner(this), routeOrCallback);
+      } else if (argType === 'function') {
+        routeOrCallback();
+      } else {
+        assert(`The second argument to requireAuthentication must be a String of Function, got "${argType}"!`, false);
```
xomaczar, line 302: :s/String of Function/String `or` Function/

packages/ember-simple-auth/addon/services/session.js
```diff
@@ -212,4 +271,99 @@ export default Service.extend(Evented, {
...
+  prohibitAuthentication(routeOrCallback) {
+    let isAuthenticated = this.get('isAuthenticated');
+    if (isAuthenticated) {
+      let argType = typeof routeOrCallback;
+      if (argType === 'string') {
+        prohibitAuthentication(getOwner(this), routeOrCallback);
+      } else if (argType === 'function') {
+        routeOrCallback();
+      } else {
+        assert(`The second argument to prohibitAuthentication must be a String of Function, got "${argType}"!`, false);
```
xomaczar, line 326: :s/String of Function/String `or` Function/

packages/ember-simple-auth/addon/services/session.js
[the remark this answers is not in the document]
marcoow: Resolving for now – @jelhan feel free to unresolve if you think this remains relevant.

packages/ember-simple-auth/addon/mixins/application-route-mixin.js
marcoow: No, the initializer does everything automatically now – see https://github.com/simplabs/ember-simple-auth/pull/2198/files#diff-db561403ef670c68769b736f6ccfacfcR14

README.md
```diff
@@ -593,22 +596,23 @@ The session can then be authenticated or invalidated from the host app or any
...
-  }
-}
+export default class IndexRoute extends Route {
+  @service session;
 
-```
+  beforeModel(transition) {
+    this.get('session').requireAuthentication(transition, () => this.transitionToExternal('login'));
+  },
+}
```
patsy-issa, line 615: Can't add the suggestion but missing a closing triple ` here

marcoow: fixed 👍

patsy-issa:
hmm any ideas on why the specs are failing?

marcoow:
@patsy-issa I think we're running into this: ember-cli/babel-plugin-ember-modules-api-polyfill#64

rwjblue:
@marcoow - ember-cli-babel@7.20.1 was just released, and should fix the underlying issue referenced in https://github.com/ember-cli/babel-plugin-ember-modules-api-polyfill/issues/64.

https://github.com/babel/ember-cli-babel/releases/tag/v7.20.1

marcoow:
Thanks @rwjblue ❤️

marcoow:
Unfortunately we're running into ember-cli/babel-plugin-ember-modules-api-polyfill#110 now…

marcoow:
@patsy-issa CI is green now

pjcarly:
Great to see this merged. Thanks for the amazing work!

Any idea when we can expect a release?

guides/managing-current-user.md
```diff
@@ -162,3 +151,28 @@ export default class ApplicationRoute extends Route.extend(ApplicationRouteMixin
...
+be called in that cases as well. We can do that by overriding the session
+sevice's method in a custom extension of Ember Simple Auth's standard session
+service:
+
+```js
+import { inject as service } from '@ember/service';
+import SessionService from 'ember-simple-auth/services/session';
+
+export default SessionService.extend({
+  @service currentUser;
```
knownasilya, line 166: This is using class decorator syntax but not a class. Would submit a pr but on my phone.

marcoow: thanks – see #2210

sdebarros:
@pjcarly we released a [beta version](https://github.com/simplabs/ember-simple-auth/releases/tag/3.1.0-beta.0) including this today. 🎉

PR #2209, opened 2020-06-05
Re-enable FastBoot tests

### LINKED ISSUE #1719: Re-enable *mandatory* fastboot tests in CI

marcoow:
This re-enables the FastBoot tests using ember-cli-fastboot-testing. It also moves the FastBoot tests into the new `test-app`. It removes some of the old tests that didn't really make sense anyway – we had special routes in the app to trigger certain functionality and I think we should never have done that in the first place and definitely not port these tests over now.

closes #1719


marcoow:
Hm, not sure why the tests are failing – seems like something is wrong with the test that passes an authenticated session manually in the cookie: https://github.com/simplabs/ember-simple-auth/pull/2209/files#diff-4507cbd2f9d136062847ed8f9547b662R19

sdebarros:
@marcoow I updated the test, it was failing because some API calls where not mocked and the error would cause the session to be invalidated. Now they failed with this error:
```
Error while processing route: protected najax does not seem to be defined in your app. Did you override it via `addOrOverrideSandboxGlobals` in the fastboot server?
```

I finally figured out why it happens, we are using ember-cli-fastboot-testing v0.4.0, which has fastboot 3 as a dependency, and fastboot 3 doesn't include `najax` as a global anymore, so we either have to set the global in a config (see https://embermap.github.io/ember-cli-fastboot-testing/docs/fastboot-configuration) (and maybe also update our fastboot dependencies to v3?) or use v0.3.0 of ember-cli-fastboot-testing.

marcoow:
@sdebarros great 🎉 The FastBoot tests have been disabled for some time so the dependencies are most likely horribly outdated, yeah. Feel free to upgrade everything as necessary – it's all just `devDependencies` anyway…

marcoow:
Ah, nevermind – I had removed the outdated dependencies already of course 🤦 But upgrading to the latest ember-cli-fastboot-testing and configuring `najax` as a global seems good. Can you push that?

sdebarros:
@marcoow done, tests are passing now 🎉

marcoow:
Awesome – last step missing is only to re-enable them to run in CI 🎉

marcoow:
@sdebarros nevermind me again, I guess I'm already in vacation mode – of course the tests run as part of `ember test`…

PR #2210, opened 2020-06-05
Fix code example in the "Managing Current User" guide

marcoow:
We were using a class decorator inside of an Ember.Object – see https://github.com/simplabs/ember-simple-auth/pull/2198#discussion_r435963882

### DIFF
--- a/guides/managing-current-user.md
+++ b/guides/managing-current-user.md
@@ -160,9 +160,9 @@ service:
 
 ```js
 import { inject as service } from '@ember/service';
-import SessionService from 'ember-simple-auth/services/session';
+import BaseSessionService from 'ember-simple-auth/services/session';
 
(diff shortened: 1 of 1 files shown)


guides/managing-current-user.md
```diff
@@ -160,9 +160,9 @@ service:
 
 ```js
-import SessionService from 'ember-simple-auth/services/session';
+import BaseSessionService from 'ember-simple-auth/services/session';
 
-export default SessionService.extend({
+export default SessionService extends BaseSessionService {
```
knownasilya, line 165: class missing 😂

knownasilya: Maybe something like https://github.com/knownasilya/eslint-plugin-doc-code-blocks could help

PR #2215, opened 2020-06-12
fix: Incorrect path to fetch the routeAfterAuthentication property in the config object

LuisAverhoff:
Simple fix to properly get the `routeAfterAuthentication` property from the `config` object upon application load.

### DIFF
--- a/packages/ember-simple-auth/addon/configuration.js
+++ b/packages/ember-simple-auth/addon/configuration.js
@@ -40,6 +40,10 @@ export default {
 
   load(config) {
     this.rootURL = getWithDefault(config, 'rootURL', DEFAULTS.rootURL);
-    this.routeAfterAuthentication = getWithDefault(config, 'ember-simple-auth.routeAfterAuthentication', DEFAULTS.routeAfterAuthentication);
-  }
+    this.routeAfterAuthentication = getWithDefault(
+      config,
+      'routeAfterAuthentication',
+      DEFAULTS.routeAfterAuthentication
+    );
+  },
 };
--- a/packages/ember-simple-auth/tests/unit/configuration-test.js
+++ b/packages/ember-simple-auth/tests/unit/configuration-test.js
@@ -31,7 +31,7 @@ describe('Configuration', () => {
     });
 
     it('sets routeAfterAuthentication correctly', function() {
-      Configuration.load({ 'ember-simple-auth': { routeAfterAuthentication: '/some-route' } });
+      Configuration.load({ routeAfterAuthentication: '/some-route' });
 
       expect(Configuration.routeAfterAuthentication).to.eql('/some-route');
     });


marcoow:
I think this might be a misunderstanding – we actually want the configuration to be

```js
'ember-simple-auth': {
  routeAfterAuthentication: '/some-route'
}
```

in `config/environment.js`. Scoping an addon's config to its name prevents clashes with other addons that might have the same settings.

LuisAverhoff:
@marcoow I made this pr because in the app folder of the addon fir the initializer, it is grabbing the ember simple auth config and not the entire ENV config object. If it was the latter, then `ember-simple-auth.routeAfterAuthentication` would be correct but since you are grabbing the config for ember simple auth, to make it work currently you have to do                            


```
'ember-simple-auth:{
    'ember-simple-auth:{
        routeAfterAuthentication: 'some-route'
     }
}

I could update the initializer to just use the entire Env config if that is what you were going for.

**Comment by LuisAverhoff** [comment]
Tell if this is what you were going for.

`app/initializers/ember-simple-auth.js`

```js
import ENV from '../config/environment';
import Configuration from 'ember-simple-auth/configuration';
import setupSession from 'ember-simple-auth/initializers/setup-session';
import setupSessionService from 'ember-simple-auth/initializers/setup-session-service';
import setupSessionRestoration from 'ember-simple-auth/initializers/setup-session-restoration';

export default {
  name: 'ember-simple-auth',

  initialize(registry) {
    Configuration.load(ENV);

    setupSession(registry);
    setupSessionService(registry);
    setupSessionRestoration(registry);
  }
};
```

`addon/configuration.js`

```js
import { getWithDefault } from '@ember/object';

const DEFAULTS = {
  rootURL: '',
  routeAfterAuthentication: 'index',
};

/**
  Ember Simple Auth's configuration object.
  @class Configuration
  @extends Object
  @module ember-simple-auth/configuration
  @public
*/
export default {
  /**
    The root URL of the application as configured in `config/environment.js`.
    @property rootURL
    @readOnly
    @static
    @type String
    @default ''
    @public
  */
  rootURL: DEFAULTS.rootURL,

  /**
    The route to transition to after successful authentication.
    @property routeAfterAuthentication
    @readOnly
    @static
    @type String
    @default 'index'
    @public
  */
  routeAfterAuthentication: DEFAULTS.routeAfterAuthentication,

  load(config) {
    this.rootURL = config.rootURL || config.baseURL || DEFAULTS.rootURL;
    this.routeAfterAuthentication = getWithDefault(config, 'ember-simple-auth.routeAfterAuthentication', DEFAULTS.routeAfterAuthentication);
  }
};
```

LuisAverhoff:
@marcoow Based on your comment, I reverted back how the test was and passed in the Entire ENV config object.

marcoow:
Ah, sorry my comment was wrong of course and what you were saying initially is completely correct – this is in fact a bug. Can you remove the last commit though? I think the original fix is better. Sorry for the confusion!

LuisAverhoff:
@marcoow Alright, I've reverted to the previous commit as you asked.

bitwolfe:
@marcoow Any chance we could get a new release with this fix?

PR #2227, opened 2020-07-22
Removed Google Analytics

### LINKED ISSUE #2225: Remove Google Analytics from ember-simple-auth.com
We currently use Google Analytics without a cookie banner or opt-in process for users. That's not legal as far as I understand it and we need to stop it. The numbers we get from GA are not really relevant for anyone anyway so we might as well stop collecting the data altogether.

Simple removing the [`<script>` block](https://github.com/simplabs/ember-simple-auth/blob/gh-pages/index.html#L509) should do the job.

JaninaBodi:
closes #2225

### DIFF
--- a/index.html
+++ b/index.html
@@ -508,13 +508,6 @@
     <script src="https://yandex.st/highlightjs/7.4/highlight.min.js"></script>
     <script type="text/javascript">
       hljs.initHighlightingOnLoad();
-      (function(i,s,o,g,r,a,m){i['GoogleAnalyticsObject']=r;i[r]=i[r]||function(){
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -506,15 +506,5 @@ <h2>Projects and Companies using Ember Simple Auth</h2>
       </div>
     </div>
     <script src="https://yandex.st/highlightjs/7.4/highlight.min.js"></script>
-    <script type="text/javascript">
-      hljs.initHighlightingOnLoad();
```
mansona, line 510: do we still need highlightjs here?

marcoow: I think it's fine to remove this but should be done in a separate PR.

marcoow (approved):

PR #2239, opened 2020-08-27
add docs and test for `sameSite` attribute in AdaptiveStorage

### LINKED ISSUE #2230: `sameSite` isn't proxied to cookie store when using adaptive store
Hi,

I was trying to configure the `sameSite` property which was introduced in https://github.com/simplabs/ember-simple-auth/pull/1848 .

It seems like the `sameSite` property isn't proxied to the cookie store when it's created:

https://github.com/simplabs/ember-simple-auth/blob/master/packages/ember-simple-auth/addon/session-stores/adaptive.js#L124

If I simply add `sameSite` to the properties list of `getProperties`, it works as expected. 

Would it be ok if I create a PR to add the property there or do you also want an additional test (it might be a bit complicated because in adaptive store test, the cookie store is created [outside the `_createStore` method](https://github.com/simplabs/ember-simple-auth/blob/master/packages/ember-simple-auth/tests/helpers/create-adaptive-store.js#L12-L22))?

makepanic:
#2312 added proxying of the attribute but did not add proper documentation and testing.

### DIFF
--- a/packages/ember-simple-auth/addon/session-stores/adaptive.js
+++ b/packages/ember-simple-auth/addon/session-stores/adaptive.js
@@ -66,6 +66,23 @@ export default Base.extend({
   cookieDomain: proxyToInternalStore(),
 
   /**
+    Allows servers to assert that a cookie ought not to be sent along with
+    cross-site requests, which provides some protection against cross-site
+    request forgery attacks (CSRF).
+
+    Available options:
+    - "Strict"
+    - "Lax"
+
+    @property sameSite
+    @type String
+    @default null
+    @public
+  */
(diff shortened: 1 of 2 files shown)


makepanic:
From the issue description:

> adding the new property to the adaptive store and documenting it

I've copied the documentation from the cookie store `sameSite` jsdoc comment.

> passing it to the cookie store

done 

> adding a test for that behavior (I know the adaptive store tests are a mess – if you need help with that, let me know)

I've extended the existing cookie attribute related adaptive test.

BobrImperator (approved):

PR #2242, opened 2020-09-18
[WIP] Remove ember-cli-mocha dependency

sdebarros:
We're including both `ember-cli-mocha` (which is deprecated) and `ember-mocha`.


marcoow:
this was done in #2330

PR #2245, opened 2020-10-01
set up code scanning

marcoow:
GitHub code scanning is generally available and this sets it up for ESA.

I haven't looked into it **at all** and don't really know what this will bring or if it makes sense for us at all so let's see what it even comes up with…

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -1,0 +1,37 @@
+name: CodeQL
+
+on:
+  push:
+    branches:
+      - master
+      - 'v*'
+  pull_request: {}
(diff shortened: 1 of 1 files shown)


.github/workflows/codeql.yml
```diff
@@ -0,0 +1,37 @@
...
+
+jobs:
+  analyze:
+    name: Analyze
+    runs-on: ubuntu-latest
+
+    strategy:
+      fail-fast: false
+      matrix:
+        language: ['javascript']
```
sdebarros, line 20: Do we need a matrix if it's just one language?

PR #2251, opened 2020-11-06
Add release-it to simplify the release process

sdebarros:
Using template from https://github.com/rwjblue/create-rwjblue-release-it-setup. This automates the process of updating the CHANGELOG, creating the tag and the Github release, the npm publishing is still done by the CI.

### DIFF
--- a/.release-it.json
+++ b/.release-it.json
@@ -1,0 +1,19 @@
+{
+  "plugins": {
+    "release-it-lerna-changelog": {
+      "infile": "CHANGELOG.md"
+    },
+    "release-it-yarn-workspaces": {
+      "publish": false
+    }
+  },
+  "git": {
+    "commitMessage": "v${version}"
+  },
+  "github": {
+    "release": true,
+    "releaseName": "${version}",
+    "tokenRef": "GITHUB_AUTH"
+  },
+  "npm": false
+}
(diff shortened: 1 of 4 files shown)


.release-it.json
```diff
@@ -0,0 +1,19 @@
...
+    "release-it-lerna-changelog": {
+      "infile": "CHANGELOG.md"
+    },
+    "release-it-yarn-workspaces": {
+      "publish": false
+    }
+  },
+  "git": {
+    "commitMessage": "v${version}"
+  },
```
Turbo87, line 12: I would recommend https://github.com/simplabs/qunit-dom/blob/master/.release-it.js#L11 to match what we're currently doing here

package.json
```diff
@@ -4,13 +4,16 @@
...
   "workspaces": [
     "packages/*"
   ],
   "devDependencies": {
     "@mermaid-js/mermaid-cli": "^8.4.8",
-    "lerna-changelog": "^1.0.1"
+    "lerna-changelog": "^1.0.1",
+    "release-it": "^14.2.0",
+    "release-it-yarn-workspaces": "^2.0.0"
   }
```
Turbo87, line 18: release-it-lerna-changelog missing here?

.release-it.json
sdebarros: Our tags so far have been without the `v` 😬 So I was maintaining that, but I'm fine with changing it.

Turbo87: 😱 sorry, nevermind then 😅

package.json
sdebarros: Added.

Turbo87 (approved):

PR #2261, opened 2021-01-22
Allow Ember canary tests to fail

sdebarros:
It seems the original intention was to allow Ember canary tests to fail, but the current configuration wasn't doing that (see https://github.com/simplabs/ember-simple-auth/runs/1746492928?check_suite_focus=true).

This change fixes that.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -61,7 +61,6 @@ jobs:
           - test:one ember-lts-3.8
           - test:one ember-lts-3.12
           - test:one ember-beta
-          - test:one ember-canary
         allow-failure: [false]
         include:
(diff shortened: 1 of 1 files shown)


marcoow (approved):
CI still seems to fail though?

sdebarros:
> CI still seems to fail though?

@marcoow This is still in Draft 😅 The failing tests are for beta, not canary, and should be solved by https://github.com/simplabs/ember-simple-auth/pull/2262. Do we want to allow beta tests to also fail?

marcoow:
Ah sorry, I'd say ideally we'd want beta tests to pass

marcoow (approved):

PR #2262, opened 2021-01-22
Fix tests failing in master

sdebarros:
[Tests in master are failing](https://github.com/simplabs/ember-simple-auth/runs/1746492928?check_suite_focus=true#step:7:193) for the canary and beta versions, it seems to be related to the latest Ember not working well with a dependency of the `ember-bootstrap` we have. 

This is a temporary solution to allow us to release version 3.1.0, since we only use `ember-bootstrap` for the test apps it shouldn't be an issue. Upgrading `ember-bootstrap` on the actual app, breaks tests for Ember 3.12. I think we should consider removing the `ember-bootstrap` dependency and just have our own CSS.

### DIFF
--- a/packages/ember-simple-auth/config/ember-try.js
+++ b/packages/ember-simple-auth/config/ember-try.js
@@ -105,6 +105,7 @@ module.exports = function() {
             devDependencies: {
               'ember-data': 'beta',
               'ember-source': betaUrl,
+              'ember-bootstrap': '^4.6.2',
             },
           },
         },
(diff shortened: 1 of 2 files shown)


marcoow (approved):
Makes sense 👍 

I think we can also replace the dependency with a simple `<link>` tag to load the bootstrap styles from their CDN – I don't think we use any of the addon's additional functionality. Bootstrap as such might be a good thing to keep (for now) since moving to our own CSS would be a bit of a bigger step **and** we'd be adding more code to maintain that's not actually relevant for ESA.

PR #2263, opened 2021-01-28
Fix empty object check in `_renew()` cookie session store

reidab:
The `_renew` implementation in the cookie session store includes a check for `data !== {}`.

While this deceptively looks like it's checking for an empty object, `{} === {}` is `false` in JavaScript so the check never resolved. This replaces that check with an alternative that correctly identifies empty objects.

### DIFF
--- a/packages/ember-simple-auth/addon/session-stores/cookie.js
+++ b/packages/ember-simple-auth/addon/session-stores/cookie.js
@@ -287,7 +287,7 @@ export default BaseStore.extend({
 
   _renew() {
     return this.restore().then((data) => {
-      if (!isEmpty(data) && data !== {}) {
+      if (!isEmpty(data) && !(data.constructor === Object && Object.keys(data).length === 0)) {
         data = typeOf(data) === 'string' ? data : JSON.stringify(data || {});
         let expiration = this._calculateExpirationTime();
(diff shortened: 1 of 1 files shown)


packages/ember-simple-auth/addon/session-stores/cookie.js
```diff
@@ -287,7 +287,7 @@ export default BaseStore.extend({
 
   _renew() {
     return this.restore().then((data) => {
-      if (!isEmpty(data) && data !== {}) {
+      if (!isEmpty(data) && !(data.constructor === Object && Object.keys(data).length === 0)) {
```
marcoow, line 290: I'm not sure why `data !== {}` was added at all since [`isEmpty`](https://api.emberjs.com/ember/3.24/functions/@ember%2Futils/isEmpty) should handle the case we're trying to prevent against here anyway?

reidab: Also unintuitively, Ember's `isEmpty` also doesn't check for empty objects: https://api.emberjs.com/ember/release/classes/@ember%2Futils/methods?anchor=isEmpty

```javascript
isEmpty({});               // false
```

marcoow (approved):

PR #2264, opened 2021-01-29
Copy README to ember-simple-auth package before NPM release

sdebarros:
Our last releases don't show the README file in [npm](https://www.npmjs.com/package/ember-simple-auth), this is because the file now doesn't leave at the root of the package.

This copies the README to the ember-simple-auth package folder before doing the release, so it shows up in NPM, while still remaining at the root of our repository and visible by default in https://github.com/simplabs/ember-simple-auth.


marcoow (approved):
😱 🙌

PR #2266, opened 2021-02-01
remove `tryInvoke` usage

bekzod:
As `tryInvoke` seems to be deprecated


marcoow (approved):
🙌 

Linting seems to fail though – I assume the eslint config needs to be changed or we'll have to use something like

```js
server && server.shutdown();
```

bekzod:
@marcoow thank you, updated

PR #2284, opened 2021-03-24
Update README.md

mcfiredrill:
It's not necessary to use `get` when accessing the session service in the example in the README.


pichfl (approved):
👍

PR #2285, opened 2021-03-24
add routeAfterAuthentication to the README

mcfiredrill:
Also I linked to the API docs on the "configurable route" text.


mcfiredrill:
My bad, I know this method of configuring routeAfterAuthentication was dropped in favor of the mixin, but now the mixin is deprecated.
So how *exactly* does one set routeAfterAuthentication ???

I searched the discord a bit and found someone who got it working by extending ember-simple-auth session service and overwriting the handleAuthentication() method. Is this the recommended solution?
https://discord.com/channels/480462759797063690/496695347582861334/738599436917407934
```js
import SimpleAuthSessionService from 'ember-simple-auth/services/session';

export default class SessionService extends SimpleAuthSessionService {
    /**
     * Inject the router service
     *
     * @var {Service}
     */
    @service router;

    /**
     * Inject the current user service
     *
     * @var {Service}
     */
    @service currentUser;

    /**
     * Overwrite the handle authentication method
     *
     * @var {Service}
     */
    handleAuthentication() {
        this.router.transitionTo('dashboard.boards');
        this.loadCurrentUser();
    }

    /**
     * Loads the current authenticated user
     *
     * @void
     */
    async loadCurrentUser() {
        try {
            const user = await this.currentUser.load();
            return user;
        } catch (err) {
            await this.session.invalidate();
        }
    }
}
```

mcfiredrill:
> Yes, what you found is the recommended solution +1

All right thanks in that case I will add some docs for that in another PR. :ok_person:

PR #2294, opened 2021-04-27
Fix tests

marcoow:
This fixes CI by:

* running `ember init` on all packages so that the project files are up to date with the latest blueprints in ember-cli and use the latest dependencies for the ember and ember-cli versions we're using
* allowing the tests for the release version of ember for the test app and classic test app to fail since we're running into emberjs/ember.js#19432 (which should be fixed by dropping ember-bootstrap – see #2292, #2293)
* fixing some tests for deprecations that were breaking since they didn't take into account that additional deprecations could be introduced by other packages, e.g. Ember itself


packages/classic-test-app/testem.js
```diff
@@ -1,7 +1,6 @@
-/* eslint-env node */
+'use strict';
 
 module.exports = {
-  framework: 'mocha',
```
sdebarros, line 4: Is this not needed anymore?

marcoow: This is what `ember init` generates but it seems fine since the [testem docs](https://github.com/testem/testem/blob/master/docs/config_file.md#common-configuration-options) say:

> **framework** - the test framework that you are using, in the browser, in the case that you are not also using the test_page option.

locks (approved):

PR #2297, opened 2021-05-07
Run CI on latest Ember LTS releases

marcoow:
This adds the more recent LTS releases of Ember (3.16, 3.20, 3.24) to our testing matrix on CI.


.github/workflows/ci.yml
```diff
@@ -60,6 +60,9 @@ jobs:
           - test:one ember-lts-3.4
           - test:one ember-lts-3.8
           - test:one ember-lts-3.12
```
sdebarros, line 62: Do we still want to keep these? I think the idea was to test the last 3 lts releases.

marcoow: I think I'd test all LTS releases that we support (until that leads to problems when we're using too many GitHub Actions resources or so).

sdebarros (approved):

PR #2304, opened 2021-05-28
move application route to dummy

### LINKED ISSUE #1854: Inconsistent builds with ember-cli-typescript

inputs:

`/app/routes/application.ts`

how to fix?

why this issue?

### LINKED ISSUE #2279: Default application route breaks embroider

BobrImperator:
ESA main package doesn't have any need for application route.
This PR moves application route to the dummy app which needs it for testing, but doesn't have to ship it to user apps.

- Move application route to dummy

Should close #1854, and close #2279


marcoow:
I think we need the application route to exist for [this initializer](https://github.com/simplabs/ember-simple-auth/blob/master/packages/ember-simple-auth/addon/initializers/setup-session-restoration.js) to work. Try removing it and the dummy app should fail.

That said, it should be possible I think to get the application's `application` route without the file being present (we should simply receive the route instance that Ember generates on-the-fly in that case) but I'm not sure why that's not being done currently – maybe wasn't possible with an earlier version of Ember?

BobrImperator:
@marcoow I know that Application route is needed for the initializer, but it doesn't have to exist in ember-simple-auth addon, but only in the dummy that is the 'real' app that uses the addon.

Basically meaning that user apps have their application routes defined already.

marcoow:
Yes but we don't want people to have to create the file in their apps for ESA to work since that's a major footgun.

BobrImperator:
Apps *must* have an application route, be it `application.js` or `application.hbs` where for the latter there's an implicit js class created, or am I wrong here?

marcoow:
You don't have to have the `app/routes/application.js` file.

BobrImperator:
closed in favor of #2307

PR #2307, opened 2021-06-09
Lookup and modify ApplicationRoute after EmberRouter is initialized

### LINKED ISSUE #1854: Inconsistent builds with ember-cli-typescript

### LINKED ISSUE #2279: Default application route breaks embroider

BobrImperator:
closes #1854, and closes #2279

It extends EmberRouter#setupRouter, looks up ApplicationRouter and extends it with session service methods.
The reason for extending EmberRouter is that we need to extend user's ApplicationRoute without providing our own, or even users might not have *explicit* ApplicationRoute themselves, but only it's template.

Routes are dynamically registered and instantiated during transitions.

### DIFF
--- a/packages/ember-simple-auth/addon/initializers/setup-session-restoration.js
+++ b/packages/ember-simple-auth/addon/initializers/setup-session-restoration.js
@@ -1,27 +1,35 @@
 import { getOwner } from '@ember/application';
 
 export default function setupSessionRestoration(registry) {
-  const ApplicationRoute = registry.resolveRegistration
-    ? registry.resolveRegistration('route:application')
-    : registry.resolve('route:application');
+  const Router = registry.resolveRegistration ? registry.resolveRegistration('router:main') : registry.resolve('router:main');
 
-  ApplicationRoute.reopen({
-    init() {
-      this._super(...arguments);
+  Router.reopen({
+    setupRouter() {
+      const superValue = this._super(...arguments);
+      const ApplicationRoute =
+        getOwner(this).factoryFor('route:application').class;
 
-      const originalBeforeModel = this.beforeModel;
-      this.beforeModel = function() {
-        if (!this.__usesApplicationRouteMixn__) {
-          const sessionService = getOwner(this).lookup('service:session');
-          sessionService._setupHandlers();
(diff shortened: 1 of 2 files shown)


packages/ember-simple-auth/addon/initializers/setup-session-restoration.js
```diff
@@ -1,27 +1,35 @@
 import { getOwner } from '@ember/application';
 
 export default function setupSessionRestoration(registry) {
-  const ApplicationRoute = registry.resolveRegistration
-    ? registry.resolveRegistration('route:application')
-    : registry.resolve('route:application');
+  const Router = registry.resolveRegistration ? registry.resolveRegistration('router:main') : registry.resolve('router:main');
```
Turbo87, line 4: I'm somewhat sure that this is private API, and I think using `reopen()` is generally deprecated 🤔

BobrImperator: I Wasn't really thinking about doing any refactoring here, but yeah, EmberObject apis in general are deprecated.
I'm not sure what we can do about it right now, IMO the way forward would be to remove the initializer, and if we have any crucial setup then users should probably call some method in their application routes.

marcoow (approved):
I think this is generally fine for now. Not sure about the private API topic etc. – I'd leave that to the 2 of you to decide.

In the mid/long term, I agree we should deprecate the initializer completely. In the end, this could be replaced with something in the README telling people to add

```js
async beforeModel() {
  await this.session.setup();
}
```

to their `application` routes. We could even check in all other methods on the session service whether `setup` was called correctly or not and raise an error if necessary. Until we do that, this is fine from my perspective as a temporary solution to make it easier for people to use TypeScript (I think the fact that we merge an `app/routes/application.js` into apps is a hard to understand problem for many that want to use TypeScript).

azhiv:
Hey guys, when are you going to release this?

BobrImperator:
Closing in favor of #2318 which is going to depreacate the use of the initializer.

PR #2312, opened 2021-07-13
[2308] Refactor Adaptive store to use DI

### LINKED ISSUE #2308: AdaptiveStore should use DI instead of initializing other stores by itself
It splits #2302
While working on #2306 we've discovered that store related tests are failing due to them not being isolated - after refactoring internals to use explicitly injected services and refactoring tests accordingly the tests would leak their instances to other tests in some cases e.g itBehavesLikeStore causes this to happen, this issue should make it easier to test AdaptiveStore

# TODOS
- [ ] register `session-store:cookie` `session-store:local-storage`
- [ ] use registered session-stores in AdaptiveStore directly
- [ ] refactor AdaptiveStore tests to use Ember DI

BobrImperator:
- registers session-stores in Ember's DI system.
- refactors AdaptiveStore to use registered CookieStore and LocalStorageStore.
- modifies test to use `setupTest` so there's access to ember's container
- the side effect of `setupTest` is that it seems to require the test to be asynchronous, so all test functions are marked as `async` so they return a promise implicitly
- changes how CookieStore is initialized by implementing `_initialize` method that handles setting of properties that would normally be provided during creation of CookieStore instance.
- changed e.g `cookieName` property to `_cookieName` in some tests because using 'non private' property would override the underlying computed property which acts as a proxy, because of this behaviour, the private 'default' values are changed and not the public ones.

closes #2308

### DIFF
--- a/packages/ember-simple-auth/addon/session-stores/adaptive.js
+++ b/packages/ember-simple-auth/addon/session-stores/adaptive.js
@@ -139,10 +139,13 @@ export default Base.extend({
       options._isFastBoot = false;
       store.setProperties(options);
     } else {
-      store = getOwner(this).lookup(`session-store:cookie`, { dupa: true });
+      store = this.get('cookie');
       const options = this.getProperties('sameSite', 'cookieDomain', 'cookieName', 'cookieExpirationTime', 'cookiePath');
+
       store._initialize(options);
+      this.set('cookieExpirationTime', store.get('cookieExpirationTime'));
     }
+
     this.set('_store', store);
     this._setupStoreEvents(store);
   },
--- a/packages/ember-simple-auth/addon/session-stores/adaptive.js
+++ b/packages/ember-simple-auth/addon/session-stores/adaptive.js
@@ -5,12 +5,6 @@ import Base from 'ember-simple-auth/session-stores/base';
 
 const LOCAL_STORAGE_TEST_KEY = '_ember_simple_auth_test_key';
 
-const injectStore = (store) => {
-  return computed(function() {
-    return getOwner(this).lookup(`session-store:${store}`);
-  });
-};
-
 const proxyToInternalStore = function() {
   return computed({
     get(key) {
@@ -56,8 +50,6 @@ export default Base.extend({
     @public
   */
   localStorageKey: 'ember_simple_auth-session',
-
-  localStorage: injectStore('local-storage'),
 
   /**
     The domain to use for the cookie if `localStorage` is not available, e.g.,
@@ -112,7 +104,6 @@ export default Base.extend({
   sameSite: proxyToInternalStore(),
 
   _cookies: service('cookies'),
-  cookie: injectStore('cookie'),
 
   _isLocalStorageAvailable: computed({
     get() {
@@ -134,16 +125,21 @@ export default Base.extend({
 
     let store;
     if (this.get('_isLocalStorageAvailable')) {
-      store = this.get('localStorage');
+      const localStorage = owner.lookup('session-store:local-storage');
       const options = { key: this.get('localStorageKey') };
+
       options._isFastBoot = false;
-      store.setProperties(options);
+      localStorage.setProperties(options);
+
+      store = localStorage;
     } else {
-      store = this.get('cookie');
+      const cookieStorage = owner.lookup('session-store:cookie');
       const options = this.getProperties('sameSite', 'cookieDomain', 'cookieName', 'cookieExpirationTime', 'cookiePath');
 
-      store._initialize(options);
-      this.set('cookieExpirationTime', store.get('cookieExpirationTime'));
+      cookieStorage._initialize(options);
+      this.set('cookieExpirationTime', cookieStorage.get('cookieExpirationTime'));
+
+      store = cookieStorage;
     }
 
     this.set('_store', store);
--- a/packages/ember-simple-auth/addon/session-stores/adaptive.js
+++ b/packages/ember-simple-auth/addon/session-stores/adaptive.js
@@ -2,10 +2,14 @@ import { computed } from '@ember/object';
 import { inject as service } from '@ember/service';
 import { getOwner } from '@ember/application';
 import Base from 'ember-simple-auth/session-stores/base';
-import LocalStorage from 'ember-simple-auth/session-stores/local-storage';
-import Cookie from 'ember-simple-auth/session-stores/cookie';
 
 const LOCAL_STORAGE_TEST_KEY = '_ember_simple_auth_test_key';
+
+const injectStore = (store) => {
+  return computed(function() {
+    return getOwner(this).lookup(`session-store:${store}`);
+  });
+};
 
 const proxyToInternalStore = function() {
   return computed({
@@ -52,6 +56,8 @@ export default Base.extend({
     @public
   */
   localStorageKey: 'ember_simple_auth-session',
+
+  localStorage: injectStore('local-storage'),
 
   /**
     The domain to use for the cookie if `localStorage` is not available, e.g.,
@@ -102,7 +108,21 @@ export default Base.extend({
   _cookieExpirationTime: null,
   cookieExpirationTime: proxyToInternalStore(),
 
+  _sameSite: null,
+  sameSite: proxyToInternalStore(),
+
   _cookies: service('cookies'),
+  cookie: injectStore('cookie'),
+
+  _isLocalStorageAvailable: computed({
+    get() {
+      return testLocalStorageAvailable();
+    },
+
+    set(key, value) {
+      return value;
+    }
+  }),
 
   init() {
     this._super(...arguments);
@@ -112,28 +132,22 @@ export default Base.extend({
       this._fastboot = owner.lookup('service:fastboot');
     }
 
-    this._isLocalStorageAvailable = this.hasOwnProperty('_isLocalStorageAvailable') ? this._isLocalStorageAvailable : testLocalStorageAvailable();
-
     let store;
     if (this.get('_isLocalStorageAvailable')) {
+      store = this.get('localStorage');
       const options = { key: this.get('localStorageKey') };
       options._isFastBoot = false;
-      store = this._createStore(LocalStorage, options);
+      store.setProperties(options);
     } else {
-      const options = this.getProperties('cookieDomain', 'cookieName', 'cookieExpirationTime', 'cookiePath');
-      options._fastboot = this.get('_fastboot');
-      options._cookies = this.get('_cookies');
-
-      store = this._createStore(Cookie, options);
-      this.set('cookieExpirationTime', store.get('cookieExpirationTime'));
+      store = getOwner(this).lookup(`session-store:cookie`, { dupa: true });
+      const options = this.getProperties('sameSite', 'cookieDomain', 'cookieName', 'cookieExpirationTime', 'cookiePath');
+      store._initialize(options);
     }
     this.set('_store', store);
+    this._setupStoreEvents(store);
   },
 
-  _createStore(storeType, options) {
-    let owner = getOwner(this);
-    const store = storeType.create(owner.ownerInjection(), options);
-
+  _setupStoreEvents(store) {
     store.on('sessionDataUpdated', (data) => {
       this.trigger('sessionDataUpdated', data);
     });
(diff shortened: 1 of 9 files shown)


packages/ember-simple-auth/addon/session-stores/adaptive.js
```diff
@@ -53,6 +57,8 @@ export default Base.extend({
   */
   localStorageKey: 'ember_simple_auth-session',
 
+  localStorage: injectStore('local-storage'),
```
marcoow, line 60: Could this be set in `init` instead?

```js
this.localStorage = getOwner(this).lookup(`session-store:local-storage`);
```

packages/ember-simple-auth/addon/session-stores/adaptive.js
```diff
@@ -102,7 +108,21 @@ export default Base.extend({
   _cookieExpirationTime: null,
   cookieExpirationTime: proxyToInternalStore(),
 
+  _sameSite: null,
+  sameSite: proxyToInternalStore(),
+
   _cookies: service('cookies'),
+  cookie: injectStore('cookie'),
```
marcoow, line 115: see above

packages/ember-simple-auth/addon/session-stores/adaptive.js
```diff
@@ -112,28 +132,25 @@ export default Base.extend({
...
+      store.setProperties(options);
     } else {
-      const options = this.getProperties('cookieDomain', 'cookieName', 'cookieExpirationTime', 'cookiePath');
-      options._fastboot = this.get('_fastboot');
-      options._cookies = this.get('_cookies');
+      store = this.get('cookie');
+      const options = this.getProperties('sameSite', 'cookieDomain', 'cookieName', 'cookieExpirationTime', 'cookiePath');
 
-      store = this._createStore(Cookie, options);
+      store._initialize(options);
```
marcoow, line 145: Does

```suggestion
      store.setProperties(options);
```

not work here?

packages/ember-simple-auth/tests/unit/session-stores/adaptive-test.js
```diff
@@ -1,98 +1,104 @@
...
-      store() {
+      store(sinon, owner) {
+        let store;
+        let cookieService;
+        owner.register('service:cookies', FakeCookieService);
+        cookieService = owner.lookup('service:cookies');
+        sinon.spy(cookieService, 'read');
+        sinon.spy(cookieService, 'write');
+        store = createAdaptiveStore(cookieService, {
+          _isLocal: false,
```
marcoow, line 23: This doesn't seem to exist anywhere actually? 🤔

packages/ember-simple-auth/tests/unit/session-stores/shared/cookie-store-behavior.js
```diff
@@ -311,7 +302,8 @@ export default function(options) {
       );
     });
 
-    it('clears cached expiration times when setting expiration to null', function() {
+    // eslint-disable-next-line require-await
+    it('clears cached expiration times when setting expiration to null', async function() {
```
marcoow, line 306: Is this `async` in fact necessary?

packages/ember-simple-auth/tests/unit/session-stores/shared/cookie-store-behavior.js
```diff
@@ -333,18 +325,21 @@ export default function(options) {
...
-      store = createStore(cookieService);
-    });
-
-    afterEach(function() {
-      cookieService.clear(cookieName);
     });
 
-    it('restores expiration time from cookie', function() {
+    // eslint-disable-next-line require-await
+    it('restores expiration time from cookie', async function() {
```
marcoow, line 342: Is this `async` necessary? There's nothing asynchronous going on in the test?

marcoow (commented):
great work – this is one of the messiest parts of the codebase 🙌

packages/ember-simple-auth/tests/unit/session-stores/adaptive-test.js
BobrImperator: Yeah, that's true, I will remove it, also noticed that some of the `_isFastBoot` properties were miswritten, we have `_isFastBoot` and `_isFastboot` currently somewhere i believe.

packages/ember-simple-auth/tests/unit/session-stores/shared/cookie-store-behavior.js
BobrImperator: Yeah, otherwise the mocha test is failing, it's throwing some error saying `done() wasn't called after the test finished` and it's failing because of that. The test works correctly, the expectations inside in fact are working, though for some reason I wasn't able to debunk the test has to be async.

packages/ember-simple-auth/addon/session-stores/adaptive.js
BobrImperator: I could check that out, although `_initalize` currently needs to treat the `cookieName` property in a special way.

The proxy properties that were receiving some default values on initialization weren't causing any issues before, but right now since we can't really pass any properties during init, but we set them after instead - its causing e.g `cookieName` to be cached with a wrong value.

packages/ember-simple-auth/addon/session-stores/adaptive.js
BobrImperator: I think so, yes 👍

packages/ember-simple-auth/tests/unit/session-stores/shared/cookie-store-behavior.js
marcoow: This sounds like we're working around something – this test itself has nothing asynchronous. We should look into this a bit closer – happy to pair on it next week.

BobrImperator: OK 👍 
I was able to remove it from one of the places, I figured that it was caused by beforeEach calling an async method on store.
Though can't find the cause for the other one so far, although I wouldn't be surprised if it actually be caused by mocha `setupTest` which expects the hook callback to be `thenable` https://github.com/emberjs/ember-mocha/blob/master/addon-test-support/ember-mocha/setup-test.js#L21

packages/ember-simple-auth/addon/session-stores/adaptive.js
```diff
@@ -102,8 +100,21 @@ export default Base.extend({
   _cookieExpirationTime: null,
   cookieExpirationTime: proxyToInternalStore(),
 
+  _sameSite: null,
+  sameSite: proxyToInternalStore(),
+
   _cookies: service('cookies'),
 
+  _isLocalStorageAvailable: computed({
```
marcoow, line 108: Could we not keep the regular property as it's used in [line 115](https://github.com/simplabs/ember-simple-auth/pull/2312/files#diff-db3487bdb1f8c84714dedc96d1cce936b66f804e8a3c58ebe9d88dfb883e50deL115). Once we move away from Ember objects/computeds, that's how we'd have to do it anyway?

packages/ember-simple-auth/addon/session-stores/adaptive.js
```diff
@@ -112,28 +123,32 @@ export default Base.extend({
...
     if (this.get('_isLocalStorageAvailable')) {
+      const localStorage = owner.lookup('session-store:local-storage');
       const options = { key: this.get('localStorageKey') };
+
       options._isFastBoot = false;
-      store = this._createStore(LocalStorage, options);
+      localStorage.setProperties(options);
+
+      store = localStorage;
+      this.localStorage = localStorage;
```
marcoow, line 135: Is this used anywhere?

packages/ember-simple-auth/addon/session-stores/adaptive.js
```diff
@@ -112,28 +123,32 @@ export default Base.extend({
...
+      const cookieStorage = owner.lookup('session-store:cookie');
+      const options = this.getProperties('sameSite', 'cookieDomain', 'cookieName', 'cookieExpirationTime', 'cookiePath');
+
+      cookieStorage._initialize(options);
+      this.set('cookieExpirationTime', cookieStorage.get('cookieExpirationTime'));
 
-      store = this._createStore(Cookie, options);
-      this.set('cookieExpirationTime', store.get('cookieExpirationTime'));
+      store = cookieStorage;
+      this.cookieStorage = cookieStorage;
```
marcoow, line 144: Is this used anywhere?

packages/ember-simple-auth/tests/unit/session-stores/shared/cookie-store-behavior.js
marcoow: Ok, let's figure out how we can fix this one as well.

packages/ember-simple-auth/addon/session-stores/adaptive.js
BobrImperator: 👍  I was under the impression that we want to expose the services as properties as well.

I'll remove that next time.

packages/ember-simple-auth/addon/session-stores/adaptive.js
BobrImperator: Not sure at the moment if this is needed - will check once I pick up the PR again, although I'm pretty sure I needed that to correctly test initialization of services

packages/ember-simple-auth/tests/unit/session-stores/shared/cookie-store-behavior.js
BobrImperator: Yeah, for now I've removed the eslint disable line, so it fails PR until it's fixed.

packages/ember-simple-auth/addon/session-stores/adaptive.js
BobrImperator: Unfortunately we need this to be here, in tests to tell the adaptive store that localStorage is not available we use `Adaptive.extend({ _isLocalStorageAvailable: false })` which would be overwritten on `init`.

I see that the previous init was doing `hasOwnProperty` as well, but that was working only because we were creating the stores manually passing properties to `create`, but `extend` extends the prototype and `_isLocalStorage` is no longer an 'own property'.

packages/ember-simple-auth/addon/session-stores/cookie.js
```diff
@@ -173,11 +173,20 @@ export default BaseStore.extend({
...
+  },
+
+  _initialize(options) {
+    const clonedOptions = Object.assign({}, options);
+    if (clonedOptions.cookieName) {
+      this.set('_cookieName', clonedOptions.cookieName);
+    }
+
+    delete clonedOptions.cookieName;
+    this.setProperties(clonedOptions);
```
marcoow, line 189: Could this not remain in `init`? Having a special `_initialize` method makes things a bit more complex 🤔

packages/ember-simple-auth/tests/helpers/create-adaptive-store.js
```diff
@@ -1,25 +1,13 @@
...
-
-    return cookieStore;
-  };
+  owner.register('session-store:adaptive', Adaptive.extend(Object.assign({
+    _isLocalStorageAvailable: false,
+  }, options)));
 
-  return Adaptive.extend(props).create(options);
+  return owner.lookup('session-store:adaptive');
 }
```
marcoow, line 13: Is this function still necessary? I think it could be replaced in tests with

```js
const store = owner.lookup('session-store:adaptive');
store.setProperties(options);
```

?

packages/ember-simple-auth/addon/session-stores/cookie.js
BobrImperator: No because of this https://github.com/simplabs/ember-simple-auth/pull/2312#discussion_r675576773.

Basically tests had the luxury to add properties while creating the store, but now that the service is initialized by the container, and we still need to set values I've added `_initialize`.
Unfortunately this isn't something we could take care of with just `setProperties` because it'd invalidate a cookie with incorrect name which would make all tests fail.

packages/ember-simple-auth/tests/helpers/create-adaptive-store.js
BobrImperator: No because we need to set `isLocalStorageAvailable` before `init` where the `session-store:local-storage` is initialized. So the only way to do so is extend the class and register.

It's also a part of the reason why I moved `isLocalStorageAvailable` as computed property, so I can override it and read the value on the init. But this instead could potenitally be handled by checking if `'_isLocalStorageAvailable in this` instead of `this.hasOwnProperty('_isLocalStorageAvailable')` but I thought that computed property is a better way.

marcoow (approved):

PR #2313, opened 2021-07-14
Drop node 10 in favor of 12

BobrImperator:
Pipelines are failing to due to some dependencies being incompatible with node 10.
Node 10 also already reach it's end of life.

see #2312

### DIFF
--- a/packages/ember-simple-auth/package.json
+++ b/packages/ember-simple-auth/package.json
@@ -78,7 +78,7 @@
     "torii": "^0.10.0"
   },
   "engines": {
-    "node": "10.* || >= 12.*"
+    "node": ">= 12"
   },
   "ember-addon": {
     "configPath": "tests/dummy/config",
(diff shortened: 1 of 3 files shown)


packages/ember-simple-auth/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "ember-simple-auth",
-  "version": "3.1.0",
+  "version": "4.0.0",
```
marcoow, line 3:
```suggestion
  "version": "3.1.0",
```

marcoow (commented):
Makes sense 👍 (in particular given Node 10 has reached end of life). Let's just not increment the version in this – we can do that once we release.

packages/ember-simple-auth/package.json
BobrImperator: Got it 👍

marcoow (approved):

PR #2315, opened 2021-08-13
[2302]  put internal-session into DI system

BobrImperator:
remove setup-session-service initializer
add computed property to internal session to lazily lookup session-store.
add computed property to session service to lazily lookup session:main.
modifies tests to inject owner to entities that need to lookup dependencies.


packages/ember-simple-auth/tests/unit/services/session-test.js
```diff
@@ -147,19 +135,19 @@ describe('SessionService', () => {
     it("is read from the session's content", function() {
       session.set('some', 'data');
 
-      expect(sessionService.get('data')).to.eql({ some: 'data' });
+      expect(sessionService.get('data')).to.eql({ some: 'data', authenticated: {} });
```
BobrImperator, line 138: This is a valid change IMO since the internal-session initializes it's `content` to `{ authenticated: {} }` while the test session proxy initialized an empty object instead.

packages/ember-simple-auth/tests/unit/internal-session-test.js
```diff
@@ -808,11 +804,4 @@ describe('InternalSession', () => {
       });
     });
   });
-
-  it('does not share the content object between multiple instances', function() {
```
BobrImperator, line 812: It didn't make sense to me since there's only 1 `session:main` in the app anyway?

snewcomer (approved):
This looks great.  Thank you!

Turbo87 (approved):
looks good at first glance, but my knowledge of the ESA codebase is quite limited 😅

snewcomer:
Should we also take care of this deprecation or do that in a separate PR?


```
import { deprecate } from '@ember/application/deprecations';` has been deprecated, please update to `import { deprecate } from '@ember/debug';` [deprecation id: old-deprecate-method-paths]
```

Turbo87:
@snewcomer seems unrelated to the DI stuff, so I'd suggest to do that in a dedicated PR

patrickberkeley:
@Turbo87 or @sdebarros would you mind cutting a release that includes this please?

BobrImperator:
@patrickberkeley https://github.com/simplabs/ember-simple-auth/releases/tag/4.0.0

patrickberkeley:
🎉  Thank you @BobrImperator !

PR #2318, opened 2021-09-03
[2314] Add setup method in session service

### LINKED ISSUE #1854: Inconsistent builds with ember-cli-typescript

### LINKED ISSUE #2279: Default application route breaks embroider

### LINKED ISSUE #2314: Add #setup method to session service
Related to #2307 

During our internal talks We've decided that the way forward is to have a some kind of setup method that users will need to call to spin up the session service. Currently it's done 'automatically' via initializers, but this isn't the best solution since we're extending user code and it also leads to build problems described in #2307.

### TODO

- [ ]  implement *#setup* method on *session* service
- [ ]  copy setup code from `/initializers/setup-session-restoration`
- [ ]  Add build configuration to conditionally funnel out the `/application` route
- [ ]  Based on the configuration make it so initializer's code doesn't fire `/initializers/setup-session-restoration`

The goal API for users to use would be like this:

```
@service session;

async beforeModel() {
  await this.session.setup();
}
```
Users would explicitly call setup method without relying on initializers.

BobrImperator:
closes #2314 
closes #1854, and closes #2279

- Adds #setup method for session service
- Adds `useSessionSetupMethod` 'run time' configuration for `setup-session-restoration` intiializer to return early
- Adds deprecation when initializer is being used indicating the migration to session#setup instead.
- Documents how to use the #setup method along with the configuration options `guides/upgrate-to-v4.md`.

I was able to successfully confirm this working in my private project which conveniently is written in typescript #1854, #2279.

### DIFF
--- a/guides/upgrade-to-v4.md
+++ b/guides/upgrade-to-v4.md
@@ -1,0 +1,64 @@
+### Call session#setup method in your ApplicationRoute
+ESA implemented setup method on session service for users to manually implement in their ApplicationRoutes.
+Reason being that ESA wants to move away from relying on initializers which have been spinning up the session services for you by default.
+
+***READ THIS if you are on ESA 4.1.0 and want to opt-in into this behavior***
+In order to opt-in into this early, you'll need to add some configuration for ember-simple-auth.
+
+Inside your `ember-cli-build.js` you'll need to add `useSessionSetupMethod` flag and set it to `true`:
+```js
+  'ember-simple-auth': {
+    useSessionSetupMethod: true,
+  }
+```
+This will tell ESA to not use initializer to spin-up the session service.
+And will exclude `routes/application.js` file from Ember Simple Auth addon which might've been causing some build issues while using with typescript.
+
+With initializers gone, you'll need to call `session#setup` method directly in your application route.
+
+```js
+import Route from '@ember/routing/route';
+import { inject as service } from '@ember/service';
+
+export default class Application extends Route {
+  @service session;
+
+  async beforeModel() {
+    await this.session.setup();
+  }
+}
+```
+
+If you had any custom setup in your `beforeModel` then you'll want to move this below the session setup to preserve timings.
+
+***Old***
+```
+import Route from '@ember/routing/route';
+import { inject as service } from '@ember/service';
+
+export default class Application extends Route {
+  @service intl;
+
+  beforeModel() {
+    this.intl.setLocale('pl-PL');
+  }
+}
+```
+
+
+***New***
+```js
+import Route from '@ember/routing/route';
+import { inject as service } from '@ember/service';
+
+export default class Application extends Route {
+  @service session;
+  @service intl;
+
+  async beforeModel() {
+    await this.session.setup();
+    this.intl.setLocale('pl-PL');
+  }
+}
+```
+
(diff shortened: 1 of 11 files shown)


packages/ember-simple-auth/addon/services/session.js
```diff
@@ -368,5 +368,29 @@ export default Service.extend(Evented, {
...
+
+    This method must be called for ESA to work properly,
+    previously it was done automatically inside initilizers.
+
+    @method setup
+    @param {callback} callback to run after session restoration,
+    usually the contents of beforeModel should go here.
+    @public
+  */
+  setup(callback) {
```
BobrImperator, line 384: It feels like this method doesn't need to receive a callback, but something like this should do instead:
```js
async beforeModel(transition) {
  await this.session.restore();
  ...
}
```
But I'm not sure if this is something we'd want as it's expected for `restore` to reject in case an authenticator does so, so it'd mean that users would have to actually wrap the `await this.session.restore()` inside `try catch` which might not make sense.

guides/upgrade-to-v4.md
```diff
@@ -0,0 +1,70 @@
...
+With initializers gone, you'll need to call `session#setup` method directly in your application route.
+
+```js
+import Route from '@ember/routing/route';
+import { inject as service } from '@ember/service';
+
+export default class Application extends Route {
+  @service session;
+
+  async beforeModel(transition) {
```
marcoow, line 30:
```suggestion
  async beforeModel() {
```

…otherwise when you copy/paste this you get a linter error.

packages/ember-simple-auth/addon/initializers/setup-session-restoration.js
```diff
@@ -1,6 +1,18 @@
 import { getOwner } from '@ember/application';
+import Configuration from '../configuration';
+import { deprecate } from '@ember/application/deprecations';
 
 export default function setupSessionRestoration(registry) {
+  if (Configuration.useSessionSetupMethod) {
+    // return early in case users chose to opt out of initializer behavior.
+    return;
+  }
```
marcoow, line 9: Can we strip the initializer from the build in this case?

marcoow (commented):
This is great! I think we should do a few more things though:

* in all public methods of the `session` service, we should assert that `setup` has been called so that people that forget to do it get better feedback
* use this new option in the test app
* strip the initializer from the app build completely if the new option is set

packages/ember-simple-auth/addon/initializers/setup-session-restoration.js
BobrImperator: Not really since it's imported in other module https://github.com/simplabs/ember-simple-auth/blob/master/packages/ember-simple-auth/app/initializers/ember-simple-auth.js so we can't just remove it from build.

marcoow: ah, right 👍

bertdeblock:
This PR also introduces 3 new deprecations (aside from the intended `setup` method one):
```
`import { deprecate } from '@ember/application/deprecations';` has been deprecated, please update to `import { deprecate } from '@ember/debug';`
```

```
When calling `deprecate` you must provide `for` in options. Missing options.for in "ember-simple-auth.initializer.setup-session-restoration" deprecation
```

```
When calling `deprecate` you must provide `since` in options. Missing options.since in "ember-simple-auth.initializer.setup-session-restoration" deprecation
```

steveszc:
Should there be a link to the upgrade guide in the README? 
Right now the README does not mention anything about the session's setup method...

BobrImperator:
@steveszc Thanks, will try to work out better documentation for this

PR #2324, opened 2021-09-21
Update cookie.js

dacoho:
Fix protocol expected value in _secureCookies - fastboot add ':' at the end (https://github.com/ember-fastboot/ember-cli-fastboot/blob/2a6ed5d76b00da2106a9ba10f048aa2bef989a80/packages/fastboot/src/fastboot-request.js#L10)


marcoow (commented):
Hm, this is surprising but seems correct (https://github.com/ember-fastboot/ember-cli-fastboot/blob/8be7162fbf4a41ae48ad8c0688a8d55cbabd8579/README.md#protocol). Do you know why FastBoot would add the `:` though?

dacoho:
I really don't know. It seems that its purpose is to make it easier to build urls with the value. Maybe @xg-wang wil answer this question?

xg-wang:
It seems including `:` is standard https://developer.mozilla.org/en-US/docs/Web/API/URL/protocol
> The protocol property of the URL interface is a USVString representing the protocol scheme of the URL, including the final ':'.

PR #2330, opened 2021-10-05
2316 refactor mocha tests to qunit

candunaj:
Removed mocha. Added Qunit. Refactored unit tests.


packages/ember-simple-auth/tests/index.html
```diff
@@ -1,33 +1,43 @@
...
-    <script src="{{rootURL}}assets/vendor.js"></script>
-    <script src="{{rootURL}}assets/test-support.js"></script>
-    <script src="{{rootURL}}assets/dummy.js"></script>
-    <script src="{{rootURL}}assets/tests.js"></script>
-
-    {{content-for "body-footer"}}
-    {{content-for "test-body-footer"}}
-  </body>
-</html>
+
```
BobrImperator, line 3: are changes in this file necessary?

packages/ember-simple-auth/package.json
```diff
@@ -31,11 +31,13 @@
...
   "devDependencies": {
     "@ember/optional-features": "^1.3.0",
+    "@ember/test-helpers": "^2.4.2",
+    "@embroider/test-setup": "^0.43.5",
     "@glimmer/component": "^1.0.0",
     "@glimmer/tracking": "^1.0.0",
     "babel-eslint": "^10.1.0",
     "broccoli-asset-rev": "^3.0.0",
-    "chai": "^4.1.0",
+    "chai": "^4.3.4",
```
BobrImperator, line 40: chai is a part of mocha, so we want to remove it as well.

BobrImperator (changes requested):

packages/ember-simple-auth/package.json
```diff
@@ -31,11 +31,13 @@
   },
   "devDependencies": {
     "@ember/optional-features": "^1.3.0",
+    "@ember/test-helpers": "^2.4.2",
+    "@embroider/test-setup": "^0.43.5",
```
BobrImperator, line 35: Embroider test setup is not needed for the migration

packages/ember-simple-auth/package.json
```diff
@@ -31,11 +31,13 @@
...
     "@glimmer/component": "^1.0.0",
     "@glimmer/tracking": "^1.0.0",
     "babel-eslint": "^10.1.0",
     "broccoli-asset-rev": "^3.0.0",
-    "chai": "^4.1.0",
+    "chai": "^4.3.4",
     "ember-cli": "~3.18.0",
     "ember-cli-blueprint-test-helpers": "^0.19.2",
     "ember-cli-chai": "^0.5.0",
```
BobrImperator, line 43: chai ✂️

packages/ember-simple-auth/tests/index.html
candunaj: Change in this file fixed route /tests when you run it with "ember s", so I believe that the change in this file is valid.

BobrImperator: Now I see, qunit needs some element it can hook into, so we should probably just go back to the old `index.html` file and add this below test-body block:

```html
  <div id="qunit"></div>
  <div id="qunit-fixture">
    <div id="ember-testing-container">
      <div id="ember-testing"></div>
    </div>
  </div>
  ```

File: .github/workflows/ci.yml [hunk 5]
```diff
@@ -56,8 +56,6 @@ jobs:
           - classic-test-app
           - test-app
         test-suite:
-          - test:one ember-3.0
-          - test:one ember-lts-3.4
```
marcoow, line 60: Why does this change? Can we not support 3.0 and 3.4 when using QUnit?

packages/classic-test-app/package.json
```diff
@@ -70,4 +68,4 @@
     "cssstyle": "~1.2.2",
     "jsdom": "^11.10.0"
   }
-}
+}
```
marcoow, line 71:
```suggestion
}

```

marcoow: this seems unrelated anyway

packages/classic-test-app/tests/.eslintrc.js
```diff
@@ -2,8 +2,6 @@
 
 module.exports = {
   extends: [
-    'simplabs/configs/ember-mocha',
-    'simplabs/plugins/mocha',
```
marcoow, line 6: Should this not be changed to `"plugin:qunit/recommended"`?

packages/ember-simple-auth/.eslintrc.js
```diff
@@ -57,6 +57,11 @@ module.exports = {
         node: true,
         mocha: true
       }
-    }
+    },
+    {
+      // test files
+      files: ['tests/**/*-test.{js,ts}'],
+      extends: ['plugin:qunit/recommended'],
+    },
```
marcoow, line 65: See above – I assume this should go into the config that previously contained the Mocha linting settings?

packages/ember-simple-auth/addon/services/session.js
```diff
@@ -1,4 +1,4 @@
-import { alias, oneWay } from '@ember/object/computed';
+import { alias, readOnly } from '@ember/object/computed';
```
marcoow, line 1: This seems unrelated?

packages/ember-simple-auth/package.json
```diff
@@ -94,4 +93,4 @@
   "peerDependencies": {
     "ember-fetch": "^8.0.1"
   }
-}
+}
```
marcoow, line 96: This seems unrelated – can you revert?

packages/ember-simple-auth/tests/.eslintrc.js
```diff
@@ -2,8 +2,6 @@
 
 module.exports = {
   extends: [
-    'simplabs/configs/ember-mocha',
-    'simplabs/plugins/mocha',
```
marcoow, line 6: see aboev

packages/ember-simple-auth/tests/unit/authenticators/oauth2-password-grant-test.js
```diff
@@ -315,108 +327,123 @@ describe('OAuth2PasswordGrantAuthenticator', () => {
...
+          assert.expect(1);
           server.post('/revoke', (request) => {
             let { requestBody } = request;
             let body = parsePostData(requestBody);
 
-            expect(body).to.eql({
-              'token_type_hint': 'refresh_token',
-              'token': 'refresh token!'
-            });
+            if (body.token_type_hint === 'refresh_token') {
```
marcoow, line 353: Why is this necessary? It ends up changing the test as well?

packages/ember-simple-auth/tests/unit/authenticators/oauth2-password-grant-test.js
```diff
@@ -315,108 +327,123 @@ describe('OAuth2PasswordGrantAuthenticator', () => {
...
+      assert.expect(1);
       server.post('/token', (request) => {
         let { requestBody } = request;
         let body = parsePostData(requestBody);
 
-        expect(body).to.eql({
-          'grant_type': 'refresh_token',
-          'refresh_token': 'refresh token!'
-        });
+        if (body.grant_type === 'refresh_token') {
```
marcoow, line 397: see above – why is this necessary?

packages/ember-simple-auth/tests/unit/internal-session-store-test.js
```diff
@@ -1,40 +1,39 @@
...
 
-  describe('session store injection', function() {
-    afterEach(function() {
+  module('session store injection', function(hooks) {
+    hooks.afterEach(function() {
       Ember.testing = true; // eslint-disable-line ember/no-ember-testing-in-module-scope
     });
 
-    it('looks up the test session store when Ember.testing true', function() {
+    test('looks up the test session store when ENV.environment true', function(assert) {
```
marcoow, line 25:
```suggestion
    test('looks up the test session store when Ember.testing true', function(assert) {
```

Not sure why this changes?

marcoow: This is only one example case of course – there are plenty of other similar changes as well?

packages/ember-simple-auth/tests/unit/internal-session-test.js
```diff
@@ -23,307 +22,343 @@ describe('InternalSession', () => {
...
       });
 
-      it('is not authenticated', async function() {
-        await session.restore();
+      test('is not authenticated', async function(assert) {
+        try {
+          await session.restore();
+        } catch (e) {
+          // ignore
+        }
```
marcoow, line 243: Can we remove the `try`/`catch`? I assume we want this to raise if `restore` raises an error?

packages/ember-simple-auth/tests/unit/internal-session-test.js
```diff
@@ -23,307 +22,343 @@ describe('InternalSession', () => {
...
```
marcoow, line 259: see above – not sure why we're hiding the error?

packages/ember-simple-auth/tests/unit/internal-session-test.js
```diff
@@ -332,471 +367,533 @@ describe('InternalSession', () => {
...
       });
 
-      it('is not authenticated', async function() {
-        await session.authenticate('authenticator:test');
+      test('is not authenticated', async function(assert) {
+        try {
+          await session.authenticate('authenticator:test');
+        } catch (e) {
+          // ignore
+        }
```
marcoow, line 398: see above

packages/ember-simple-auth/tests/unit/internal-session-test.js
```diff
@@ -332,471 +367,533 @@ describe('InternalSession', () => {
...
+      test('does not process the event', async function(assert) {
         sinon.spy(authenticator, 'restore');
         let restoration = session.restore();
         store.trigger('sessionDataUpdated', { some: 'other property', authenticated: { authenticator: 'authenticator:test' } });
-        await restoration;
+        try {
+          await restoration;
+        } catch (e) {
+          // ignore
+        }
```
marcoow, line 619: see above

packages/ember-simple-auth/tests/unit/session-stores/shared/storage-event-handler-behavior.js
```diff
@@ -1,52 +1,51 @@
+
```
marcoow, line 1:
```suggestion
```

packages/test-app/package.json
```diff
@@ -73,4 +71,4 @@
     "cssstyle": "~1.2.2",
     "jsdom": "^11.10.0"
   }
-}
+}
```
marcoow, line 74:
```suggestion
}

```

packages/test-app/config/fastboot-testing.js
```diff
@@ -1,4 +1,10 @@
-const najax = require('najax');
+let najax;
+try {
+  // eslint-disable-next-line node/no-missing-require
+  najax = require('najax');
+} catch (err) {
+  // ignore the error
```
marcoow, line 6: Why is this error ignored?

packages/test-app/tests/.eslintrc.js
```diff
@@ -2,8 +2,6 @@
 
 module.exports = {
   extends: [
-    'simplabs/configs/ember-mocha',
-    'simplabs/plugins/mocha',
```
marcoow, line 6: see above – I think this is where the qunit config should go

packages/test-app/tests/index.html
```diff
@@ -1,33 +1,43 @@
...
+  <script src="{{rootURL}}assets/vendor.js"></script>
+  <script src="{{rootURL}}assets/test-support.js"></script>
+  <script src="{{rootURL}}assets/test-app.js"></script>
+  <script src="{{rootURL}}assets/tests.js"></script>
+
+  {{content-for "body-footer"}}
+  {{content-for "test-body-footer"}}
+</body>
+
+</html>
```
marcoow, line 43: Seems like this diff of this file is a bit messed up because of the reformatting (and now missing indentation)?

marcoow (changes requested):
Very nice work :raised_hands:

I found some things that I think still should be addressed but I'm looking forward getting this merged!

One general thing I recognized is that this replaces the nested `describe` blocks with a flat structure with just one `module` and `test`s nested directly into that (at least in some of the test files).Can we use nested `module`s instead? I feel a flat list of tests is much harder to navigate (plus, in a nested structure it's much easier to build up preconditions for tests in nested blocks incrementally).

.github/workflows/ci.yml
candunaj: Sorry, I wanted to write ember-qunit 5. ember-qunit 5 supports ember 3.8+. 
https://github.com/emberjs/ember-qunit

packages/ember-simple-auth/addon/services/session.js
candunaj: There are 2 unit tests which expect three properties in session service to be readOnly and fail when set. So I have changed properties isAuthenticated, data and store to readOnly.

marcoow: But `oneWay` should be the same? My point is migrating the tests from Mocha to Qunit shouldn't require any changes in the tested code?

.github/workflows/ci.yml
marcoow: Can we use a version of ember-qunit that supports Ember 3.0 as well? I'd rather upgrade to ember-qunit 5 (and drop support for Ember < 3.8) in a separate PR then.

candunaj: Yes, I can use ember-qunit 4 with some change in the source code. I will do it.

BobrImperator:
I went through the changes and there isn't anything from my side apart from what Marco's pointed out.
Although, I'm not sure if we were supposed to change `test` and `classic-test`apps as well, thought that just ESA package refactor is enough.

packages/ember-simple-auth/addon/services/session.js
candunaj: oneWay will not throw an error when set. I have found the bug in mocha tests so the bug was always there.

```
it('is read-only', function() {
      expect(() => {
        sessionService.set('store', 'some other store');
      }).to.throw;
    });
```
it should be
```
}).to.throw();
```
Change is authenticated from oneWay to readOnly has fixed 3 tests, but it can be considered as breaking change. So maybe we should remove these tests?

marcoow: Hm, we should definitely move that into a separate PR to not also mix actual changes to the behavior into this – it's already a huge PR (necessarily) so I'd keep it as limited as possible :)

packages/classic-test-app/package.json
```diff
@@ -14,24 +14,22 @@
     "test:fastboot": "ember fastboot:test"
   },
   "devDependencies": {
+    "@ember/jquery": "^2.0.0",
```
marcoow, line 17: Is this related?

packages/ember-simple-auth/tests/unit/authenticators/oauth2-password-grant-test.js
marcoow: @candunaj I assume the `if` could just be dropped?

marcoow: @candunaj didn't you extract this into a separate PR?

marcoow (approved):
great job 🎉

BobrImperator (approved):

PR #2334, opened 2021-10-29
ember-release must pass CI

candunaj:
ember-release must pass CI


candunaj:
In the current version, ember-release must pass, so PR is not needed.

PR #2335, opened 2021-10-30
Fix deprecations related to `deprecate` util

bertdeblock:
Fixes the new deprecations introduced in #2318.

### DIFF
--- a/packages/ember-simple-auth/addon/initializers/setup-session-restoration.js
+++ b/packages/ember-simple-auth/addon/initializers/setup-session-restoration.js
@@ -1,6 +1,6 @@
 import { getOwner } from '@ember/application';
 import Configuration from '../configuration';
-import { deprecate } from '@ember/application/deprecations';
+import { deprecate } from '@ember/debug';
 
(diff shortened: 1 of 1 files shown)


BobrImperator (approved):
Thanks for this :)

snewcomer (approved):

lougreenwood:
Can we get this or one of the other 2 PRs which fix this merged please?

https://github.com/simplabs/ember-simple-auth/pull/2335
https://github.com/simplabs/ember-simple-auth/pull/2347

BobrImperator:
Sorry for the delay on this 🥲 

@bertdeblock do you think we still need the changes?
Is there any advantage to having the `since.available` property defined?

bertdeblock:
TBH, I'm not _super_ familiar with the deprecation staging RFC and how it should be applied within addons. I'm guessing the deprecation warnings are gone now?

BobrImperator:
I think they are, and probably the available property is not needed anymore.
Should we close this?

PR #2336, opened 2021-10-30
Always assert the looked up authenticator

bertdeblock:


packages/ember-simple-auth/addon/internal-session.js
```diff
@@ -203,6 +202,7 @@ export default ObjectProxy.extend(Evented, {
   _lookupAuthenticator(authenticatorName) {
     let owner = getOwner(this);
     let authenticator = owner.lookup(authenticatorName);
+    assert(`No authenticator for factory "${authenticatorName}" could be found!`, !isNone(authenticator));
     setOwner(authenticator, owner);
```
bertdeblock, line 206: This line throws with a vague error when `authenticator` is `undefined`. Moving the assertion here makes it clearer as to what went wrong.

BobrImperator: Sorry for late response.

Can I ask you for a regression test for this? It seems like tests don't anticipate undefined input.

PR #2349, opened 2021-12-02
Fixed failing unit test - invalidate call revoke endpoint twice. Unit test did not validate it correctly.

candunaj:
## Issue
The unit test was testing the authenticator.invalidate method. It is calling revoke endpoint twice. One time revoke refresh_token and the second time revoke access_token. Unit test expected only one request for refresh_token. Call for revoke access_token was not expected and was failing quietly.

## Fix
When the fake revoke endpoint is called, only the refresh token is validated.

### DIFF
--- a/packages/ember-simple-auth/tests/unit/authenticators/oauth2-password-grant-test.js
+++ b/packages/ember-simple-auth/tests/unit/authenticators/oauth2-password-grant-test.js
@@ -333,17 +333,23 @@ describe('OAuth2PasswordGrantAuthenticator', () => {
 
       describe('when a refresh token is set', function() {
         it('sends an AJAX request to invalidate the refresh token', async function() {
+          let refreshTokenValidated = false;
           server.post('/revoke', (request) => {
             let { requestBody } = request;
             let body = parsePostData(requestBody);
 
-            expect(body).to.eql({
-              'token_type_hint': 'refresh_token',
-              'token': 'refresh token!'
-            });
+            if (body.token_type_hint === 'refresh_token') {
+              expect(body).to.eql({
+                'token_type_hint': 'refresh_token',
+                'token': 'refresh token!'
+              });
+              refreshTokenValidated = true;
+            }
           });
 
           await authenticator.invalidate({ 'access_token': 'access token!', 'refresh_token': 'refresh token!' });
+
+          expect(refreshTokenValidated).to.equal(true);
         });
       });
     });


marcoow:
Isn't this working correctly on `master` though, e.g. https://github.com/simplabs/ember-simple-auth/actions/runs/1530141659 ?

candunaj:
It is failing on master also. Revoke endpoint is called twice and it fails on the first request and succeeds on the second. It is not caught by mocha, because it is not correctly implemented. I can prove it by debugging.

packages/ember-simple-auth/tests/unit/authenticators/oauth2-password-grant-test.js
```diff
@@ -333,17 +333,23 @@ describe('OAuth2PasswordGrantAuthenticator', () => {
...
           await authenticator.invalidate({ 'access_token': 'access token!', 'refresh_token': 'refresh token!' });
+          expect(revokeRequests.length).to.eq(2);
+          expect(revokeRequests.find(body => body.token_type_hint === 'refresh_token')).to.eql({
+            'token_type_hint': 'refresh_token',
+            'token': 'refresh token!'
+          });
+          expect(revokeRequests.find(body => body.token_type_hint === 'access_token')).to.eql({
+            'token_type_hint': 'access_token',
+            'token': 'access token!'
+          });
```
marcoow, line 352: Given the messages of the `describe` and `test` this is in, I would assume there's a dedicated test for asserting the access token is invalidated?

packages/ember-simple-auth/tests/unit/authenticators/oauth2-password-grant-test.js
```diff
@@ -333,17 +333,23 @@ describe('OAuth2PasswordGrantAuthenticator', () => {
...
-            });
+            revokeRequests.push(body);
           });
 
           await authenticator.invalidate({ 'access_token': 'access token!', 'refresh_token': 'refresh token!' });
+          expect(revokeRequests.length).to.eq(2);
+          expect(revokeRequests.find(body => body.token_type_hint === 'refresh_token')).to.eql({
+            'token_type_hint': 'refresh_token',
+            'token': 'refresh token!'
+          });
```
marcoow, line 348: I assume it would be simpler to just wrap the [`expect` in line 340](https://github.com/simplabs/ember-simple-auth/pull/2349/files#diff-eb0b12906e6123e788d141897aca52bfb626458de79e71dec56d390561510bc0L340) in an `if body.token_type_hint === 'refresh_token'`.

Sorry, I think that's what you had originally but I didn't fully understand the issue before…

BobrImperator (approved):

marcoow (approved):

PR #2350, opened 2021-12-09
Upgrade to qunit 5

candunaj:
Currently ember-simple-auth is using ember-qunit 4 which does not support ember 4. 
I have upgraded ember-simple-auth to ember-qunit 5. ember-qunit 5 does not support ember 3.0 so I have removed it from ci.


packages/ember-simple-auth/config/ember-try.js
```diff
@@ -14,21 +14,6 @@ module.exports = function() {
     return {
       useYarn: true,
       scenarios: [
-        {
-          name: 'ember-3.0',
```
BobrImperator, line 18: I believe we already spoke about this, but just to make sure - we have to drop the ember 3.0 in order to use this qunit version?

Qunit 5 is the one compatible with embroider, right?

candunaj: ember-qunit 4 does not support ember 4. I have upgraded ember-qunit to 5 to be able to support ember-4.

BobrImperator: Got it 🚀

marcoow:
This can only go out with 5.0.0 so we need to put this on hold for now…

marcoow:
@candunaj I'd like to get something like #2353 working first tbh

BobrImperator (approved):

PR #2353, opened 2021-12-10
Fix ember-release + ember-beta build

marcoow:
This adds testing for Ember 4 on CI. This is focused on making things work to unblock us but we should clean up afterwards. There's probably a number of things we can clean up etc. but we should first have coverage for Ember 4 on CI.

closes #2332


marcoow:
@BobrImperator do you have any idea why the [tests fail for the engine in `test-app` for Ember 4.0](https://github.com/simplabs/ember-simple-auth/runs/6613444959?check_suite_focus=true) but [work for the classic-test-app with Ember 4.0](https://github.com/simplabs/ember-simple-auth/runs/6613444378?check_suite_focus=true)?

I can't seem to spot the difference…

BobrImperator:
I believe the problem might be embroider macros acting up, this happens because of a condition inside ember-engines that conditionally registers a custom link-to for versions older than 3.24.1, for some reason in the test-app it's detecting older versions while in the classic it seems to be correct.

image

marcoow:
Awesome 🎉 I also added the Ember 4.0 scenario back in – can't hurt to test that as well

BobrImperator (approved):

PR #2355, opened 2022-01-12
Remove use of `keys` and `merge` utils coming from `@ember/polyfills`

bertdeblock:
- `keys` does not exist anymore
- `merge` was deprecated in v3 and removed in v4

### DIFF
--- a/packages/ember-simple-auth/tests/helpers/start-app.js
+++ b/packages/ember-simple-auth/tests/helpers/start-app.js
@@ -1,12 +1,11 @@
 import Application from '../../app';
 import config from '../../config/environment';
-import { merge } from '@ember/polyfills';
 import { run } from '@ember/runloop';
 
 export default function startApp(attrs) {
-  let attributes = merge({}, config.APP);
+  let attributes = Object.assign({}, config.APP);
   attributes.autoboot = true;
-  attributes = merge(attributes, attrs); // use defaults, but you can override;
+  attributes = Object.assign(attributes, attrs); // use defaults, but you can override;
 
   return run(() => {
     let application = Application.create(attributes);
--- a/packages/ember-simple-auth/tests/helpers/start-app.js
+++ b/packages/ember-simple-auth/tests/helpers/start-app.js
@@ -1,11 +1,12 @@
 import Application from '../../app';
 import config from '../../config/environment';
 import { run } from '@ember/runloop';
+import assign from 'ember-simple-auth/utils/assign';
 
 export default function startApp(attrs) {
-  let attributes = Object.assign({}, config.APP);
+  let attributes = assign({}, config.APP);
   attributes.autoboot = true;
-  attributes = Object.assign(attributes, attrs); // use defaults, but you can override;
+  attributes = assign(attributes, attrs); // use defaults, but you can override;
 
   return run(() => {
     let application = Application.create(attributes);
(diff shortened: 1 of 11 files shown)


marcoow (commented):
Look great in general – thanks for the effort! Just one thing I think we should change: as long as ESA supports Ember 3.x we should also support IE11 (as Ember 3.x does). Since [IE11 doesn't have `Object.assign`](https://caniuse.com/mdn-javascript_builtins_object_assign), we should still be getting `assign` from the polyfill.

In that sense, the fact that ESA already uses `Object.assign` in one place should be considered a but that should be fixed.

bertdeblock:
@marcoow I dropped the commit that removed the use of `assign`.

packages/ember-simple-auth/tests/helpers/start-app.js
```diff
@@ -1,12 +1,11 @@
 import Application from '../../app';
 import config from '../../config/environment';
-import { merge } from '@ember/polyfills';
 import { run } from '@ember/runloop';
 
 export default function startApp(attrs) {
-  let attributes = merge({}, config.APP);
+  let attributes = Object.assign({}, config.APP);
```
marcoow, line 6: Can we use `assign` from `@ember/polyfills` here as well?

packages/ember-simple-auth/tests/helpers/start-app.js
```diff
@@ -1,12 +1,11 @@
...
   attributes.autoboot = true;
-  attributes = merge(attributes, attrs); // use defaults, but you can override;
+  attributes = Object.assign(attributes, attrs); // use defaults, but you can override;
```
marcoow, line 8: Can we use `assign` from `@ember/polyfills` here as well?

packages/ember-simple-auth/tests/helpers/start-app.js
bertdeblock: Is that needed in test land? I see that `Object.assign` is used in other test files as well:
https://github.com/simplabs/ember-simple-auth/search?q=Object.assign

marcoow: Yeah that was my earlier point – we should actually never just use it in Ember code (`.eslintrc.js` etc. is fine). The current usages should be considered a bug really.

Windvis: @marcoow Isn't it an option to let consumers bring an actual `Object.assign` polyfill? I would guess they are using that already anyways if they still support IE11 in 2022.

Using the ember assign polyfill means that 4.0 apps will show that deprecation warning until ember-simple-auth officially drops 3.x support, which doesn't seem ideal.

Alternatively, I think it should be possible to use `@embroider/macros` and only use Ember's assign if the consuming app is on a 3.x release.

marcoow: Sure but it we're already cleaning things up (and will be using `assign` from `@ember/polyfill` in other places still anyway) we might just as well update the existing usages?

> Using the ember assign polyfill means that 4.0 apps will show that deprecation warning until ember-simple-auth officially drops 3.x support, which doesn't seem ideal.

I don't think that's a problem tbh and we won't drop support for 3.x for quite some time. Aiming for 0 deprecations at all times isn't something anyone should be doing…

bertdeblock: We could also do something like this:

```js
const assign = Object.assign || emberAssign;
```

So similar to what was done before when falling back to `merge`.
This won't trigger the deprecation in Ember v4.

marcoow: I'd maybe have a util module in ESA for that so we don't have to repeat that everywhere in the app 👍

marcoow (approved):
🙌

BobrImperator:
Sorry for the delay - I just released the change, thought we did so already 🙏

Issue #2356, opened 2022-01-20
Ember 4.1.0: Use of `assign` has been deprecated. Please use `Object.assign` or the spread operator instead.


xomaczar:
Internal session uses assign from `@ember/polyfills` which has been deprecated since version 4.0

See: [Code Reference](https://github.com/simplabs/ember-simple-auth/blob/bca31c8055840ce5537c2d910482200d18070018/packages/ember-simple-auth/addon/internal-session.js?_pjax=%23js-repo-pjax-container%2C%20div%5Bitemtype%3D%22http%3A%2F%2Fschema.org%2FSoftwareSourceCode%22%5D%20main%2C%20%5Bdata-pjax-container%5D#L6)

If there are no arguments against it I can PR Object.assign replacement to remove this deprecation.

BobrImperator:
closed by #2355

PR #2357, opened 2022-01-20
Replace @ember/polyfills assign with Object.assign

xomaczar:
Closes https://github.com/simplabs/ember-simple-auth/issues/2356

### DIFF
--- a/packages/ember-simple-auth/addon/authenticators/devise.js
+++ b/packages/ember-simple-auth/addon/authenticators/devise.js
@@ -1,11 +1,11 @@
 import { Promise } from 'rsvp';
 import { isEmpty } from '@ember/utils';
 import { run } from '@ember/runloop';
-import { merge, assign as emberAssign } from '@ember/polyfills';
+import { merge } from '@ember/polyfills';
 import BaseAuthenticator from './base';
 import fetch from 'fetch';
 
-const assign = emberAssign || merge;
+const assign = Object.assign || merge;
 
 const JSON_CONTENT_TYPE = 'application/json';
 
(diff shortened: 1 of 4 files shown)


packages/ember-simple-auth/addon/authenticators/devise.js
```diff
@@ -1,11 +1,11 @@
 import { Promise } from 'rsvp';
 import { isEmpty } from '@ember/utils';
 import { run } from '@ember/runloop';
-import { merge, assign as emberAssign } from '@ember/polyfills';
+import { merge } from '@ember/polyfills';
```
marcoow, line 4: Isn't `merge` the one that was [removed from `@ember/polyfills` in 4.0](https://api.emberjs.com/ember/4.0/modules/@ember%2Fpolyfills)?

packages/ember-simple-auth/addon/authenticators/torii.js
```diff
@@ -73,7 +72,7 @@ export default BaseAuthenticator.extend({
       return this.get('torii').fetch(data.provider, data).then(
         (fetchedData) => {
           this._authenticateWithProvider(provider, fetchedData);
-          return emberAssign(data, fetchedData);
+          return Object.assign(data, fetchedData);
```
marcoow, line 75: This won't work in IE11 (which I think ESA still supports as long as we support Ember 3.x).

marcoow (commented):
Thanks for the effort – I'm not sure this is quite right yet though tbh…

Windvis:
This seems like a duplicate of #2355?

packages/ember-simple-auth/addon/authenticators/torii.js
bertdeblock: Note that ESA already uses Object.assign here:
https://github.com/simplabs/ember-simple-auth/blob/master/packages/ember-simple-auth/addon/session-stores/cookie.js#L183

marcoow: Ouch – thanks for pointing that out!

marcoow:
indeed, it's a duplicate of #2355 – closing

PR #2359, opened 2022-01-25
Always assert the looked up authenticator

bertdeblock:
Redo of #2336.

Instead of:

```
TypeError: Cannot set properties of undefined (setting '__OWNER__ember16431396598451637272055234__')
```

This will throw:

```
Error: Assertion Failed: No authenticator for factory "authenticator:foo" could be found!
```

Which should be clearer to the user what went wrong. This will be the case everywhere were `_lookupAuthenticator` is called.


marcoow (approved):
Nice 👍

PR #2366, opened 2022-04-01
[2365] internal-session to call _initialize on session-store during init

### LINKED ISSUE #2365: CookieStore is not initializing/syncing correctly
In version `4.0.0`, a new method `_initialize()` was added to help instantiate the cookie store from the adaptive store. However, if you are explicitly extending the cookie store, this method is never called, therefore no syncing is enabled. This prevents `handleAuthentication`/`handleInvalidation` from being called across tabs/windows

BobrImperator:
closes #2365

- Changes internal-session to call `_initialize` after it looks up a session-store.


BobrImperator:
@marcoow No, because adaptive store needs to pass options before that happens, otherwise session-store:cookie could run with incorrect arguments.
I don't really see much we can do without some refactoring. 

The code that caused it:
https://github.com/simplabs/ember-simple-auth/pull/2312/files#diff-db3487bdb1f8c84714dedc96d1cce936b66f804e8a3c58ebe9d88dfb883e50deR137

marcoow:
Can the adaptive store not set the options after getting the cookie store from the container so that the cookie store would init itself as it would do when used standalone in `init` but the adaptive store simply overrides those settings afterwards?

BobrImperator:
Indeed it seems like it does the trick and no tests are failing - though I swear that was the case.
I dunno what I saw before 😅

PR #2370, opened 2022-04-29
ci(renovate): add renovate configuration

oscard0m:
using simplabs' preset

PR #2379, opened 2022-05-02
disable dependabot

marcoow:
…since we're now using Renovate (see #2370)


oscard0m (approved):

oscard0m:
When the list of repos is long there is this script 😎 : https://github.com/octoherd/script-remove-dependabot

PR #2419, opened 2022-05-26
Run CI on Node 14

marcoow:
Node 12 reached EOL


marcoow:
@BobrImperator merging this to unblock myself 🙀

PR #2429, opened 2022-06-30
Update the oauth2 password grant to support client_secret

TikiTDO:
Earlier version of ESA supported passing the `client_id` and the `client_secret` values [in the header](https://github.com/simplabs/ember-simple-auth/blob/1815e03a6ed3c5f05f27153a2f1c927014db1ffc/addon/authenticators/oauth2-password-grant.js#L130). More recently this feature was changed to pass `client_id` in the request body, but in the process support for passing in a `client_secret` was removed entirely.

The OAuth server I use does not support generating a `client_id` without a `client_secret`, so this breaks my login flow unless I implement a custom client parser.

[The spec](https://datatracker.ietf.org/doc/html/rfc6749#section-2.3.1) suggests that the recommended way to accomplish this would be to pass both in the `Authorization` header as it was done before, but in the absence of the header it would be great to have the ability to pass in a `client_secret` value directly. 

Alternatively, the spec dictates that the server must accept the `HTTP Basic authentication scheme`, so it may be easier to simply add a `clientHeader` value instead.

### DIFF
--- a/packages/ember-simple-auth/addon/authenticators/oauth2-password-grant.js
+++ b/packages/ember-simple-auth/addon/authenticators/oauth2-password-grant.js
@@ -46,6 +46,19 @@ export default BaseAuthenticator.extend({
     @public
   */
   clientId: null,
+  
+    /**
+    The client_secret to be sent to the authentication server (see
+    https://tools.ietf.org/html/rfc6749#appendix-A.2). __This should only be
+    used for statistics or logging etc. as it cannot actually be trusted since
+    it could have been manipulated on the client!__
+
+    @property clientSecret
+    @type String
+    @default null
+    @public
+  */
+  clientSecret: null,
 
   /**
     The endpoint on the server that authentication and token refresh requests
@@ -290,6 +303,11 @@ export default BaseAuthenticator.extend({
     const clientId = this.get('clientId');
     if (!isEmpty(clientId)) {
       data['client_id'] = this.get('clientId');
+    }
+    
+    const clientSecret = this.get('clientSecret');
+    if (!isEmpty(clientSecret)) {
+      data['client_secret'] = this.get('clientSecret');
     }
 
     const body = Object.keys(data).map((key) => {


marcoow:
Thanks for the effort but ESA is not going to support client secrets.

This has been discussed several times before and the point is: if you put a secret in your JS app it's no longer a secret. If your server demands a secret, you'll need to use a proxy server that adds it into requests coming in from the web app (even then the secret is pointless though since anyone could call the proxy of course).

TikiTDO:
@marcoow So in that case just call the key `theSecondPartOfClientIdWhichIsNotReallyASecretEvenThoughSomeServersMightCallItThat` to reflect what it actually does. Or perhaps update your documentation to mention that you explicitly disallow use cases mentioned in the spec, because they use a word you think some people might find confusing, ideally in a place that is not in a PR comment that arose after trying to fix an issue after hours of debugging.

Or perhaps don't call this module the `oauth2-password-grant` since you're literally ignoring parts of the spec, because you think a word might confuse the dozen or so people that still use Ember for some horrific legacy reason. That way people that use the spec to validate their code with the assumption of "of course that lib that says it's oauth2 follows the spec" don't need to dig into some deeper modules of that lib to figure out that the authors have their own take on things.

Anyway, since it's clear that anyone involved in this library would rather stick to excuses than make the environment usable... for anyone comes here stuck because you have the misfortune of still using Ember with rails doorkeeper and a desire to identify your clients, you can get around it as so in your doorkeeper config. That way you don't have to fork your own copy of this module. For other servers good luck on your own: 

```ruby
  client_credentials :from_basic, lambda { |request|
    # TODO: Remove this if https://github.com/simplabs/ember-simple-auth/pull/2429 is merged
    if request.params["client_id"].present? && request.params["client_secret"].present?
      next request.params["client_id"], request.params["client_secret"]
    elsif request.params["client_id"].present?
      potential_auth_payload = request.params["client_id"]
      potential_decoded_payload = Base64.decode64(potential_auth_payload)
      is_base64 = potential_auth_payload.is_a?(String) && Base64.strict_encode64(potential_decoded_payload) == potential_auth_payload

      if is_base64
        parsed_payload = potential_decoded_payload.split(":")
        next parsed_payload.first, parsed_payload.last
      else
        next potential_auth_payload, ""
      end
    end
  }
```

Needless to say, my real recommendation is to wipe all traces of Ember from your codebase. You'll be glad you did.

PR #2509, opened 2023-01-23
Test test-app with Embroider

marcoow:


BobrImperator:
I was under impression that `embroiderSafe` `embroiderOptimized` scenarios are for addon, not for apps.

Also engines need additional configurations https://github.com/embroider-build/embroider#lazy-engines

Apps should probably implement https://github.com/embroider-build/embroider#how-to-try-it instead of the test scenarios.

marcoow:
> I was under impression that embroiderSafe embroiderOptimized scenarios are for addon, not for apps.

I'm not sure. We only have unit tests left in the addon's dummy app so I guess we should really test the actual test apps with Embroider?

BobrImperator:
To my understanding, since embroider is a build system - the test-setup in an addon makes sure that the addon is able to build under embroider and it can do static analysis.
Whether there are integration and acceptance tests it probably shouldn't matter 🤔 

To test apps under embroider, they should probably just switch to embroider build system i.e. configure ember-cli-build and turn the static splitting features.

PR #2510, opened 2023-01-23
Fix running apps on fresh clone

marcoow:
Running the apps from a fresh clone of the repo doesn't work currently. This fixes that by

* upgrading to the latest torii
* upgrading the `ember-engines` dependency to the latest version
* adding the `@ember/legacy-built-in-components` dependency which is necessary for `ember-engines` to work
* upgrading the ember-data dependency to 4.4

### DIFF
--- a/packages/classic-test-app/package.json
+++ b/packages/classic-test-app/package.json
@@ -17,6 +17,7 @@
     "test:fastboot": "ember fastboot:test"
   },
   "devDependencies": {
+    "@ember/legacy-built-in-components": "^0.4.1",
     "@ember/optional-features": "^2.0.0",
     "@ember/test-helpers": "~2.7.0",
     "@glimmer/component": "^1.0.0",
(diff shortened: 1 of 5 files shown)


marcoow:
Argh, don't know why it doesn't work 😞

BobrImperator:
ember-engines 0.9 most likely won't work with Ember <= 3.24. https://github.com/ember-engines/ember-engines/commit/31bc30929f8f17fbdadb54e22c9a6d128864133a

@ember/legacy-built-in-components is not required, is breaking for 'older' ember. And is only needed for newer ember versions as per https://github.com/ember-engines/ember-engines/pull/798

AFAIK ember torii 1.0.0 doesn't target older ember only 3.24 and newer as per https://github.com/ember-engines/ember-engines/pull/798

marcoow:
See the [changes in `ember-try.js`](https://github.com/mainmatter/ember-simple-auth/pull/2510/files#diff-8721608d58b759442a39c6ef5cb2b431698cdeb646a8b224e2db3539ffbf8bb9) – everything should be as before for older Ember versions.

I'll try again but pretty sure `yarn workspace test-app start` didn't work with a fresh clone yesterday.

marcoow:
Yep, does indeed not work:

Bildschirm­foto 2023-01-24 um 08 03 53

BobrImperator:
Just checked `yarn workspace test-app start` on a fresh clone and indeed it doesn't build.
I realized I always used ember-try e.g. `yarn workspace test-app ember try:one ember-4.0 --- ember serve` `yarn workspace test-app ember try:one ember-4.0`.

Also it seems this wasn't picked up because our CI doesn't test against `ember-default` test scenario

BobrImperator:
Fixed by #2512

BobrImperator:
Fixed by https://github.com/mainmatter/ember-simple-auth/pull/2512
This is a follow up which adds a CI job for the default scenario https://github.com/mainmatter/ember-simple-auth/pull/2515

PR #2512, opened 2023-01-30
Upgrade torii devDependency to 1.0.0-beta.1

Gaurav0:
Maintainer edit:
- adds `ember-lts-4.8` try scenario.
- bumps `torii` to the adopted 1.0.0-beta.1 https://github.com/adopted-ember-addons/torii
- bumps packages in the package.jsons for the default ember scenario.

### DIFF
--- a/packages/classic-test-app/package.json
+++ b/packages/classic-test-app/package.json
@@ -19,6 +19,7 @@
   "devDependencies": {
     "@ember/optional-features": "^2.0.0",
     "@ember/test-helpers": "~2.7.0",
+    "@ember/legacy-built-in-components": "^0.4.1",
     "@glimmer/component": "^1.0.0",
     "@glimmer/tracking": "^1.0.0",
     "babel-eslint": "^10.1.0",
@@ -39,7 +40,7 @@
     "ember-cli-terser": "4.0.2",
     "ember-data": "~4.4.0",
     "ember-disable-prototype-extensions": "^1.1.2",
-    "ember-engines": "^0.8.22",
+    "ember-engines": "^0.9.0",
     "ember-export-application-global": "^2.0.1",
     "ember-fetch": "^8.0.1",
     "ember-load-initializers": "^2.1.1",
--- a/packages/classic-test-app/package.json
+++ b/packages/classic-test-app/package.json
@@ -37,7 +37,7 @@
     "pretender": "3.4.7",
     "ember-cli-sri": "^2.1.1",
     "ember-cli-terser": "4.0.2",
-    "ember-data": "~3.17.0",
+    "ember-data": "~4.4.0",
     "ember-disable-prototype-extensions": "^1.1.2",
     "ember-engines": "^0.8.22",
     "ember-export-application-global": "^2.0.1",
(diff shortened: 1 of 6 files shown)


BobrImperator:
Thanks 👍 

My concern though is dropping the older ember support, is it necessary?
I'd also prefer to wait for official 1.0.0 release rather than installing beta if possible, this is why we haven't upped this one yet.

packages/classic-test-app/package.json
```diff
@@ -37,7 +37,7 @@
     "pretender": "3.4.7",
     "ember-cli-sri": "^2.1.1",
     "ember-cli-terser": "4.0.2",
-    "ember-data": "~3.17.0",
+    "ember-data": "~4.4.0",
```
BobrImperator, line 40: ✂️ 
Let's keep torii upgrade as the only change if possible

Gaurav0: The "default" test fails without this.

Gaurav0:
> I'd also prefer to wait for official 1.0.0 release rather than installing beta if possible, this is why we haven't upped this one yet.

Ok. I'll talk to @Halfbyte about releasing a 1.0.

Gaurav0:
> My concern though is dropping the older ember support, is it necessary?

Not for this, as torii is just a devdependency. Once I get tests working otherwise, I'll reenable testing on older versions and use the old version of torii in those ember-try scenarios.

packages/ember-simple-auth/package.json
```diff
@@ -31,6 +31,7 @@
     "silent-error": "^1.0.0"
   },
   "devDependencies": {
+    "@ember/legacy-built-in-components": "^0.4.1",
```
BobrImperator, line 34: `@ember/legacy-built-in-components` shouldn't be a part of the `ember-simple-auth` addon.
It's an optional peerDependency of `ember-engines`.

packages/ember-simple-auth/package.json
```diff
@@ -49,7 +50,7 @@
     "ember-cli-sri": "^2.1.1",
     "ember-cli-terser": "4.0.2",
     "ember-cli-yuidoc": "^0.9.1",
-    "ember-data": "~3.18.0",
+    "ember-data": "~4.4.0",
```
BobrImperator, line 53: Hmm, now that I think about it, ember-data likely can be removed from ESA.
Will do separately.

packages/ember-simple-auth/config/ember-try.js
```diff
@@ -160,17 +186,12 @@ module.exports = function() {
...
           }
         }),
         embroiderOptimized({
           npm: {
-            devDependencies: {
-              'ember-data': '~4.4.0',
-            }
+            devDependencies: {}
           }
         })
```
BobrImperator, line 196:
```
        embroiderSafe(),
        embroiderOptimized()
   ```

packages/ember-simple-auth/package.json
Gaurav0: Devdependencies are not "a part of" on addon, they are everything needed to build and test an addon. Only dependencies are "part of" an addon, as in when you install an addon, you only get the dependencies.

`@ember/legacy-built-in-components` was required to pass the default scenario. It's not optional if you are testing with ember 4+ because ember-engines uses it and we are testing ember engines in the tests.

BobrImperator: ESA addon doesn't use either `@ember/legacy-builtin-components` or the `ember-engines`, the test apps do.
I understand this isn't shipped to the users but it shouldn't be here at all since it's simply not used inside the addon itself.

Gaurav0: Ah ok, let me try that.

PR #2517, opened 2023-02-09
Update test-app js files to use modern syntax

Gaurav0:
For #2516


BobrImperator:
Nice, thanks 🙌 
Wasn't aware that test-app was secretly the classic-test-app 😄

BobrImperator (approved):

PR #2521, opened 2023-02-26
Fix the link to README.md.

canburak:
Nothing special. The link to README is broken and the fix is trivial.


BobrImperator:
Thanks!

PR #2525, opened 2023-03-23
chore: Drop node - set minimal node version to >=16

BobrImperator:
https://github.com/nodejs/release#release-schedule

- Node 12 has reached its EOL
- Node 14 will be reaching its EOL in about a month
- Node 16 will reach it later this year **our minimal supported version**


.github/workflows/release.yml
```diff
@@ -14,7 +14,7 @@ jobs:
       - uses: actions/checkout@2541b1294d2704b0964813337f33b291d3f8596b # tag=v3
       - uses: actions/setup-node@969bd2663942d722d85b6a8626225850c2f7be4b # tag=v3
         with:
-          node-version: 12.x
+          node-version: 18.x
```
marcoow, line 17:
```suggestion
          node-version: 16.x
```

Should we run CI on the **minimal** supported version instead?

BobrImperator: That makes sense.

Changed the project to run on 16.

marcoow (approved):

PR #2526, opened 2023-03-23
feat: remove deprecated Evented API usage from the public session service

BobrImperator:
- Removes deprecated Evented API usage from the public session service.
`handleAuthentication` and `handleInvalidation` methods should be used instead.

Example usage of `handleAuthentication`    https://github.com/mainmatter/ember-simple-auth/blob/master/guides/managing-current-user.md


marcoow (approved):

RobbieTheWagner:
I'm not sure what I am doing wrong, but `handleAuthentication` never gets called for me. I used to use `this.session.on('authenticationSucceeded',` which worked great, but now I have nothing that works.

BobrImperator:
@RobbieTheWagner 
The only thing I could think of is not calling the setup method https://github.com/mainmatter/ember-simple-auth/blob/master/guides/upgrade-to-v4.md#call-sessionsetup-method-in-your-applicationroute

RobbieTheWagner:
I am calling the setup method and have been for a long time, so it's not that.

BobrImperator:
Other scenario could be misplaced `handleAuthentication` method.
I remember it used to exist for an application route at some point but was removed with mixins and the service should be extended instead.

I don't see how else it could break other than the service not existing at all.

You could inspect the `classic-test-app` and `test-app` to see how to use session service now. This PR specifically that removes mixins  https://github.com/mainmatter/ember-simple-auth/pull/2480/files

The event logic was also removed from the public service only while the internal one still has them and they are actually being used for triggering the `handleAuthentication` method.
We didn't change much in terms of internal logic so far, just the public interface.

Technically you could do `this.session.session.on('authenticationSucceeded')` and it should work in the same way.

It's hard to talk about this without any code, but everything should work the same as long as mixins were correctly replaced with the session methods.

RobbieTheWagner:
@BobrImperator this is the PR updating things. It has a failing test because `handleAuthentication` is never called https://github.com/shipshapecode/swach/pull/1732

We do use https://github.com/adopted-ember-addons/ember-cognito and I am not sure if it needs changes to work with latest?

BobrImperator:
I made a PR for Swach, please see: https://github.com/shipshapecode/swach/pull/1736

BobrImperator:
Hi @hoIIer 

The Evented API had been removed from the service, a direct replacement would be extending the service and implementing a `handleAuthentication` and `handleInvalidation`.

e.g. [Test app example](https://github.com/mainmatter/ember-simple-auth/blob/master/packages/test-app/app/services/session.js)
```js

import Session from 'ember-simple-auth/services/session';

export default class SessionService extends Session {

  handleAuthentication() {
    super.handleAuthentication(...arguments);
    
    // Do you thing here
  }
}
```

Although if your only requirement is to render a specific thing on authenticated session, then you should be able to use the service's `isAuthenticated` property.


```js
{{!-- app/templates/application.hbs --}}
<div class="menu">
  …
  {{#if this.session.isAuthenticated}}
    <a {{on "click" this.invalidateSession}}>Logout</a>
  {{else}}
    {{#link-to 'login'}}Login{{/link-to}}
  {{/if}}
</div>
<div class="main">
  {{outlet}}
</div>
```

See [Walkthrough](https://github.com/mainmatter/ember-simple-auth?tab=readme-ov-file#walkthrough) and *Other guides* in the [readme](https://github.com/mainmatter/ember-simple-auth)

Lastly
> Also the event api still in the docs?

That appears to be a mistake, since the public SessionService doesn't emit anything and only hooks itself up to the internal, private one.

PR #2527, opened 2023-03-23
chore: prolong Torri authenticator deprecation period until v6

BobrImperator:
We won't be able to migrate from Torii or recommend a solution until v5.
Torii initially got deprecated due to being unable to run under Ember v4.
For now this has changed and gave us a little bit more leeway: https://github.com/adopted-ember-addons/torii


marcoow (approved):
👍

PR #2534, opened 2023-03-29
Re-polyfill Assign

jrjohnson:
Copying this polyfill from ember maintains compatibility with IE 11 and Ember 3.x and allows this addon to be consumed by applications which use Ember 5.x where the assign polyfill has been removed.

This helps with addons that consume this addon where the canary scenario is currently failing with `Global error: Uncaught Error: Could not find module `@ember/polyfills` imported from `ember-simple-auth/utils/assign``

### DIFF
--- a/packages/ember-simple-auth/addon/utils/assign.js
+++ b/packages/ember-simple-auth/addon/utils/assign.js
@@ -1,3 +1,24 @@
-import { assign } from '@ember/polyfills';
+/**
+ * Assign from @ember/polyfill, copied here to maintain compatilibity
+ * with Ember 3.x (and IE 11) as well as 5.x which drops this polyfill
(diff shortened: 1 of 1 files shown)


BobrImperator:
Thanks for making the PR, I wasn't aware that the polyfill is finally getting removed.
However currently ESA doesn't support ember-canary/v5. We could be aiming for this with the next major. There are many things needed to be done though, so I can't say when would that be.

As for the PR, I'm not too fond of hardcoding this polyfill in the code.
I've played around with this for a bit though, and I'm wondering if we could and should use macros instead to handle this situation.
Here's a branch that implements it, would you mind checking if this works in your application as well? https://github.com/mainmatter/ember-simple-auth/compare/use-macro-for-ember-polyfill
I'm also noticing that our test suite is not fully passing on ember-canary scenario which could mean that either of the fixes won't help us really and there's more needed to be done.

jrjohnson:
Yup, that's a way better solution!

PR #2538, opened 2023-04-03
feat: import @ember/polyfills assign conditionally

BobrImperator:
- Uses `@embroider/macros` to conditionally import `assign` polyfill to support older and the newest ember versions.
Should become obsolete once we decide to support only v4 and onwards where IE11 support is dropped.

Should go first #2537 

Inspired by https://github.com/mainmatter/ember-simple-auth/pull/2534


packages/ember-simple-auth/addon/utils/assign.js
```diff
@@ -1,3 +1,12 @@
...
+import { importSync } from '@embroider/macros';
+
+let assign;
+
+try {
+  assign = importSync('@ember/polyfills').assign;
+} catch (error) {
+  // Couldn't import @ember/polyfills
+  // Doesn't exist in v5
+}
```
BobrImperator, line 10: It wraps importSync in try catch becasue `dependencySatisfies` macro can't reliably work if the dependency isn't a direct dependency or a peer dependency.
Seems like `ember-source` matching in addons misbehaves especially

RobbieTheWagner: ember-source needs to be a peerDependency of this addon, I think. That should make it work.

RobbieTheWagner:
I would recommend dropping `Ember.assign` entirely. IE11 is not supported by Microsoft anymore anyway. Also, if someone really needs to support IE11 for some reason, it's easy to add an Object.assign polyfill and they probably already have one.

RobbieTheWagner:
I opened a PR to use Object.assign directly. I could add some documentation on how to add an Object.assign polyfill for IE11 if you'd like, but I think we're pretty safe not supporting IE11 at all. https://github.com/mainmatter/ember-simple-auth/pull/2542

BobrImperator:
@RobbieTheWagner Ember as a framework dropped IE11 support with v4.
The problem with removing polyfills altogether is not the work needed to be done, but dropping support for older ember versions. However I agree we need to get more radical and just do it:tm:  seeing how behind we are.

I'll try to resolve this in the coming days. Also I've been working on ember-cookies recently for the same reasons, since ESA uses it internally https://github.com/mainmatter/ember-cookies

packages/ember-simple-auth/addon/utils/assign.js
BobrImperator: Thanks for the pointer :+1: , I also realized this later on but I did get some breakage under fastboot while working on https://github.com/mainmatter/ember-cookies/pull/782

RobbieTheWagner:
I would argue that dropping older Ember versions isn't bad. They can always use older versions of this library. Supporting older Ember really holds a lot of addons back.

PR #2542, opened 2023-04-15
Use Object.assign and drop Ember.assign

RobbieTheWagner:


derrabauke:
Hey folks, any ambitions to release this soon? :-)

BobrImperator:
Hi @derrabauke 

Yes, however I'm vacationing at the moment and there's some manual checks I need to run before releasing.

Especially need to verify if our test-apps are failing on `ember-beta` and `ember-canary` channels due to fastboot/ember-data issues or there are problems with the library.

for context:

https://github.com/mainmatter/ember-simple-auth/pull/2411#issuecomment-1530965817
https://github.com/mainmatter/ember-simple-auth/issues/2550

jrjohnson:
Enjoy your vacation @BobrImperator! When you return (and not a moment sooner) I'm happy to test a beta release in our app or anything else I can do to be helpful pre-release.

derrabauke:
> Hi @derrabauke
> 
> Yes, however I'm vacationing at the moment and there's some manual checks I need to run before releasing.
> 
> Especially need to verify if our test-apps are failing on `ember-beta` and `ember-canary` channels due to fastboot/ember-data issues or there are problems with the library.
> 
> for context:
> 
> [#2411 (comment)](https://github.com/mainmatter/ember-simple-auth/pull/2411#issuecomment-1530965817) #2550

Thanks for the information and enjoy your holidays! :+1:

BobrImperator:
Hi folks, I've just released a `6.0.0-rc.1` version of ESA.

The reason why it's an RC version is that at this very moment `ember-canary`, `ember-beta` channels fail due to some `ember-source` issues in these versions.

Our tests rely on Fastboot which fails for `ember-data >= 4.12` versions 🤷 
Because of that I can't be confident that everything works correctly. I have tested things manually without Fastboot and it seems things are working.

So unless you're using Fastboot it seems like everything should work, but until I've confirmed what's going on you'll need to use the rc.1 version for the time being. If you're blocked, then use that.

I'm sorry for the inconvenience, I'll try to sort that out as soon as possible.


Changes made are here https://github.com/mainmatter/ember-simple-auth/blob/master/CHANGELOG.md

jrjohnson:
Initial testing is looking good for us. We had a small problem with embroider, but updating to the latest version of all embroider libs fixed that.

PR #2565, opened 2023-06-11
Remove IE 11 from browser targets

gilest:
Removed IE 11 from test apps and "dummy" app which should resolve `emberBabel.extends` errors.

Possible fix for https://github.com/emberjs/ember.js/issues/20458

IE 11 support was dropped in Ember 4.0

### DIFF
--- a/packages/classic-test-app/config/targets.js
+++ b/packages/classic-test-app/config/targets.js
@@ -6,13 +6,6 @@ const browsers = [
   'last 1 Safari versions'
 ];
 
-const isCI = Boolean(process.env.CI);
-const isProduction = process.env.EMBER_ENV === 'production';
-
-if (isCI || isProduction) {
-  browsers.push('ie 11');
-}
-
 module.exports = {
   browsers
 };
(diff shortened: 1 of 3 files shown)


packages/classic-test-app/package.json
```diff
@@ -38,7 +38,7 @@
     "ember-cli-inject-live-reload": "^2.0.1",
     "ember-cli-sri": "^2.1.1",
     "ember-cli-terser": "4.0.2",
-    "ember-data": "~4.4.0",
+    "ember-data": "~4.10.0",
```
BobrImperator, line 41: Please rollback the change and we should be good to go with the PR.

[Test scenarios](https://github.com/mainmatter/ember-simple-auth/blob/master/packages/ember-simple-auth/config/ember-try.js) should be using new ember-data correctly now :+1:

BobrImperator:
Please change the PR title to something like "Remove IE11 from browser targets" or "Enable testing against Ember v5"

The browser targets change is not really targeting any users and I'd be hesitant saying that this enables Ember v5 support.

BobrImperator:
Thanks for the PR!

PR #2570, opened 2023-06-23
Move to pnpm

BobrImperator:
- removes yarn and uses pnpm instead
- adds @embroider/test-setup to all packages because of symlinked ember-try scenarios that attempt to use it.

Note:
`ember-release` scenario should start passing after https://github.com/mainmatter/ember-simple-auth/pull/2565


Procfile.client
```diff
@@ -1 +1 @@
-web: yarn workspace test-app start:fastboot
+web: pnpm --filter test-app start:fastboot
```
BobrImperator, line 1:
```suggestion
web: pnpm run --filter test-app start:fastboot
```

marcoow (commented):
I’d suggest keeping Volta - that has beta pnpm support now

BobrImperator:
@marcoow It'd seem that this volta issue is blocking https://github.com/volta-cli/action/issues/121

marcoow:
We're using it in Gravity and (I think) it works quite well – https://github.com/mainmatter/gravity/blob/master/.github/workflows/ci.yml

BobrImperator:
`ember-lts-4.12` seems to be the only version failing and might be bugged, see https://github.com/emberjs/data/issues/8609

BobrImperator:
Merged since it seems like I've cleared out the concerns, I'd like to see if renovate works correctly and so we can merge new package versions without the need to manually maintain lock files in this PR anymore.

PR #2600, opened 2023-07-07
chore(test-app): enable embroider build

BobrImperator:
requires #2599 
closes #2509


marcoow (approved):

packages/test-app/ember-cli-build.js
```diff
@@ -20,5 +20,19 @@ module.exports = function(defaults) {
 
   app.import('node_modules/bootstrap/dist/css/bootstrap.css');
 
-  return app.toTree();
+  if (process.env.ESA_TEST_EMBROIDER) {
+    const { Webpack } = require('@embroider/webpack');
+    return require('@embroider/compat').compatBuild(app, Webpack, {
```
mansona, line 25: why did you do it this way and not use the `maybeEmbroider()` thing from `@embroider/test-setup`? 

The reason I ask is because we're going to be adding some more options to the compatBuild and this will default to being slightly less compatible if you don't use the tool provided

BobrImperator: This is for an app, mimicking a real world usage.

`maybeEmbroider` is used in other test app that was derived from ember-simple-auth monolith:
https://github.com/mainmatter/ember-simple-auth/pull/2604/files#diff-bf02927f31fb97e0a498f9d69b486f281b2b7e686c19018d6c3c96885c016aa3

PR #2613, opened 2023-07-31
Route After Invalidation Config

devotox:


packages/ember-simple-auth/addon/configuration.js
```diff
@@ -38,6 +53,12 @@ export default {
 
   load(config) {
     this.rootURL = config.rootURL !== undefined ? config.rootURL : DEFAULTS.rootURL;
+    
+    this.routeAfterInvalidation =
+      config.routeAfterInvalidation !== undefined
+        ? config.routeAfterInvalidation
+        : DEFAULTS.routeAfterInvalidation;
```
BobrImperator, line 60: Interesting addition, I like it, it's more descriptive than `rootURL` but they have the same function.
However I'm not a fan of resolving what route we should redirect to later in the service and I think it should be handled here in the configuration.

`rootURL` isn't used for anything else and `routeAfterInvalidation` is basically and alias for it.

I'd suggest something like this, where we either take `routeAfterInvalidation` value from the configuration or rootURL.
 
```suggestion
    this.routeAfterInvalidation =
      config.routeAfterInvalidation !== undefined
        ? config.routeAfterInvalidation
        : this.rootURL;
```

This would also need to be explained in the doc accordingly.

BobrImperator (commented):
It'd also be nice to have an acceptance test in the `test-app` that makes sure the configuration is used correctly.

PR #2712, opened 2024-01-09
chore(docs): migrate yuidoc to jsdoc

BobrImperator:
- Migrates yuidoc to jsdoc
There are small differences between jsdoc and yuidoc.

  - In our case jsdoc needed specyfing class members explicity.
  - {{crosslink}} is @link
  - In line code samples with decorator are not supported correctly an "at" sign (@) has to be written in unicode (&# 64;)
  
- Prepares the docs for automation
The most straighforward way to make it deployable to gh-pages was to move gh-pages configuration over to main repo, output the library docs inside it and then deploy all of it at once.

closes https://github.com/mainmatter/ember-simple-auth/issues/2709


marcoow (commented):
This seems fine overall but I wonder what specifically we're deploying to ember-simple-auth.com with this – wouldn't that be the docs so that in the root of ember-simple-auth.com you'd see those and we'd be losing the current landing page? We need to keep that and keep the docs under `/api` (or similar).

marcoow (approved):
nvm my previous comment - I see now that's not the case actually 👍

PR #2796, opened 2024-06-28
Create store.js

peter-scholtens:
Explicitly define store, to suppress deprecation warning, which appeared after logging in.

```
DEPRECATION: You are relying on ember-data auto-magically installing the store service. Use `export { default } from 'ember-data/store\';` in app/services/store.js instead [deprecation id: ember-data:deprecate-legacy-imports] This will be removed in ember-data 6.0.
```

### DIFF
--- a/packages/test-app/app/services/store.js
+++ b/packages/test-app/app/services/store.js
@@ -1,0 +1,1 @@
+export { default } from 'ember-data/store';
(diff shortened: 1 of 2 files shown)


peter-scholtens:
It is not clear to me if the other examples (**classic-test-app** and **test-esa**) also need this store.js file, as I see no site at all and lots of **Content-Security-Policy** warnings in the console.

BobrImperator:
`test-esa` is a pretty much empty app exclusively for unit tests.

`classic-test-app` should work though and [this](https://github.com/mainmatter/ember-simple-auth/pull/2798) was the reason why it didn't.

peter-scholtens:
Not needed anymore.

PR #2804, opened 2024-07-07
chore(lint): Migrate eslint config

BobrImperator:
- Migrates eslint config to the new syntax
- Unifies configs so they're defined only once and imported where needed
- Applies autofixes
- Turned off the 'no-prototype-builtins' rule.


marcoow:
what about the todos https://github.com/mainmatter/ember-simple-auth/pull/2804/files#diff-9601a8f6c734c2001be34a2361f76946d19a39a709b5e8c624a2a5a0aade05f2R66 ?

BobrImperator:
@marcoow they were there in the original file, I don't think we'd like to address them here?

This PR only addresses config migration and formatting that was somehow not applied for other files.
e.g. inconsitencies around spacing and `""` -> `''` conversion.

marcoow:
ah I see

marcoow (approved):

PR #2811, opened 2024-08-26
feat: support partitioned cookies

Gaurav0:


Gaurav0:
@BobrImperator Updated. Note that it is only failing in ember-release scenario (Ember 6.0).

See #2866

PR #2813, opened 2024-09-03
feat: Add scope in restore for OAuth2PasswordGrant

bpetetot:
I'm currently using the `OAuth2PasswordGrant` authenticator.

I'd like to use the `scope` property:

- The `scope` property is correctly passed to the `authenticate` method of `OAuth2PasswordGrant`.

- But it is not passed to the `restore` method when refreshing the access token. The specification says that it is possible to pass an optional `scope` parameter to the token refresh request. [rfc6749](https://datatracker.ietf.org/doc/html/rfc6749#page-47)

SCR-20240903-papt

So I made this PR (including tests) to add the `scope` property to the refresh request.

Thanks for the review and `ember-simple-auth` 🙂

### DIFF
--- a/packages/ember-simple-auth/src/authenticators/oauth2-password-grant.js
+++ b/packages/ember-simple-auth/src/authenticators/oauth2-password-grant.js
@@ -91,6 +91,18 @@ export default BaseAuthenticator.extend({
   refreshAccessTokens: true,
 
   /**
+    Sets whether the authenticator use the scope when refreshing access tokens
+    if the server supports it.
+
+    @memberof OAuth2PasswordGrantAuthenticator
+    @property refreshAccessTokensWithScope
+    @type Boolean
+    @default false
+    @public
+  */
+  refreshAccessTokensWithScope: false,
+
+  /**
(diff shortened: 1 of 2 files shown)


BobrImperator:
Hi, thanks for the PR.
It seems like some tests are failing, some fails are expected e.g. `ember-beta`, `ember-release` but `ember-4.0` should be passing.

bpetetot:
I've fixed the code and the tests are now passing (locally on `ember-4.0`)
I've also added some tests about the `scope` in the `module('#_refreshAccessToken')` part.

I hope all tests will run now :)

bpetetot:
You're right, a feature flag might be needed to avoid a potential break in some cases.

I've added a `refreshAccessTokensWithScope` property which is `false` by default (I can change the name if necessary).

All operations concerning the `scope` in the `refreshAccessToken` function are only activated if the `refreshAccessTokensWithScope` property is `true`. I've updated the tests accordingly.

I'm not sure how I should update the documentation on this, I've just added the JSDoc on this.

PR #2883, opened 2024-12-26
refactor(typescript): migrate BaseAuthenticator and OAuthPasswordGrant

BobrImperator:
- Refactors BaseAuthenticator and OAuth2PasswordGrantAuthenticator to typescript
- Adds typed-event-target as devDependency to allow typing CustomEvents emmited by authenticators.
- Removes a bunch of `isEmpty` calls from the authenticators and relies on native mechanisms instead.
- Deprecates `responseText` and `responseJSON` returned by the OAuth2PasswordGrantAuthenticator.
- Bumps torii deprecation to 8.0.0

### DIFF
--- a/packages/ember-simple-auth/package.json
+++ b/packages/ember-simple-auth/package.json
@@ -62,7 +62,8 @@
     "prettier": "3.3.3",
     "rollup": "4.25.0",
     "rollup-plugin-copy": "3.5.0",
-    "typescript": "^5.7.2"
+    "typescript": "^5.7.2",
+    "typescript-event-target": "^1.1.1"
   },
   "publishConfig": {
     "registry": "https://registry.npmjs.org"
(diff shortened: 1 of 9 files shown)


BoussonKarel:
What are the alternatives for `reponseJSON` and `reponseText`? I use these to show the error returned by the backend on login and since the response you return has already been read, it's not an option to use that one.\

Is there any chance you could return a copy of the response (before it was read)? (or copy the response, read that one and send the original)

BobrImperator:
Hi @BoussonKarel. As you've noticed, there's no clear migration path from these properties which was an oversight.
I've taken a look and made an issue along with a draft PR that implements your suggestion see [issue](https://github.com/mainmatter/ember-simple-auth/issues/2923) and a [PR](https://github.com/mainmatter/ember-simple-auth/pull/2922).

I'm not sure if this is the way forward however, I should be able to take a better look later this week to align the password authenticator with other authenticators instead in backwards compatible way.

Issue #2864, opened 2024-11-12
Build error while trying to run tests locally


Gaurav0:
Versions:

ember-cli: 5.12.0
node: 20.18.0
pnpm: 9.12.3
os: darwin arm64 (MacOS Sequoia 15.1)

In package test-esa:

```
[gaurav.munjal@MacBookPro test-esa (gaurav/feat/support_partitioned_cookies)]$ ember test   
Environment: test
⠋ BuildingTorii is installed but not configured in config/environment.js!
cleaning up...
Cleanup error.
⠋ cleaning upCannot read properties of undefined (reading 'cleanup')


Stack Trace and Error Report: /var/folders/5n/x1_c935x0sq7lnqzfpsjkt080000gn/T/error.dump.873624881ff2ed0e4589d628770e0a6b.log
ENOENT: no such file or directory, stat '/Users/gaurav.munjal/dev/work/mainmatter/ember-simple-auth/packages/ember-simple-auth/node_modules/.bin/pnpm'


Stack Trace and Error Report: /var/folders/5n/x1_c935x0sq7lnqzfpsjkt080000gn/T/error.dump.9fa6036dd8b668dd27bb4d3ea3bda321.log
[gaurav.munjal@MacBookPro test-esa (gaurav/feat/support_partitioned_cookies)]$ node -v
v20.18.0
[gaurav.munjal@MacBookPro test-esa (gaurav/feat/support_partitioned_cookies)]$ npm -v
10.8.2
[gaurav.munjal@MacBookPro test-esa (gaurav/feat/support_partitioned_cookies)]$ pnpm -v
9.12.3
[gaurav.munjal@MacBookPro test-esa (gaurav/feat/support_partitioned_cookies)]$ ember -v
ember-cli: 5.12.0
node: 20.18.0
os: darwin arm64
```

Note that I get the same error running:
pnpm run --filter test-esa test
pnpm run --filter test-esa test:one ember-default

Gaurav0:
It looks like adding

```
hoist=true
hoist-workspace-packages=true
```

to .npmrc and rerunning `pnpm i` solves this problem. Would you like me to open a PR with this change?

BobrImperator:
Thanks for vigilance!

We don't want to enable hoisting or any additional configuration.
It seems like the issue might've been a bloated lockfile, usually renovate mitigates the problem but we might've missed a lockfile maintenance PR. Everything should be fine now I think.

[...]

PR #2905, opened 2024-12-30
feat(test-esa) fix missing test scenarios

Gaurav0:
Adds missing test scenarios for Ember 5.4, 5.12 LTS

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -59,7 +59,9 @@ jobs:
           - test:one ember-lts-4.4
           - test:one ember-lts-4.8
           - test:one ember-lts-4.12
+          - test:one ember-lts-5.4
           - test:one ember-lts-5.8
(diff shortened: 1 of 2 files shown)


.npmrc
```diff
@@ -1 +1,3 @@
 resolution-mode=highest
+hoist=true
+hoist-workspace-packages=true
```
BobrImperator, line 3: This configuration would likely indicate problems with peer dependencies / dependencies of packages. 
It used to be the trick in yarn to make monorepos work.

We should make the package resolutions stricter rather than relaxing it.
e.g.

```
# resolution-mode is here because of old pnpm problem in pre-8.0. This setting is now the default
resolution-mode=highest
auto-install-peers=false
resolve-peers-from-workspace-root=false
```

Gaurav0: Reverted this; but I'm still having the issue in #2864

PR #2951, opened 2025-03-05
session.setup must be guarded to avoid double calling

betocantu93:
If you call setup twice, multiple event listeners are added: `authenticationSucceeded` `invalidationSucceeded`

This could happen easily if you place the setup call on any beforeModel, say application route beforeModel, without manual guarding. Any transition that happens in between while executing the beforeModel aborts the transition and triggers a new one, which causes the beforeModel to be triggered again, thus calling setup again. This prob should be guarded from within the lib.

As a work around we extended the service with:

```ts
async safeSetup() {
    if (!this._setupIsCalled) {
      return this.setup();
    }
  }
```

### DIFF
--- a/packages/ember-simple-auth/src/services/session.ts
+++ b/packages/ember-simple-auth/src/services/session.ts
@@ -334,11 +334,13 @@ export default class SessionService<Data = DefaultDataShape> extends Service {
     @public
   */
   setup() {
-    this._setupIsCalled = true;
-    this._setupHandlers();
-
-    return this.session.restore().catch(() => {
-      // If it raises an error then it means that restore didn't find any restorable state.
-    });
+    if(!this._setupIsCalled) {
+      this._setupIsCalled = true;
+      this._setupHandlers();
+  
+      return this.session.restore().catch(() => {
+        // If it raises an error then it means that restore didn't find any restorable state.
+      });
+    }
   }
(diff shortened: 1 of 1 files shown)


packages/ember-simple-auth/src/services/session.ts
```diff
@@ -334,11 +334,13 @@ export default class SessionService<Data = DefaultDataShape> extends Service {
...
-      // If it raises an error then it means that restore didn't find any restorable state.
-    });
+    if(!this._setupIsCalled) {
+      this._setupIsCalled = true;
+      this._setupHandlers();
+  
+      return this.session.restore().catch(() => {
+        // If it raises an error then it means that restore didn't find any restorable state.
+      });
+    }
```
BobrImperator, line 344: Thanks for the PR :+1: 
It seems like one alternative could be to use an `instance-initializer` instead.

Could you add a test asserting that `_setupHandlers` can be called only once?

We might want to make sure that the setupHandlers runs only once while `restore` continues to be called.
```suggestion
    if (!this._setupIsCalled) {
      this._setupIsCalled = true;
      this._setupHandlers();
    }
      
    return this.session.restore().catch(() => {
      // If it raises an error then it means that restore didn't find any restorable state.
    });
```

PR #2953, opened 2025-03-18
Addon dependency maintenance

Windvis:
- remove `silent-error` and `ember-cli-is-package-missing` since they are no longer used
- add support for [`@ember/test-waiters` v4](https://github.com/emberjs/ember-test-waiters/blob/3f977b5d9f2c4f6fe051b0b41e8c30ca367b4045/CHANGELOG.md#release-2024-12-05) which allows apps to use v4 without using overrides


BobrImperator:
Note: #2954 introduces a change that requires Ember 4.1.0 >= . This PR requests support of test-waiters v4 which also have dropped support of Ember < v4

PR #2954, opened 2025-03-21
Replaced all 'inject as service' to 'service' as it will fail in ember 7.0.0.

peter-scholtens:
As [inject as service](https://deprecations.emberjs.com/id/importing-inject-from-ember-service/) is deprecated and will be removed in Ember 7.0.0, I run the [script](https://github.com/ijlee2/ember-codemod-remove-inject-as-service) and aligned it.


BobrImperator:
Note: this means we need to drop < 4.1.0, this is when the import was introduced.

Windvis:
> Note: this means we need to drop < 4.1.0, this is when the import was introduced.

It's also possible to do something like this so you don't need to drop ember support:

```js
import * as s from '@ember/service';
const service = s.service ?? s.inject;
```

BobrImperator:
That's not a great solution I think.
The comment wasn't a prompt to the author to fix this but just a note to myself.

`import * as s from '@ember/service';` is not great for tree shaking and I'm fine with dropping a version.

@peter-scholtens could you please revert the `import * from ...`  change?

peter-scholtens:
__Done.__
I already wondered how to revert `import * as s from '@ember/service';` after support for < 4.10 would be dropped, so it's cleaner to avoid problem at all.
_(I couldn't read the discord channel, so I had no idea what the intention was)._

Windvis:
> `import * as s from '@ember/service';` is not great for tree shaking and I'm fine with dropping a version.

To be fair, no tree shaking is going to happen in this scenario, so I still think it's a good solution if you want to keep supporting older versions in an addon.

It's also fine if you just want to drop support and cut a major (although it seemed overkill to me for such a simple thing, hence my suggestion :smile:).

peter-scholtens:
@BobrImperator Okay. no modification on that (I'm new with JS, so followed the suggestion from the warning itself).

PR #2959, opened 2025-03-26
Add `waitForPromise` calls to the SessionService methods

### LINKED ISSUE #2272: Using ESA in application (acceptance) tests requires manual use of test waiters
It seems that to safely use ESA in an application test requires each app to wrap any action handlers that trigger async methods on the ESA session with the `waitFor` decorator from `@ember/test-waiters`

See details in [this discuss thread](https://discuss.emberjs.com/t/ember-test-with-async-await-functions-not-working-as-intended/18435/6?u=gabrielgrant)

While explicitly using `waitFor` in app code does seem to work, it also doesn’t feel like the type of thing every app should have to concern itself with. Would it make sense to have (ie would the maintainers be open to receiving PRs that implement) test waiting behavior within the [internal session](https://github.com/simplabs/ember-simple-auth/blob/3.1.0-beta.0/packages/ember-simple-auth/addon/internal-session.js#L24) itself, so that tests would be paused until promises from authenticators' hooks either resolve or reject?

Or is the intention that this be something that should be handled in each authenticator? Or does it really need to be handled by each app individually? In either of these cases, it would be helpful to make note of it (either in the Implementing a Custom Authenticator](https://github.com/simplabs/ember-simple-auth#implementing-a-custom-authenticator) section or in the [Testing](https://github.com/simplabs/ember-simple-auth#testing) section respectively)

Windvis:
I'm not 100% sure this is the correct place to add these waiters, and if this can be tested somehow.

Closes #2272

### DIFF
--- a/packages/ember-simple-auth/src/services/session.ts
+++ b/packages/ember-simple-auth/src/services/session.ts
@@ -2,6 +2,7 @@ import Service from '@ember/service';
 import { getOwner } from '@ember/application';
 import { assert } from '@ember/debug';
 import Configuration from '../configuration';
+import { waitForPromise } from '@ember/test-waiters';
 
 import {
   requireAuthentication,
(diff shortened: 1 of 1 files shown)


Windvis:
Closing again since I won't have time in the near future to look into why CI is failing.

PR #2969, opened 2025-04-14
incorrect credentials doesn't show data-test-error-message

peter-scholtens:
Resolves error when trying to log in with with incorrect credentials: "Uncaught (in promise) TypeError: Response.clone: Body has already been consumed."

### DIFF
--- a/packages/test-app/app/components/login-form.js
+++ b/packages/test-app/app/components/login-form.js
@@ -25,8 +25,7 @@ export default class LoginFormComponent extends Component {
         this.session.set('store.cookieExpirationTime', null);
       }
     } catch (response) {
-      let responseBody = await response.clone().json();
-      this.errorMessage = responseBody;
+      this.errorMessage = response.responseJSON.error;
     }
   }
 


peter-scholtens:
Unfortunately, the test [logging in with incorrect credentials shows an error](https://github.com/mainmatter/ember-simple-auth/blob/c2c1d1e5ab9838bd1d3f3fe0daaba65d85454a26/packages/test-app/tests/acceptance/authentication-test.ts#L67) doesn't reveal the error. It is only visible as missing in the browser (Firefox 137.0.1 / Chromium 135.0.7049.52) itself.

peter-scholtens:
As a reminder: using `responseJSON` is not futureproof, see [issue 2923](https://github.com/mainmatter/ember-simple-auth/issues/2923).

BobrImperator:
Well spotted! I'd like to hold off on this though because using the `responseJSON` property would be incorrect just as per the comment you've linked.

I've been thinking about a migration path for a while but it's not exactly as easy as I initially thought because we rely on the response internally and simply cloning it leads to a very weird API.

Let's leave this PR open as a reminder,  I probably won't merge it however.

BobrImperator:
https://github.com/mainmatter/ember-simple-auth/pull/3007

PR #2983, opened 2025-09-13
chore(deps): use torii 1.0.0

BobrImperator:


BobrImperator:
The latest `master` branch already had torii 1.0.0 :sweat_smile:

PR #3002, opened 2025-11-02
test: setup playwright-tests repo with e2e tests

BobrImperator:
- Replaces ember-6.0 scenario with ember-lts-6.4

This is so we're able to test various synchronization scenarios across tabs and perhaps, move some of the acceptance tests over to as well.
Currently there's no good, reliable way to test an actual outcome between tabs and we could only perform tests against method calls.


BobrImperator:
Merging since the `ember-release` fails appear unrelated to the changes

PR #3003, opened 2025-11-02
cleanup: remove ember classic-test-app

BobrImperator:
- Removes the `classic-test-app`
This is very ancient stuff by now and it should be OK to sunset it.
- Removes `node-tests` from CI
Ember addons don't have a dependency on node since the V2 addons format
- Removes `extra-tests` (Floating dependencies) test
`pnpm install` already throws errors for `pnpm install` when `CI=true`. This explicit test is not needed.


BobrImperator:
Merging since ember-release fail is unrelated

Issue #2808, opened 2024-07-30
Retry transitions after page reloads


Windvis:
ember-simple-auth supports retrying transitions after the session is authenticated, but this only works if the page isn't refreshed. `session.attemptedTransition` isn't persisted, so the info is gone after a reload. 

ember-simple-auth _does_ already support a cookie-based retry, but that is [only enabled during FastBoot](https://github.com/mainmatter/ember-simple-auth/blob/1404c501c2ba9c8c6c071b05f38d42a057522318/packages/ember-simple-auth/src/-internals/routing.js#L9-L15). Would it be an option to either always use this cookie-based setup, or make it an opt-in somehow?

**Our use-case**
We have an authenticator that redirects users to an authentication server on a different domain, but as a result the original transition information is lost.

We can hook into the cookie-based redirect setup by writing the url to the `ember_simple_auth-redirectTarget` cookie ourselves (in the overriden `requireAuthentication` method of our app's SessionService). But it would be nice if this wouldn't require custom app code.

<details>
<summary>Code of our workaround</summary>

```js
export default class SessionService extends BaseSessionService {
  @service router;

  requireAuthentication(transition, routeOrCallback) {
    const isAuthenticated = super.requireAuthentication(
      transition,
      routeOrCallback
    );

    if (!isAuthenticated) {
      const COOKIE_NAME = 'ember_simple_auth-redirectTarget';
      const redirectUrl = routeInfoUrl(transition.to, this.router); // This is a util that reimplements `transition.intent.url` with public APIs
      document.cookie = `${COOKIE_NAME}=${redirectUrl};path=/;samesite=strict`;
    }

    return isAuthenticated;
  }
```
</details>

betocantu93:
We did this by adding a Query Param on application route

`next`

Basically saves the URL in the QP when auth is required, and reads from it on handleAuthentication.

But, this cookie approach seems more plug-n-play! and also seems more resilient for integrations redirects I guess.

BobrImperator:
Here's an implementation of this feature which I should be able to release this week, but here's an early look. Feel free to comment on this. 
https://github.com/mainmatter/ember-simple-auth/pull/3006

The following is an example of how you could use it in your apps:
```ts
import type { Transition, RouteOrCallback, ExtraAuthenticationArgs } from 'ember-simple-auth/services/session.ts';


[...]

PR #3006, opened 2025-11-03
Feature: formalize `redirectTarget` across stores. Add ability to set `redirectTarget` with `requireAuthentication` method

BobrImperator:
- Adds persisting a `redirectTarget` for all stores.
**Old**
Until now `redirectTarget` was an implementation of `attemptedTransition` retrying for Fastboot, where storing Transition object would not be possible. This resulted in (Only Fastboot) environment to have a persisted `redirectTarget`  using the cookie store.
**New**
Storing `redirectTarget` is now more intentional and available for all `session-stores` i.e. your `session-store/application.js` will be used for persisting the `redirectTarget`.
Additionally it's possible to pass a `redirectTarget` to `requireAuthentication` method:

```js
this.session.requireAuthentication(transition, 'login', { redirectTarget: "/my-designated-route" })
```

For users who want to set the `redirectTarget` based on route query params or any other source they can override the `requireAuthentication` method on the `session` service.

```ts
import type { Transition, RouteOrCallback, ExtraAuthenticationArgs } from 'ember-simple-auth/services/session.ts';


requireAuthentication(transition: Transition, routeOrCallback: RouteOrCallback, extraArgs: ExtraAuthenticationArgs = {}) {

  return super.requireAuthentication(transition, routeOrCallback, {
    // order might matter depending on what you want to achieve
     ...extraArgs,
     redirectTarget: '/i-want-to-go-there-after-login'
  })
}
```

related to #2808

### DIFF
--- a/packages/ember-simple-auth/src/-internals/routing.js
+++ b/packages/ember-simple-auth/src/-internals/routing.js
@@ -7,7 +7,7 @@ export function requireAuthentication(owner, transition, extraArgs) {
   let isAuthenticated = sessionService.get('isAuthenticated');
   if (!isAuthenticated) {
     const internalSession = sessionService.session;
-    let redirectTarget = extraArgs.redirectTarget;
+    let redirectTarget = extraArgs?.redirectTarget;
 
     if (transition) {
       sessionService.set('attemptedTransition', transition);
--- a/packages/test-esa/tests/unit/services/session-test.js
+++ b/packages/test-esa/tests/unit/services/session-test.js
@@ -135,9 +135,11 @@ module('SessionService', function (hooks) {
   module('requireAuthentication', function (hooks) {
     let transition;
     let router;
+    let redirectTarget;
 
     hooks.beforeEach(async function () {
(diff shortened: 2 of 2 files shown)


pomm0:
This is a breaking change for custom stores, would have expected it to be released as a major version.

will [the api-docs](https://ember-simple-auth.com/api/BaseStore.html) be updated?

BobrImperator:
@pomm0 what's the error you're experiencing?
Is the issue that the new methods don't exist for custom stores? If so then we should be able to provide a blanket/noop implementation.

For the time being, you should be safe to provide these methods on your custom store and make them noop e.g.
```js
getRedirectTarget() { return null }
```

Otherwise you might implement these methods for the storage you're using in your app.

pomm0:
@BobrImperator yes, its because of the missing new methods. I'm getting this error after the login:

`this.session.authenticate('authenticator:CUSTOM_NAME', this.username, this.password);`

Selection_1525


I did implement these methods properly, just wanted to inform about this breaking change

PR #3014, opened 2025-11-07
Fix FastBoot check in LocalStorageStore always returning false

felixbuenemann:
Currently the FastBoot check in LocalStorageStore always returns `false` because `_isFastBoot` is initialised to false and then this value is always used because `hasOwnProperty()` returns true, so the `isFastBoot()` check is never executed.

This is a bug from conversion to class syntax and typescript.


BobrImperator:
Thanks :open_hands:

PR #3091, opened 2026-04-11
ember 7 readiness

NullVoxPopuli:


package.json
```diff
@@ -29,6 +29,11 @@
     "release-it": "19.0.6",
     "release-plan": "0.17.2"
   },
+  "pnpm": {
+    "overrides": {
+      "ember-cli-htmlbars": "^7.0.0"
```
NullVoxPopuli, line 34: there are a good few legacy dependencies in the dep tree that I don't want to uppdate, so just adding an override for now

NullVoxPopuli:
> mainmatter/continue-on-error-comment@v1
> exit 1

You've become the very thing you swore to destroy

PR #3105, opened 2026-06-12
Remove `@embroider/macros` from `dependencies`

NullVoxPopuli:
macros is an external as it is a build time behavior controlled by the consumer

also we don't want macros to show up in optimized deps

### DIFF
--- a/packages/ember-simple-auth/package.json
+++ b/packages/ember-simple-auth/package.json
@@ -31,7 +31,6 @@
   "dependencies": {
     "@ember/test-waiters": "^3 || ^4",
     "@embroider/addon-shim": "^1.0.0",
-    "@embroider/macros": "^1.0.0",
     "ember-cookies": "^1.4.1"
   },
   "devDependencies": {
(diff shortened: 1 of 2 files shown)


BobrImperator:
In that case it needs to be a peerDependency at the very least, since `@embroider/macros` must be present at build time for consumers. While it's also not generally installed in apps themselves unless the apps have their own need for it.
I'm not opposed here but AFAIK most addons I've seen declare it a dependency, which could've been a multi-year mistake :sweat_smile:. 
Overall this likely is a breaking change?

l[iquid-fire#package.json ](https://github.com/ember-animation/liquid-fire/blob/master/liquid-fire/package.json)
[ember-qunit@9.04#package.json](https://www.npmjs.com/package/ember-qunit/v/9.0.4?activeTab=code) -> `ember-qunit@9.1.0` already changed it.

NullVoxPopuli:
> In that case it needs to be a peerDependency at the very least, 

nay, it's not a runtime dependency

> since @embroider/macros must be present at build time for consumers.

there is nothing to be present -- macros is actually just signalling to the consumers' babel

> I'm not opposed here but AFAIK most addons I've seen declare it a dependency,

they are all wrong 🙈 lol

> which could've been a multi-year mistake 

yup!

>  Overall this likely is a breaking change?

nope -- bugfix -- in the worst case, users won't notice -- in the best case, it unblocks people from using vite (blockers include: duplicate macros state (isTesting, etc))

